PR #30, opened 2026-02-12
feat(postgres,i18n): add pg schema selection and Spanish locale

### LINKED ISSUE #15: Database explorer only shows public scheme on Postgres
When there's multiple schemes only public scheme shows on the explorer.

Reproduce steps:

Environment: Macos 15.7.3
Installed dmg version: 0.8.10

1. Add a Postgres connection with multiple schemes
2. Just view explorer and any table that's not on public scheme 

Image

I will probably take a shot at this.

kconfesor:
Fixes #15 

   - Adds PostgreSQL schema aware driver with multi-schema selection UI
   - Smart-quote all identifier references with escape_identifier for safety (MacOS)
   - Pass schema parameter through all postgres driver queries and commands
   - Add SidebarSchemaItem component for per-schema tree rendering
   - Add complete Spanish (es) i18n locale with all translation keys
   - Register Spanish in i18n config, AppLanguage type, and Settings UI
   - Add "spanish" label key to English and Italian locale files
   - Fixed mysql 8 empty table names issues (macos)

### DIFF
--- a/src/components/layout/Sidebar.tsx
+++ b/src/components/layout/Sidebar.tsx
@@ -66,6 +66,7 @@ import type { TableColumn } from "../../types/schema";
 import type { ContextMenuData } from "../../types/sidebar";
 import type { RoutineInfo } from "../../contexts/DatabaseContext";
 import { groupRoutinesByType } from "../../utils/routines";
+import { formatObjectCount } from "../../utils/schema";
 
 export const Sidebar = () => {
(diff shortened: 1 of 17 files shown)


kconfesor:
Hey there, sure i will check the issues, i will also add the summary feature to other drivers, about the schema selector yes we can abstract this one and use it to select databases, that's a great idea.

PR #39, opened 2026-02-22
fix: database dropdown selection on click

### LINKED ISSUE #38: [Bug]: Database dropdown selection fails when clicking items in the "New Connection" modal.
### Describe the bug

In the "New Connection" modal, the "Database Name" dropdown does not allow users to select a database by clicking on it. While the dropdown lists available databases after clicking "Load Databases", clicking an item in the list closes the dropdown without actually selecting the clicked item. Manual entry via typing and pressing 'Enter' works, but mouse interaction is broken.

### To Reproduce

1. Open the New Connection modal.
2. Select Postgres (or MySQL).
3. Fill in the connection details (Host, Port, Username, Password).
4. Click Load Databases.
5. Click on the database selection dropdown.
6. Attempt to click one of the databases in the resulting list.
7. Observe that the list closes instantly, but the input field remains empty or unchanged.

### OS Version

Windows 11 x64

### Tabularis Version

0.8.15

### Relevant Log Output

```shell
No explicit error logs were generated; this was a UI race condition
```

devsynck:
Fix #38

### DIFF
--- a/src/components/ui/SearchableSelect.tsx
+++ b/src/components/ui/SearchableSelect.tsx
@@ -32,6 +32,7 @@ export const SearchableSelect = ({
   const containerRef = useRef<HTMLDivElement>(null);
   const buttonRef = useRef<HTMLButtonElement>(null);
   const searchInputRef = useRef<HTMLInputElement>(null);
+  const dropdownRef = useRef<HTMLDivElement>(null);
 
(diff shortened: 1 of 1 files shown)


debba:
Good catch, thanks.
I'll merge it now.

PR #53, opened 2026-03-04
fix: handle UUID primary keys in PostgreSQL row deletion and update (#50)

### LINKED ISSUE #50: [Bug]: Can't delete row(s) in PostgreSQL table
### Describe the bug

Image

### To Reproduce

tried to delete
failed
it's probably due to UUID being primary key, instead of text which tabularis expects

### OS Version

PopOS Cosmic Linux

### Tabularis Version

0.9.4

### Relevant Log Output

```shell

```

xvchris:
## Summary

Fixes #50 — PostgreSQL row deletion (and update) fails when the primary key column is of type `UUID`.

## Root Cause

When a UUID primary key value arrives as a JSON string from the frontend, SQLx binds it as a Rust `String`, which PostgreSQL receives as `TEXT`. Due to PostgreSQL's strict type system, `TEXT` does not auto-cast to `UUID` in parameterized queries, causing the `WHERE pk = $1` clause to silently match zero rows.

## Fix

Before binding a string PK value, attempt to parse it as `uuid::Uuid`. If parsing succeeds, bind as a native UUID type; otherwise fall back to binding as a string.

Applied to both:
- `delete_record` — the reported bug
- `update_record` — same latent bug in the WHERE clause

## Changes

- `src-tauri/src/drivers/postgres/mod.rs`: 13 lines added, 2 removed

## Notes

- No changes needed for MySQL (stores UUIDs as `CHAR(36)`) or SQLite (dynamically typed)
- The `uuid` crate was already a dependency with the correct features enabled
- `cargo clippy` and `cargo fmt` pass cleanly

### DIFF
--- a/src-tauri/src/drivers/postgres/mod.rs
+++ b/src-tauri/src/drivers/postgres/mod.rs
@@ -9,6 +9,7 @@ use crate::models::{
 use crate::pool_manager::get_postgres_pool;
 use extract::extract_value;
 use sqlx::{Column, Row};
+use uuid::Uuid;
 
 // Helper function to escape double quotes in identifiers for PostgreSQL
 fn escape_identifier(name: &str) -> String {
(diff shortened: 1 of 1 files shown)


xvchris:
Thanks for the detailed fix on `fix/pg-delete-update-uuid`! I can see our PR only addressed `pk_val` binding in WHERE clauses, but missed the critical issue: **`update_record` and `insert_record` also need UUID handling for the new column values** (SET / VALUES), not just the WHERE clause.

Your approach with `CAST(... AS uuid)` is cleaner for `QueryBuilder` contexts too.

Happy to close this PR in favor of your branch — it's more comprehensive. Sorry for the incomplete fix! 🙏

PR #72, opened 2026-03-06
feat: added a new field in settings screen, regarding maxConnection c…

ealvesss:
PR Description

  feat: make connection pool size configurable via Settings

  Problem

  The maximum number of database connections per pool was hardcoded in pool_manager.rs (10 for MySQL/PostgreSQL, 5 for SQLite), with no way for users to adjust it.

  Changes

  Backend (src-tauri/src/)
  - config.rs: Added max_connections: Option<u32> to AppConfig, its merge logic in save_config, and a get_pool_max_connections() helper that reads the config file without requiring AppHandle.
  - pool_manager.rs: Replaced all hardcoded .max_connections() values in MySQL, PostgreSQL, and SQLite pool creation with crate::config::get_pool_max_connections().

  Frontend (src/)
  - contexts/SettingsContext.ts: Added maxConnections to the Settings interface and DEFAULT_SETTINGS.
  - pages/Settings.tsx: Added a number input for Connection Pool Size in the Data Editor section of the General settings tab.
  - i18n/locales/{en,it,es}.json: Added translation keys maxConnections, maxConnectionsDesc, and connections.

  Behavior

  - The setting takes effect when a new connection pool is created. Users must disconnect and reconnect for changes to apply to existing connections.
  - Falls back to the default value if no config is present.


ealvesss:
@debba the "code-review" bots run again, after change the MR?

src-tauri/src/config.rs
```diff
@@ -142,6 +142,10 @@ pub fn save_config(app: AppHandle, config: AppConfig) -> Result<(), String> {
             existing_config.plugins = config.plugins;
         }
 
+        if config.plugins.is_some() {
+            existing_config.plugins = config.plugins;
+        }
+
```
debba, line 148: This block is a duplicate of the one just above (lines 141-143). It causes a **borrow-after-move** compilation error because `config.plugins` is already moved on the first assignment.

```suggestion
```

debba (changes requested):
The `config.plugins` block in `save_config` is duplicated, causing a borrow-after-move compilation error. Remove the duplicate block (lines 145-148).

PR #76, opened 2026-03-08
feat: add connections group

fandujar:
PR Description

Feature: Add connections group

Screenhots:
Screenshot 2026-03-08 at 00 57 30

Screenshot 2026-03-08 at 00 57 47

Screenshot 2026-03-08 at 00 57 39

### DIFF
--- a/src-tauri/src/persistence.rs
+++ b/src-tauri/src/persistence.rs
@@ -1,4 +1,3 @@
-use crate::keychain_utils;
 use crate::models::{ConnectionGroup, ConnectionsFile, SavedConnection};
 use std::fs;
 use std::path::Path;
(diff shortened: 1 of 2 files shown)


src-tauri/src/persistence.rs
```diff
@@ -1,18 +1,42 @@
...
 
-    // Populate passwords from keychain if needed
-    for conn in &mut connections {
+    let mut result = ConnectionsFile {
+        groups: Vec::new(),
+        connections,
+    };
+    populate_keychain_passwords(&mut result.connections);
+    Ok(result)
+}
```
debba: **Performance fix**: `load_connections_file` is called on every DB operation (via `find_connection_by_id`), on every CRUD operation, and during migrations. It currently populates passwords from the keychain for **all** connections every time — with N saved connections, every single DB query reads the keychain N times.

`find_connection_by_id` already reads the keychain only for the specific connection it finds (in the lines that follow), so the `populate_keychain_passwords` call here is redundant and costly.

```suggestion
/// Load connections file (raw, no keychain reads).
/// Supports both old format (array of connections) and new format (with groups).
/// Callers that need passwords must fetch them explicitly for the specific connection.
pub fn load_connections_file(path: &Path) -> Result<ConnectionsFile, String> {
    if \!path.exists() {
        return Ok(ConnectionsFile::default());
    }
    let content = fs::read_to_string(path).map_err(|e| e.to_string())?;

    // Try parsing as the new format first
    if let Ok(file) = serde_json::from_str::<ConnectionsFile>(&content) {
        return Ok(file);
    }

    // Fall back to old format (array of connections)
    let connections: Vec<SavedConnection> = serde_json::from_str(&content)
        .map_err(|_| "Failed to parse connections file".to_string())?;

    Ok(ConnectionsFile {
        groups: Vec::new(),
        connections,
    })
}
```

PR #130, opened 2026-04-12
feat(ui) remove table limits for monaco autosuggestion

thomaswasle:
This PR removes the number of tables limitations which where implemented in the use of monaco.

I tested with database which have 6000+ tables and there seem to be no significant performance deteriorations.


debba:
Looks good to me, thanks @thomaswasle I will merge it

PR #133, opened 2026-04-14
Feat mysql ssl

traustitj:
New MySQL/MariaDB driver and connection pool with support for SSL certificates.
Settings tab for adding certificates for connection.


debba:
Looks good to me,
thanks a lot for your contribution.
Hope you will continue use Tabularis and improve it :)

PR #134, opened 2026-04-14
bug: Modify MySQL connection URL to use system timezone

thomaswasle:
In pool_manager.rs:43, the MySQL connection URL has no timezone parameter. The sqlx  MySQL driver defaults to setting time_zone='UTC' on every connection, so NOW() always returns UTC regardless of the server or user's local timezone.

The fix is to add timezone=SYSTEM to the connection URL, which tells MySQL to use the server's configured system timezone instead of being forced to UTC.

### DIFF
--- a/src-tauri/src/pool_manager.rs
+++ b/src-tauri/src/pool_manager.rs
@@ -52,7 +52,8 @@ fn build_mysql_options(
         .port(port)
         .username(username)
         .password(password)
-        .database(database);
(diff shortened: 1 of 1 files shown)


thomaswasle:
I updated my PR to fix a conflict after the main branch changed.

debba:
That's great, thanks a lot for this, and sorry I was working on code refactoring in builtin drivers

PR #154, opened 2026-04-28
fix: resolve double LIMIT bug in SQL pagination wrapping

midasism:
## Summary

- Fix a bug where `extract_order_by`/`remove_order_by` used naive `rfind("ORDER BY")` that captured trailing LIMIT/OFFSET clauses, producing **double LIMIT** syntax errors when the pagination layer appended its own LIMIT.
- Replace per-driver helpers with a shared, quote-and-bracket-aware SQL tokenizer (`extract_trailing_clauses` + `build_paginated_query`) in `drivers/common.rs` that correctly separates ORDER BY, LIMIT, and OFFSET from the base query.
- MCP `run_query` now accepts an optional `limit` parameter and respects user-specified LIMIT clauses (takes `min(user_limit, page_size)`).

## Bug details

When a user (or AI agent via MCP) ran a query like:
```sql
SELECT * FROM tapp_appointment_message_event_limit ORDER BY id LIMIT 10
```

The old code produced invalid SQL:
```sql
SELECT * FROM (SELECT * FROM tapp_appointment_message_event_limit) as data_wrapper ORDER BY id LIMIT 10 LIMIT 101 OFFSET 0
```

The double `LIMIT 10 LIMIT 101` is a syntax error on all database engines.

## Changes

| File | Change |
|------|--------|
| `src-tauri/src/drivers/common.rs` | Add shared `tokenize_sql`, `extract_trailing_clauses`, `build_paginated_query` + 18 unit tests |
| `src-tauri/src/drivers/postgres/mod.rs` | Remove old helpers, use shared functions |
| `src-tauri/src/drivers/mysql/mod.rs` | Same |
| `src-tauri/src/drivers/sqlite/mod.rs` | Same |
| `src-tauri/src/mcp/mod.rs` | Add optional `limit` param to `run_query`, use configurable max_rows |

## Test plan

- [x] 29 unit tests pass (`cargo test --lib drivers::common::tests`) covering:
  - Basic queries without trailing clauses
  - ORDER BY only / LIMIT only / both / all three
  - Table names containing "limit" substring (e.g. `tapp_appointment_message_event_limit`)
  - Quoted identifiers and string literals with SQL keywords
  - Subqueries with inner ORDER BY
  - Multi-column ORDER BY
  - Trailing semicolons
  - `build_paginated_query` with user limit smaller/larger than page_size
  - Pagination page 2+ offset calculation
  - SQLite (no alias) vs PG/MySQL (with alias) variants
- [x] Full `cargo test --lib` passes (115/117 pass; 2 pre-existing failures in `updater::tests` unrelated to this change)
- [x] `cargo build` succeeds with no warnings

### DIFF
--- a/src-tauri/src/drivers/common/query.rs
+++ b/src-tauri/src/drivers/common/query.rs
@@ -48,29 +48,192 @@ pub fn calculate_offset(page: u32, page_size: u32) -> u32 {
     (page - 1) * page_size
 }
 
+/// Simple SQL tokenizer that respects:
+/// - Single-quoted strings ('...')
+/// - Double-quoted identifiers ("...")
+/// - Backtick-quoted identifiers (`...`)
+/// - Parenthesized groups (treated as single tokens)
+/// - Whitespace as delimiter
+///
+/// This prevents keywords like LIMIT or OFFSET from being matched
+/// inside string literals, quoted identifiers, or table names such as
+/// `tapp_appointment_message_event_limit`.
+fn tokenize_sql(sql: &str) -> Vec<String> {
+    let mut tokens = Vec::new();
+    let chars: Vec<char> = sql.chars().collect();
+    let len = chars.len();
+    let mut i = 0;
+
+    while i < len {
+        if chars[i].is_whitespace() {
+            i += 1;
+            continue;
+        }
+
+        if chars[i] == '\'' {
+            let mut token = String::new();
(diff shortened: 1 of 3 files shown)


debba:
Thanks for PR, I will check it ASAP

debba:
i will merge it, thanks

PR #157, opened 2026-05-01
bug: changeed connections configuration so input fiels do not look prefilled if they are not

thomaswasle:
fix: make placeholder text visually distinct in the new connection modal

  Input fields in the new connection modal were indistinguishable between placeholder hints and
  actual configured values. Users reported that hint text like "root" in the username field
  appeared as if it were a saved value, causing confusion about what was actually stored in the
  connection.

  Changes

  - Italic placeholders — added placeholder:italic to all input fields in NewConnectionModal (host,
   port, username, password, database, connection string, SSL paths, SSH fields). Italic styling
  makes it immediately clear that the text is a hint, not a configured value.
  - Translated username placeholder — replaced the hardcoded "root" placeholder in the username
  field with a localized t("newConnection.usernamePlaceholder") key, consistent with how the
  password field already works.
  - i18n — added usernamePlaceholder to all 6 supported locales: en, de, es, fr, it, zh.

### DIFF
--- a/src/components/modals/NewConnectionModal.tsx
+++ b/src/components/modals/NewConnectionModal.tsx
@@ -102,7 +102,7 @@ const FieldInput = ({
       autoCapitalize="off"
       autoComplete="off"
       spellCheck={false}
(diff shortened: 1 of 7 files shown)


debba:
Looks wonderful! Thanks a lot for it!

PR #161, opened 2026-05-04
feat(data-grid): cell-level selection and copy

thomaswasle:
Summary

  Introduces two distinct, mutually exclusive selection modes in the data grid:

  Cell selection — click any data cell to focus it. The focused cell is highlighted with a blue
  inset ring. Ctrl/Cmd+C copies that cell's raw value as a string.

  Row selection — click the row-number column on the left to select one or more rows (existing
  shift-click range and Ctrl/Cmd-click toggle behaviour is unchanged). Ctrl/Cmd+C copies the
  selected rows in the configured format (CSV / JSON).

  The two modes clear each other: clicking a cell clears any row selection; clicking a row number
  clears cell focus. Clicking the # header (select all) also clears cell focus.

  A "Copy Cell" item has been added to the right-click context menu as a discoverable alternative
  to the keyboard shortcut.

  Changes

  - focusedCell state ({ rowIndex, colIndex } | null) tracks the active cell
  - copyCellValue(rowIndex, colIndex) formats the value via formatCellValue (handles null, boolean,
   JSON, geometry, BLOB) and writes it to the clipboard
  - copyCellFromContext wires the new context menu item to the same logic
  - Cell <td> click: sets focusedCell, clears row selection — no longer triggers handleRowClick
  - Row-number <td> click: clears focusedCell, then runs handleRowClick as before
  - handleSelectAll (the # header): clears focusedCell before selecting all rows
  - Focused cell renders with ring-2 ring-inset ring-blue-400 for a clear, layout-neutral visual
  indicator
  - Ctrl/Cmd+C handler simplified: focusedCell → copy cell; selectedRowIndices.size > 0 → copy rows
   (the two states are now mutually exclusive)

### DIFF
--- a/src/components/ui/DataGrid.tsx
+++ b/src/components/ui/DataGrid.tsx
@@ -172,6 +172,10 @@ export const DataGrid = React.memo(
     const [lastSelectedRowIndex, setLastSelectedRowIndex] = useState<
       number | null
     >(null);
+    const [focusedCell, setFocusedCell] = useState<{
+      rowIndex: number;
+      colIndex: number;
+    } | null>(null);
     const editInputRef = useRef<HTMLInputElement>(null);
 
     const selectedRowIndices =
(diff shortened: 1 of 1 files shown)


debba:
Interesting, I will test it ASAP

PR #173, opened 2026-05-09
fix(DataGrid): handle empty column names from drivers

saurabh500:
## Problem

Double-clicking a saved connection (or running certain queries) causes the editor window to go completely blank with no visible error. DevTools shows:

```
Uncaught Error: Columns require an id when using an accessorFn
    at createColumn (column.ts:120)
    at DataGrid.tsx:877
```

Reproducible on multiple engines:

- **SQL Server**: `SELECT @@VERSION` (returns a column with an empty name)
- **PostgreSQL**: `SELECT 1 AS """";`
- Likely any driver/query that yields an unnamed column

Because there is no top-level React error boundary, the throw in `<DataGrid>` unmounts the whole tree, producing the dark window.

## Fix

In `src/components/ui/DataGrid.tsx`, fall back to a synthetic `__unnamed_<index>__` id when `colName` is the empty string. The synthetic id is internal to react-table; the rendered header still shows the original (empty) name, and `info.column.id` consumers downstream (cell renderer, header context menu) keep working.

```ts
id: colName !== """" ? colName : `__unnamed_${index}__`,
```

## Testing

- Manually verified on the `feat/sql-server-complete` branch that `SELECT @@VERSION` no longer crashes the grid.
- Engine-agnostic — also protects PostgreSQL / MySQL / SQLite from the same class of crash.

## Follow-ups (not in this PR)

- Add a top-level React error boundary around the editor route so any future render crash doesn't blank the window.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -6154,7 +6154,7 @@ dependencies = [
 
 [[package]]
 name = "tabularis"
-version = "0.10.1"
+version = "0.10.2"
 dependencies = [
  "async-trait",
  "base64 0.22.1",
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -15,6 +15,7 @@ import { SchemaDiagramPage } from "./pages/SchemaDiagramPage";
 import { TaskManagerPage } from "./pages/TaskManagerPage";
 import { VisualExplainPage } from "./pages/VisualExplainPage";
 import { ConnectionHealthMonitor } from "./components/ConnectionHealthMonitor";
+import { EditorErrorBoundary } from "./components/ui/EditorErrorBoundary";
 import { UpdateNotificationModal } from "./components/modals/UpdateNotificationModal";
 import { CommunityModal } from "./components/modals/CommunityModal";
 import { WhatsNewModal } from "./components/modals/WhatsNewModal";
@@ -116,7 +117,14 @@ export function App() {
                         element={<Navigate to="/connections" replace />}
                       />
                       <Route path="connections" element={<Connections />} />
-                      <Route path="editor" element={<Editor />} />
+                      <Route
+                        path="editor"
+                        element={
+                          <EditorErrorBoundary>
+                            <Editor />
+                          </EditorErrorBoundary>
+                        }
+                      />
                       <Route path="mcp" element={<McpPage />} />
                       <Route path="settings" element={<Settings />} />
                     </Route>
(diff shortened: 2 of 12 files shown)


debba:
Nice catch, 
I will take in account the followup.
I can push directly on this branch, because related on main topic.

debba:
@saurabh500 

Added a React component for manage errors 
https://github.com/user-attachments/assets/e897e737-308f-490a-a79b-0dc4a424e848

debba:
Re: the follow-up item ("Add a top-level React error boundary around the editor route…"), I went ahead and prototyped it on my fork while reviewing this PR — happy to fold it in here or split it into a separate PR, whichever you prefer.

**What's in it (1 commit, ~350 LOC):**
- `src/components/ui/EditorErrorBoundary.tsx` — class boundary + hook-aware wrapper. Catches render errors thrown anywhere under the `/editor` route and shows a fallback (themed with project tokens) with three actions: **Try again** (resets the boundary), **Close current tab** (uses `useEditor().closeTab(activeTabId)` then resets — useful when the offending tab content is what's crashing), **Back to connections** (`navigate("/connections")`). Logs `[Editor] render crash caught by error boundary` with the component stack to `console.error`.
- `src/App.tsx` — wraps `<Editor />` in `<EditorErrorBoundary>` only for the `/editor` route.
- i18n keys `editor.errorBoundary.{title,description,retry,closeCurrentTab,backToConnections,showDetails}` added to all 6 locales (en/it/es/de/fr/zh).
- 7 unit tests covering happy path, fallback rendering, console logging, retry recovery, navigation, close-active-tab + reset, and the no-active-tab edge case. `tests/setup.ts` updated with the two new lucide icons used (`AlertTriangle`, `Home`).
- `pnpm tsc --noEmit` clean; `pnpm test` 2092/2092 ✅

Branch on my fork: `debba/tabularis:fix/datagrid-empty-column-id` (commit `3001409`, on top of your `b015b35`).

Let me know how you'd like to proceed — I can open a separate PR against `main` once this lands, or you can cherry-pick / pull the commit into this branch if you'd rather keep the fix and the boundary together.

saurabh500:
@debba  lets fold it into this PR, make it complete and then merge. 

I was not sure about whether I am on the right path with this PR, so I kept it in draft mode. But this can be useful for empty columns. Thanks for the collab

debba:
Can you merge it? Because I am not able to do it in your fork

saurabh500:
Done

saurabh500:
can you approve the test run workflow ?

debba:
After test run workflow, I will merge it

Issue #144, opened 2026-04-23
[Phase 2] Extend ConnectionParams with SQL Server-specific TLS/auth fields


debba:
# [Phase 2] Extend `ConnectionParams` with SQL Server-specific TLS/auth fields

Part of the SQL Server Phase 2 epic #150. Good first issue — localised change, no tricky async, clear test surface.

The driver isn't on `main`. Code lives on [`feat/sql-server`](https://github.com/TabularisDB/tabularis/tree/feat/sql-server) and isn't in any release. Check that branch out, branch off it, target your PR at it — not at `main`. The whole thing squashes into `main` when Phase 2 closes (#149).

## Task

Extend `src-tauri/src/models.rs::ConnectionParams` (struct at line ~111) with five new optional fields. All `Option<T>` with `#[serde(default, skip_serializing_if = "Option::is_none")]` so existing saved connections deserialize unchanged:

```rust
#[serde(default, skip_serializing_if = "Option::is_none")]
pub trust_server_certificate: Option<bool>,

#[serde(default, skip_serializing_if = "Option::is_none")]
pub domain: Option<String>,

#[serde(default, skip_serializing_if = "Option::is_none")]
pub instance_name: Option<String>,

#[serde(default, skip_serializing_if = "Option::is_none")]
pub encrypt: Option<String>,          // "yes" | "no" | "strict"

#[serde(default, skip_serializing_if = "Option::is_none")]
pub auth_mode: Option<String>,        // "sql" | "windows" | "azure-ad"
```

Then wire the relevant fields into `src-tauri/src/drivers/sqlserver/pool.rs::build_config`:

- `encrypt` → `EncryptionLevel::Required` (`"strict"`) / `NotSupported` (`"no"`) / `On` (default)
- `trust_server_certificate: Some(true)` → `cfg.trust_cert()`
- `instance_name: Some(name)` → `cfg.instance_name(&name)`
- `domain` and `auth_mode`: parse only — actual Windows / Azure AD wiring is Phase 3

## Rules

- Zero breaking changes: `cargo test --lib` must stay green (currently 471 tests)
- Add a unit test per new field in `pool.rs::tests` (extend the existing `build_config_*` matrix)
- Keep the Phase 1 default path working: when none of the new fields are set, `build_config` must produce exactly today's behaviour

## Reference

- Full technical plan: `docs/sql-server-implementation-plan.md` (§ Phase 2 — 2.1 + 2.2)
[...]

PR #174, opened 2026-05-09
refactor: derive Default on ConnectionParams and DatabaseSelection

saurabh500:
Precursor for #144. Add `Default` so tests can use `..Default::default()` as driver-specific fields land.

- `ConnectionParams`: derive Default
- `DatabaseSelection`: manual Default impl → `Single(String::new())`

490/490 lib tests pass. Zero changes needed outside `models.rs`.


saurabh500:
@debba the Default trait is ready

PR #176, opened 2026-05-09
feat: add connection export and import functionality

debba:
Splits #172 — second of two focused PRs.

Introduces export/import of connection configurations:

- `export_connections_payload` serializes connection groups, saved connections and SSH connections into a JSON payload, resolving keychain-stored passwords for both DB and SSH connections.
- `import_connections_payload` merges an `ExportPayload` with existing configurations and securely stores passwords in the system keychain, then persists the updated configurations to disk.
- New `ExportPayload` model with unit tests for (de)serialization.
- UI updates in `Connections.tsx` adding export/import buttons.
- Localization updates for export/import strings.
- `ConfirmModal` styling adjusted to support variants.

Co-authored-by: cole <imzhpe@qq.com>

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -10,9 +10,10 @@ use uuid::Uuid;
 use crate::credential_cache;
 use crate::keychain_utils;
 use crate::models::{
-    ColumnDefinition, ConnectionGroup, ConnectionParams, ConnectionsFile, ExplainPlan, ForeignKey,
-    Index, QueryResult, RoutineInfo, RoutineParameter, SavedConnection, SshConnection,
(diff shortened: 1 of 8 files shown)


debba:
@zhaopengme this is a very interesting feature, that's why I prefer to separate it in a different PR.
I will review it in next days.
Btw thanks a lot for your effort :)

debba:
Well done @zhaopengme .
I added the button for importing connections also in case of empty connections as a initial wizard.
I will merge it.

This is a really nice feature. If you'd like, I'd love to expand it further: for example by allowing the export of only selected connections and/or excluding passwords from the export for security reasons.
If you're interested, you could work on a phase 2 of this feature creating a new PR.
Thanks for your contributions.

PR #178, opened 2026-05-10
fix(notebook): render db selector dropdown via portal so it isn't clipped

ymadd:
## Summary
- The notebook SQL cell's database selector dropdown (introduced in #160) is rendered with `position: absolute` inside `NotebookCellWrapper`, which has `overflow-hidden`. As a result, when a connection manages many databases (e.g., a single MySQL host with dozens of schemas), the dropdown's `max-h-[230px]` scroll area is clipped by the cell's bounding rect — particularly noticeable on short cells (no result yet, collapsed neighbors, or the last cell in a notebook), where lower entries become unreachable and the scrollbar isn't visible.
- This PR renders the dropdown through a React portal (`createPortal` to `document.body`) with `position: fixed`, computed from the trigger button's `getBoundingClientRect()`. Position is also recomputed on `scroll` (capture) and `resize`, matching the existing pattern in `src/components/ui/Select.tsx`.
- Visual styling, `max-h-[230px] overflow-y-auto` behavior, backdrop click-to-close, and the show-only-when-`selectedDatabases.length > 1` condition are all preserved.

## Why
Without this change, `overflow-hidden` on the cell wrapper visually clips the absolutely-positioned dropdown regardless of its inner `overflow-y-auto`, defeating the scroll added in #160 whenever the cell is shorter than the dropdown.

## Test plan
- [ ] `pnpm typecheck` passes
- [ ] `pnpm tauri dev` — connect to a MySQL host with many databases (>12) and select multiple in the connection dialog
- [ ] Open a Notebook tab, add an SQL cell, click the database selector in the cell header
- [ ] Confirm the dropdown extends past the cell's visual bottom and scrolls to reach all databases
- [ ] Confirm position updates correctly when the notebook scroll container is scrolled and on window resize
- [ ] Confirm clicking outside / on the backdrop closes the dropdown

### DIFF
--- a/src/components/notebook/NotebookCellHeader.tsx
+++ b/src/components/notebook/NotebookCellHeader.tsx
@@ -1,4 +1,5 @@
-import { useState } from "react";
+import { useCallback, useEffect, useRef, useState } from "react";
+import { createPortal } from "react-dom";
 import { useTranslation } from "react-i18next";
 import {
   ChevronUp,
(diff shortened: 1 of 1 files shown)


debba:
Good catch, will merge it, thanks for your contribution.

PR #183, opened 2026-05-11
feat: database trigger management (PostgreSQL, MySQL, SQLite)

thomaswasle:
Summary

  Adds full trigger management across all three supported database engines — listing, viewing definitions, creating,
   editing (drop + recreate), and dropping triggers — with sidebar integration and a dedicated editor modal.

  Backend (Rust)

  - TriggerInfo model — new struct with name, table_name, event, timing, definition
  - DriverCapabilities — new triggers: bool feature flag; enabled for Postgres, MySQL, SQLite
  - Driver trait — four new default-Err methods: get_triggers, get_trigger_definition, create_trigger, drop_trigger
  - PostgreSQL — aggregates multi-event triggers via string_agg on information_schema.triggers; retrieves
  definitions via pg_get_triggerdef
  - MySQL — qualifies all trigger operations with the database name to correctly handle multi-database connections;
  uses sqlx::raw_sql for DDL (works around MySQL error 1295 — DDL not supported in prepared statement protocol)
  - SQLite — parses sqlite_master SQL text to extract timing/event metadata
  - Four new Tauri commands registered: get_triggers, get_trigger_definition, create_trigger, drop_trigger

  Frontend

  Sidebar
  - SidebarTriggerItem — new component with Zap icon, timing/event badge, table tooltip
  - Triggers accordion added to SidebarSchemaItem (PostgreSQL), SidebarDatabaseItem (MySQL multi-db), and the flat
  layout in ExplorerSidebar (MySQL single-db / SQLite)
  - Filter/search field in every triggers accordion (matches the existing table filter pattern)
  - Schema is correctly propagated through context menu data so all operations target the right database

  TriggerEditorModal
  - Two editing modes: Guided (name, table, timing toggles, event checkboxes, body editor, live SQL preview) and Raw
   SQL (full editor)
  - Opens in Guided mode by default when editing; Raw SQL tab always shows the original definition
  - Parses the loaded definition to populate Guided fields (timing, events, body extracted from FOR EACH ROW)
  - Driver-aware identifier quoting ("name" for Postgres/SQLite, `name` for MySQL)
  - MySQL: schema prefix omitted from the ON clause (MySQL resolves schema from the connection); connection pool is
  switched to the correct database before CREATE TRIGGER
  - Edit flow: warns user that drop + recreate is required, then executes atomically

  Editor
  - Tab.readOnly flag — hides Run and Explain Plan buttons for definition-view tabs
  - Persisted in cleanTabForStorage; patched retroactively via updateTab if an existing tab is re-opened via "View
  Definition"
  - "View Definition" passes the trigger's schema to the editor so the correct database is pre-selected in the
  schema switcher

  Graceful degradation

  All get_triggers calls use .catch(() => []) so plugin drivers that don't implement triggers don't break schema
  loading.


debba:
What a wonderful work @thomaswasle  :)
I'll review it asap

PR #200, opened 2026-05-14
fix(drivers): share a single connection across multi-statement scripts

### LINKED ISSUE #199: [Bug]: Multi-statement "Run All" loses connection-local session state (SET @var, LAST_INSERT_ID, BEGIN/COMMIT, TEMP TABLE) — silent data corruption
**Describe the bug**

When a multi-statement SQL script is executed via the editor's "run all queries" path, all statements are dispatched concurrently via `Promise.allSettled` rather than sequentially on a shared connection. Each `invoke("execute_query")` call acquires a separate connection from the MySQL pool, so statements that depend on connection-local state — `SET @var`, `LAST_INSERT_ID()`, transactions, `TEMPORARY TABLE`, `PREPARE` — silently break.

In practice this corrupts data without surfacing any error whenever a script depends on cross-statement session state.

**Root cause.** `src/pages/Editor.tsx#L823` (`runMultipleQueries`) wraps the per-statement `invoke("execute_query")` in `Promise.allSettled`. Each invoke lands in `src-tauri/src/drivers/mysql/mod.rs#L869` which calls `pool.acquire()` (pool size 10 — `src-tauri/src/pool_manager.rs#L279`). Session variables, `LAST_INSERT_ID()`, transactions, and temporary tables are connection-local on the server, so they cannot survive across these calls. The comment at `Editor.tsx#L805-806` notes JS is single-threaded between awaits — true for the JS array, but database-side state is held by physical connections that Tokio + sqlx happily interleave.

**Related prior fix.** v0.9.5 addressed a similar class of pool-reuse bug for multi-database connections (per-database pools keyed by connection ID + database name, so a connection used for `db_a` couldn't be reused for `db_b` and leak state). This issue is the intra-database analogue — same root cause family ("a freshly acquired pool connection has no relationship to the previous one"), different surface.

**Affected drivers.** The same pattern exists across all built-in drivers:

- `src-tauri/src/drivers/mysql/mod.rs` — `pool.acquire()` at L869, `affected_rows: 0` at L944.
- `src-tauri/src/drivers/postgres/mod.rs` — `pool.get()` at L734, `affected_rows: 0` at L803. PostgreSQL is affected by the same class of session-state breakage: `SET LOCAL`, `BEGIN; … COMMIT;`, `CREATE TEMP TABLE`, `LASTVAL()`/`currval()`, and `PREPARE` all rely on a stable connection.
- `src-tauri/src/drivers/sqlite/mod.rs` — `affected_rows: 0` at L575. Less affected by the session-state issue itself (different connection semantics), but the `affected_rows` problem applies.

**Concretely, what I observed (MySQL).** A transactional SQL file expected to insert 144 rows in a child-metadata table (8 per parent × 18 parent IDs) and 1008 rows in a related schedule table. After one run via Tabularis:

| symptom | observed |
|---|---|
| total child rows | 83 (instead of 144) |
| two parent IDs | 0 rows (entirely missing) |
| one parent's first child | inserted 5 times (duplicated) |
| two other parent IDs | 14 / 10 rows (random subsets with dupes) |
| auto-increment IDs across parents | interleaved (e.g. 3 consecutive IDs belonged to parent A / C / B) |

The exact same file ran cleanly when I executed each statement individually through the MCP `run_query` tool (one connection per call, but at least serialised) — confirming the script itself is correct.

**Affected workflows.** Any script relying on (1) `@var` / `SET LOCAL` session variables, (2) `LAST_INSERT_ID()` / `LASTVAL()` for parent/child inserts, (3) `BEGIN; … COMMIT;` transactions, (4) `TEMPORARY` / `TEMP` tables, (5) `PREPARE` / `EXECUTE`, (6) connection-level settings like `SET FOREIGN_KEY_CHECKS = 0` for bulk loads. These work in `mysql` CLI / `psql` / DBeaver / MySQL Workbench, but silently break in Tabularis.

**Side note (probably its own ticket).** `mysql::execute_query` at `src-tauri/src/drivers/mysql/mod.rs#L944` hardcodes `affected_rows: 0` in the return value — INSERT/UPDATE/DELETE never report an actual row count to the UI, which makes silent failures like this much harder to notice. The same hardcoding exists in the postgres (L803) and sqlite (L575) drivers. Happy to file separately.

I'd like to send a PR to fix this. Will follow up with proposed approaches as a comment so the issue body stays close to the template.

**To Reproduce**

Minimal reproduction against any MySQL 5.7 or 8.0 server:

1. Open the Tabularis editor on a MySQL connection.
2. Paste the following script (all statements at once):

   ```sql
   CREATE TABLE IF NOT EXISTS test_parent (id INT AUTO_INCREMENT PRIMARY KEY, name VARCHAR(50));
   CREATE TABLE IF NOT EXISTS test_child  (id INT AUTO_INCREMENT PRIMARY KEY, parent_id INT, name VARCHAR(50));
   INSERT INTO test_parent (name) VALUES ('A');
   SET @pid = LAST_INSERT_ID();
   INSERT INTO test_child (parent_id, name) VALUES (@pid, 'A-1');
   INSERT INTO test_child (parent_id, name) VALUES (@pid, 'A-2');
   ```

3. Trigger "run all" so the multi-query path is taken.
4. Inspect the result: `SELECT * FROM test_child;`
5. Observe: `parent_id` is NULL (or stale from a previous pool connection); children are not linked to the parent inserted seconds before.

   Expected: both child rows have `parent_id` equal to the newly inserted parent's `id`.

(An equivalent PostgreSQL repro using `BEGIN; CREATE TEMP TABLE …; INSERT …; SELECT * FROM …; COMMIT;` shows the same class of failure — the temp table is invisible to later statements that land on a different pooled connection.)

**OS Version**

macOS 26.4.1 (Tahoe) (Darwin 25.4.0)

**Tabularis Version**

v0.10.2

**Relevant Log Output**

\`\`\`
(No application error is surfaced; the failure mode is silent data corruption.
The Tauri layer reports success for each statement, and execute_query
hardcodes affected_rows: 0 across all three built-in drivers — so even
INSERT row counts cannot be used to detect that a statement failed to
apply as intended.)
\`\`\`

ymadd:
## Summary

Closes #199.

Multi-statement scripts run via the editor's "run all" path now share a single physical database connection across all statements, instead of fanning out via `Promise.allSettled` (each invoke acquiring its own pooled connection). Cross-statement session state — `SET @var`, `LAST_INSERT_ID()` / `LASTVAL()`, `BEGIN`/`COMMIT`, `TEMP TABLE`, `PREPARE`/`EXECUTE` — now survives the way it does in `mysql` CLI / `psql` / DBeaver.

Side fix: the three built-in drivers were hardcoding `affected_rows: 0`; they now report the real count from `execute()`.

## Approach

- **Trait**: new `execute_batch` method on `DatabaseDriver`, with a default implementation that sequentially calls `execute_query` (preserves backwards compat for plugin drivers — ordering only, no session-state continuity).
- **Built-in drivers**: MySQL / Postgres / SQLite override `execute_batch` to acquire one pooled connection (or `deadpool` client) and run every statement on it in order. Refactored `execute_query` to share `exec_on_*_conn` helpers with `execute_batch`.
- **MySQL transaction-control**: routed through `sqlx::raw_sql()` (text protocol) to bypass server error 1295 ("This command is not supported in the prepared statement protocol yet") for `BEGIN`, `START TRANSACTION`, `COMMIT`, `ROLLBACK`, `SAVEPOINT`, `RELEASE SAVEPOINT`, `LOCK TABLES`, `UNLOCK TABLES`.
- **Tauri command**: new `execute_query_batch`, single cancellation handle for the whole batch.
- **Frontend**: `runMultipleQueries` replaced with a single `invoke("execute_query_batch")`; misleading "JS is single-threaded between awaits" comment removed.
- **`affected_rows`**: new shared `returns_result_set` keyword classifier; non-result-set statements go through `execute()` and return the real `rows_affected`.

## Test plan

Integration tests (`cargo test --test integration_tests -- --ignored`, requires local MySQL on 33060 and Postgres on 54320 with credentials matching `get_*_params()`):

- [x] `test_mysql_batch_preserves_user_variable_and_last_insert_id` — Issue's exact repro
- [x] `test_mysql_batch_preserves_transaction_atomicity` — explicit BEGIN/COMMIT
- [x] `test_postgres_batch_preserves_temp_table_and_transaction` — TEMP TABLE visibility inside a transaction
- [x] `test_mysql_affected_rows_reported_correctly` — INSERT/UPDATE/DELETE row counts
- [x] `test_postgres_affected_rows_reported_correctly` — same on Postgres
- [x] Existing `test_mysql_integration_flow` / `test_postgres_integration_flow` — regression check
- [x] Manual UI verification on macOS 26.4.1 against MySQL 8.0: repro script writes 1 parent + 2 children with correct `parent_id`

7/7 integration tests pass; `cargo check` + `npm run typecheck` clean.

## Not in scope (follow-ups)

- `QueryOptions` struct to flatten the `(limit, page, schema)` parameter sprawl shared by `execute_query` / `execute_batch`
- Extract a `run_cancellable` helper to dedupe the cancellation-handle bookkeeping between `execute_query` and `execute_query_batch`
- Stream batch results via Tauri events instead of materialising the whole `Vec<BatchStatementResult>` (currently bounded by N × page_size rows)

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -10,9 +10,9 @@ use uuid::Uuid;
 use crate::credential_cache;
 use crate::keychain_utils;
 use crate::models::{
-    ColumnDefinition, ConnectionGroup, ConnectionParams, ConnectionsFile, ExplainPlan,
-    ExportPayload, ForeignKey, Index, QueryResult, RoutineInfo, RoutineParameter, SavedConnection,
-    SshConnection, SshConnectionInput, SshTestParams, TableColumn, TableInfo,
+    BatchStatementResult, ColumnDefinition, ConnectionGroup, ConnectionParams, ConnectionsFile,
+    ExplainPlan, ExportPayload, ForeignKey, Index, QueryResult, RoutineInfo, RoutineParameter,
+    SavedConnection, SshConnection, SshConnectionInput, SshTestParams, TableColumn, TableInfo,
     TestConnectionRequest, TriggerInfo,
 };
 use crate::persistence;
@@ -41,6 +41,19 @@ impl Default for QueryCancellationState {
             handles: Arc::new(Mutex::new(HashMap::new())),
         }
     }
+}
+
+/// Trims trailing semicolons and normalises Unicode smart quotes that some
+/// editors insert when the user pastes a query. Called on every query the
+/// UI hands off to a driver.
+fn sanitize_user_query(query: &str) -> String {
+    query
+        .trim()
+        .trim_end_matches(';')
+        .replace('\u{2018}', "'")
+        .replace('\u{2019}', "'")
+        .replace('\u{201C}', "\"")
+        .replace('\u{201D}', "\"")
 }
 
 // --- Persistence Helpers ---
@@ -2131,21 +2144,12 @@ pub async fn execute_query<R: Runtime>(
         query
     );
 
-    // 1. Sanitize Query (Ignore trailing semicolon + normalize smart quotes)
-    let sanitized_query = query
-        .trim()
-        .trim_end_matches(';')
-        .replace('\u{2018}', "'") // Left single quotation mark
-        .replace('\u{2019}', "'") // Right single quotation mark
-        .replace('\u{201C}', "\"") // Left double quotation mark
-        .replace('\u{201D}', "\"") // Right double quotation mark
-        .to_string();
+    let sanitized_query = sanitize_user_query(&query);
 
     let saved_conn = find_connection_by_id(&app, &connection_id)?;
     let expanded_params = expand_ssh_connection_params(&app, &saved_conn.params).await?;
     let params = resolve_connection_params_with_id(&expanded_params, &connection_id)?;
 
-    // 2. Spawn Cancellable Task
     let drv = driver_for(&saved_conn.params.driver).await?;
     let task = tokio::spawn(async move {
         drv.execute_query(
@@ -2158,22 +2162,19 @@ pub async fn execute_query<R: Runtime>(
         .await
     });
 
-    // 3. Register Abort Handle
     let abort_handle = task.abort_handle();
     {
         let mut handles = state.handles.lock().unwrap();
-        // If a query is already running for this connection, we overwrite the handle.
-        // Ideally we should cancel the previous one, but the UI should prevent double run.
+        // Overwrite any in-flight handle for this connection. The UI is
+        // responsible for preventing double-run; a leaked previous query
+        // is preferable to a missing abort handle for the new one.
         handles.insert(connection_id.clone(), abort_handle);
     }
 
-    // 4. Await & Handle Cancellation
     let result = task.await;
 
-    // 5. Cleanup
     {
         let mut handles = state.handles.lock().unwrap();
-        // Only remove if it matches (edge case: multiple queries, but connection_id is unique per tab usually)
         handles.remove(&connection_id);
     }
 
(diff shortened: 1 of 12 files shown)


debba:
Nice work on this. Sharing a single physical connection for `Run All` is a big improvement, and it fixes the class of scripts that rely on session state (`SET @var`, `LAST_INSERT_ID()`, temp tables, explicit transactions, etc.). The driver-side split between batch-level failures and per-statement results also looks like the right shape.

One edge case I am not fully sure about is pagination of a result produced inside `Run All`.

Example with MySQL, using a page size of `1`:

```sql
DROP TABLE IF EXISTS run_all_page_demo;

CREATE TABLE run_all_page_demo (
  id INT PRIMARY KEY,
  group_id INT,
  label VARCHAR(20)
);

INSERT INTO run_all_page_demo VALUES
  (1, 1, 'g1-a'),
  (2, 1, 'g1-b'),
  (3, 1, 'g1-c'),
  (4, 2, 'g2-a'),
  (5, 2, 'g2-b'),
  (6, 2, 'g2-c');

SET @group_id = 1;

SELECT id, group_id, label
FROM run_all_page_demo
WHERE group_id = @group_id
ORDER BY id;

SET @group_id = 2;
```

The first page of the `SELECT` is produced while `@group_id = 1`, so it correctly shows `g1-a`. But when the user asks for page 2, we currently re-run only the `SELECT` through the normal single-query pagination path. At that point the batch has already continued and changed `@group_id` to `2`.

So page 2 may be calculated from a different logical result set than page 1. If the same pooled connection is reused, it can page through group 2. If another connection is used, `@group_id` may be unset.

This makes me wonder whether `Run All` result sets should be treated as snapshots. In other words, the paginated pages should come from the result produced at the moment that statement ran, not from re-executing the SQL later.

Possible approaches:

- cache/spool the result set produced by the batch and page through that using a `result_set_id`;
- disable server-side pagination for batch results and show them as a capped snapshot;
- keep the current behavior but mark it as a known limitation.

ymadd:
Addressing it properly looks like a non-trivial implementation with design decisions worth discussing on their own. I'd suggest opening a separate issue to settle the approach there (I'll draft it with DBeaver's capped-snapshot behavior as the baseline and a cached-result-set alternative as a more ambitious option). Would you be open to merging #200 as-is with a "Known limitations" note added to the description? WDYT?

debba:
Yes absolutely, it's a very good work, thanks

PR #202, opened 2026-05-15
feat(i18n): add Japanese (ja) translation

ymadd:
## Summary
- Add `src/i18n/locales/ja.json` with a full Japanese translation covering every key in `en.json`.
- Register `ja` (`日本語`) in `src/i18n/config.ts` so it shows up in the language picker.
- Includes the recently added strings: trigger management UI (`sidebar.*`, `triggers.*`), connection export/import (`connections.export`, `connections.import`, `connections.exportWarning`), the editor "Accept suggestion with Enter" setting, and the Discord callout (`discordCallout.*`).

## Test plan
- [x] Build the app and switch the language to 日本語 in Settings → Appearance.
- [x] Walk through main flows (connections list, sidebar trigger CRUD, settings, AI assistant approval modal, Discord callout) and check that strings render in Japanese without broken interpolation.
- [ ] Confirm fallback works (any future string added in English will fall back gracefully until translated).


debba:
Looks good, I don't understand anything, so it's ok :)

PR #203, opened 2026-05-15
fix(commands): preserve all in-flight abort handles per connection

### LINKED ISSUE #201: [Bug]: QueryCancellationState overwrites in-flight abort handles when multiple execute_query calls share a connection_id
### Describe the bug

**Describe the bug**

  The Tauri command `execute_query` (`src-tauri/src/commands.rs`) registers its abort handle in `QueryCancellationState.handles`, a `HashMap<connection_id, AbortHandle>`. 
When two or more `execute_query` invocations are in flight against the same `connection_id`, each new
  `handles.insert(connection_id, ...)` silently overwrites the previous abort handle. A subsequent `cancel_query(connection_id)` only aborts the most-recently-registered statement; the earlier Tokio tasks keep running, still holding their pooled connection.

  PR #200 makes the "run all" path immune to this by registering one handle per batch, but the underlying state shape is unchanged and several call
  sites still invoke `execute_query` directly (sidebar, notebook, modals, the per-query path in `Editor.tsx`).

  **Root cause**

  ```rust
  let mut handles = state.handles.lock().unwrap();
  handles.insert(connection_id.clone(), abort_handle);
  ```

  **Proposed fix**

  Change `QueryCancellationState.handles` to `HashMap<String, Vec<AbortHandle>>`:

  - on register: push the new handle and `retain(|h| !h.is_finished())` to keep the `Vec` bounded,
  - on `cancel_query`: drain the `Vec` and call `abort()` on every handle.

  Preserves the existing connection-level cancel semantics and the Tauri command signatures (frontend untouched). Per-statement cancel can be layered
  later by widening the inner type if needed.

### To Reproduce

  A focused `tokio::test` that spawns two `execute_query`s concurrently against the same `connection_id`, calls `cancel_query`, and asserts both join with the cancelled outcome. (Hard to trigger reliably through the default UI today — flows tend not to fire two simultaneous invocations on the same tab.)

  Discovered while working on #199 / #200. I'll send a PR shortly.

### OS Version

macOS 26.4.1 (Tahoe) (Darwin 25.4.0)

### Tabularis Version

v0.10.2

### Relevant Log Output

```shell

```

ymadd:
## Summary

`QueryCancellationState` stored one `AbortHandle` per `connection_id`, so the second `execute_query` / `execute_query_batch` / `explain_query_plan` against the same connection overwrote the previous handle. The earlier Tokio task kept running on its pooled connection, and `cancel_query` could only stop the most recently registered one.

This PR switches the slot type to `Vec<Arc<AbortHandle>>` and updates the three register sites + `cancel_query` accordingly.

- `register_abort_handle` pushes the new handle and prunes finished ones so the Vec stays bounded.
- `unregister_abort_handle` removes the specific handle by `Arc` identity, so a completing task never clears a sibling's handle.
- `cancel_query` drains the slot and aborts every handle.

Closes #201

Builds on #200 (only the new behaviour is exercised on top of the multi-statement batch path).

## Test plan

- [x] `cargo test --lib commands::tests::cancellation_state` — 5 new unit tests covering push, drain-and-abort, identity-based unregister, empty-slot cleanup, and finished-handle pruning.
- [x] `cargo test --test integration_tests -- --ignored --test-threads=1` — adds `test_concurrent_cancel_aborts_all_in_flight_queries` (two `SELECT SLEEP(5)` against the same connection, single cancel, both tasks report `JoinError::is_cancelled()`).
- [x] All existing integration tests still pass.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -31,12 +31,17 @@ async fn driver_for(
 const DEFAULT_MYSQL_PORT: u16 = 3306;
 const DEFAULT_POSTGRES_PORT: u16 = 5432;
 
+/// Per-slot collection of abort handles for in-flight cancellable tasks.
+/// Used by `QueryCancellationState`, `ExportCancellationState`, and
+/// `DumpCancellationState`.
+pub(crate) type AbortHandleMap = HashMap<String, Vec<Arc<AbortHandle>>>;
+
 /// Tracks abort handles for in-flight queries keyed by connection id. A
 /// slot can hold multiple handles when the UI fires several queries (or
 /// an EXPLAIN alongside a query) against the same connection concurrently
 /// — `cancel_query` must abort all of them, not just the most recent.
 pub struct QueryCancellationState {
-    pub handles: Arc<Mutex<HashMap<String, Vec<Arc<AbortHandle>>>>>,
+    pub handles: Arc<Mutex<AbortHandleMap>>,
 }
 
 impl Default for QueryCancellationState {
@@ -51,7 +56,7 @@ impl Default for QueryCancellationState {
 /// have already finished so the Vec does not grow unboundedly across many
 /// sequential queries on the same connection.
 pub(crate) fn register_abort_handle(
-    handles: &Mutex<HashMap<String, Vec<Arc<AbortHandle>>>>,
+    handles: &Mutex<AbortHandleMap>,
     key: String,
     handle: Arc<AbortHandle>,
 ) {
@@ -65,7 +70,7 @@ pub(crate) fn register_abort_handle(
 /// task registered, so it cannot fire on a future query that happens to
 /// reuse the same slot.
 pub(crate) fn unregister_abort_handle(
-    handles: &Mutex<HashMap<String, Vec<Arc<AbortHandle>>>>,
+    handles: &Mutex<AbortHandleMap>,
     key: &str,
     handle: &Arc<AbortHandle>,
 ) {
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -10,9 +10,9 @@ use uuid::Uuid;
 use crate::credential_cache;
 use crate::keychain_utils;
 use crate::models::{
-    ColumnDefinition, ConnectionGroup, ConnectionParams, ConnectionsFile, ExplainPlan,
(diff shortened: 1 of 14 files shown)


debba:
@ymadd thanks for the PR!

I pushed a small follow-up on top: EXPLAIN and regular queries now share a single `QueryCancellationState` slot keyed by `connection_id`. The `explain_` prefix was a workaround for the old single-handle-per-key state and is no longer needed; as a side effect, `cancel_query` now actually cancels in-flight EXPLAINs.

`cancel_query`'s body is extracted into `cancel_query_impl` so tests can call it directly, with coverage for the empty-slot error, multi-handle abort, and the query + EXPLAIN case.

What do you think?

ymadd:
Nice refactor. Unifying the slot key by `connection_id` is the right call. The `explain_` prefix was only there because the old state could only hold one handle per key, and the side effect of `cancel_query` now actually aborting in-flight EXPLAINs is a real bug fix on top of #201.

Extracting `cancel_query_impl` and the three new tests are exactly the shape I'd want, happy to keep both as the canonical pattern.

I had one more local commit on top of #203 that applies the same per-slot fix to `ExportCancellationState` (`cancel_export` / `export_query_to_file`) and `DumpCancellationState` (`cancel_dump` / `dump_database` / `cancel_import` / `import_database`), reusing the helpers in `commands.rs`. It only touches `export.rs` and `dump_commands.rs`, so it rebased cleanly on top of your change and I just pushed it. Let me know if you'd prefer that split out into a separate PR instead.

debba:
That's great.
Sure you can go ahead in this PR

ymadd:
Thanks for the green light. Pushed one more small cleanup on top:

- `pub(crate) type AbortHandleMap = HashMap<String, Vec<Arc<AbortHandle>>>` in `commands.rs`, used in the three cancellation state structs and in the helper signatures. The 5-level generic was getting noisy, and the alias lets `export.rs` and `dump_commands.rs` drop their direct `tokio::task::AbortHandle` / `std::collections::HashMap` imports.
- `fn import_slot_key(&str) -> String` in `dump_commands.rs` so the `"{}_import"` key composition lives in one place instead of being inlined at both `cancel_import` and `import_database`.

No behavior change, all 8 `cancellation_state` unit tests still pass.

That's all the changes from my side. Branch is ready for review/merge whenever you have a chance.

PR #206, opened 2026-05-17
ci: support manual prereleases built from fork PRs

NewtTheWolf:
Adds workflow_dispatch inputs (pr_number, tag_name, prerelease) to the release workflow so a maintainer can build a prerelease directly from a fork PR without merging or pushing the branch upstream. When pr_number is set, checkout uses refs/pull/<N>/head; the dispatcher controls the tag and prerelease flag.

Tag names containing a "-" (semver pre-release qualifier) auto-flag the release as prerelease even without the explicit input. AUR, Snap, and Winget downstream workflows now skip when github.event.release.prerelease is true, so beta channels stay out of the system package managers.


.github/workflows/release.yml
```diff
@@ -6,6 +6,17 @@ on:
...
+      pr_number:
+        description: "PR number to build (from any fork). Leave empty to build current ref."
+        required: false
+      tag_name:
+        description: "Tag/release name (e.g. v0.10.0-json-beta.1). Required when pr_number is set."
+        required: false
+      prerelease:
+        description: "Mark as prerelease (skips AUR/Snap/Winget downstream)."
+        type: boolean
+        default: false
```
saurabh500: Should we assume that forks will always be pre-release and not provide the option to toggle. This would make sure that the release bits are created from main source and not the fork.

NewtTheWolf: yes i think it is better

NewtTheWolf: Implemented in 40c060c — removed the `prerelease` workflow input entirely and hard-coded `prerelease=true` on the `pr_number` branch. Fork PR builds can never be promoted to a full release anymore. Tag-name SemVer auto-detection still applies for tag-triggered / main-built dispatches.

PR #167, opened 2026-05-06
fix(postgres): switch deadpool TLS to rustls + honor ssl_ca (closes #166)

benedettoraviotta:
## Summary

Closes #166. The Postgres deadpool path used `postgres-native-tls`. On macOS this triggered two stacked failures against AWS RDS endpoints — first conn ok, every reconnect/health-check ping fails:

1. Secure Transport applies a strict `id-kp-serverAuth` EKU check to user-supplied root anchors → "The extended key usage is not valid" when the user pastes the AWS RDS bundle.
2. The macOS keychain does not trust the regional Amazon RDS root CAs out of the box → platform verification fails with `errSecNotTrusted` (-67843) even without a custom CA.

`tokio_postgres::Error::Display` only prints "error performing TLS handshake", which hid the real cause. Walking `source()` made both surface.

## Change

- Replace `postgres-native-tls` with `tokio-postgres-rustls` for the **Postgres deadpool pool only**. MySQL pool and sqlx keep `native-tls`.
- `rustls-platform-verifier` provides default trust without the Secure Transport EKU quirk.
- Honor `params.ssl_ca` (PEM file/bundle) so users pin RDS or any custom CA from the connection's "CA Certificate" field.
- Surface the full `source()` chain on ping/pool failures.

## Why no bundled RDS CA

We deliberately do **not** vendor `global-bundle.pem`:
- AWS rotates these CAs every 1-3 years; a vendored bundle silently breaks RDS users on released app versions until a new build ships.
- Repo bloat (~165 KB binary, 108 PEM blocks).

Distributors who want out-of-the-box RDS support can pull the bundle at packaging time:

```dockerfile
# Dockerfile
RUN curl -fsSLo /app/src-tauri/assets/rds-global-bundle.pem \\
    https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem
```

```bash
# build script
mkdir -p src-tauri/assets
curl -fsSLo src-tauri/assets/rds-global-bundle.pem \\
    https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem
```

## End-user workflow for RDS

1. Download \`https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem\`.
2. Connection edit → SSL → CA Certificate = path to that file.
3. Save.

## Test plan

- [x] \`cargo test --lib\` — all relevant tests pass; 2 pre-existing flakes in \`updater::tests\` (\`test_detect_installation_source_{flatpak,snap}\`) verified on \`main\` too.
[...]

PR #211, opened 2026-05-18
feat(postgresql)/issue 209 verify ca and verify full

### LINKED ISSUE #209: [Feat]: Align SslMode behavior with libpq by adding verify-ca and verify-full modes
### Prerequisites

### 🚀 Feature Proposal

Currently, the `ssl-mode` settings in Tabularis (when connecting to PostgreSQL) only support `disable`, `allow`, `prefer`, and `require`. However, the actual behavior observed is that **all modes (including `require`) seem to demand a valid server certificate**, causing connections to fail unless a CA certificate is provided.

This is inconsistent with the standard `libpq` behavior (used by `psql`, DBeaver, etc.), where:
- `require` only forces encryption but **does not require certificate validation** unless a CA file is explicitly configured.
- `verify-ca` and `verify-full` are the modes that actually validate the server certificate.

I propose to **add `verify-ca` and `verify-full` modes** and **align `require` with libpq semantics** (i.e., `require` should not demand certificate validation).

### Motivation

The current behavior breaks compatibility with existing PostgreSQL deployments that expect the standard SSL mode semantics. For example:
- Many cloud and self-hosted Postgres instances (e.g., AWS RDS with self-signed certificates) work fine with `sslmode=require` in other clients, but fail in Tabularis because it wrongly demands a CA certificate.
- Users coming from other database tools are confused and forced to work around the issue by disabling SSL entirely or manually extracting CA certificates, which is inconvenient and security‑sensitive.

Aligning with `libpq` will:
- Reduce surprise and adoption friction.
- Allow Tabularis to connect securely to a wider range of Postgres servers without unnecessary certificate gymnastics.
- Provide a clear security progression: `require` (encryption only) → `verify-ca` (validate CA) → `verify-full` (validate CA + hostname).

### Example

In the connection configuration UI or connection string, users should be able to select/use:

- `disable` – no encryption
- `allow` – try non‑SSL first
- `prefer` – try SSL first, fallback to non‑SSL
- `require` – **force encryption, but do NOT require certificate validation**
- `verify-ca` – force encryption + validate server CA
- `verify-full` – force encryption + validate CA + verify hostname

For a typical RDS connection that uses a self‑signed certificate (but the user trusts the server), the working setting would be:
ssl-mode=require
```text
Without any `sslrootcert` file. Currently this fails; after the change it would succeed.

The underlying `tokio-postgres` driver currently lacks `verify-ca`/`verify-full`. If needed, we can either:
- Switch to a different driver (e.g., `sqlx`) that already supports all modes, or
- Contribute to `tokio-postgres` to add the missing modes.

But this issue focuses on Tabularis’s observable behavior – it should match what users expect from standard PostgreSQL clients.
```

VincentZhangy:
Extend the SSL mode support in libpq to include 'verify-ca' and 'verify-full',
bringing it in line with PostgreSQL's libpq behavior.

- verify-ca: verify server certificate is signed by a trusted CA
- verify-full: additionally verify the server hostname matches the certificate CN or SAN

Fix issues in the existing SSL implementation where only 'require' mode was
properly handled. Refactor the SSL context configuration to correctly set
certificate verification flags and hostname validation callbacks.

Optimize the certificate verification logic by:
- Reusing the same verification callback for both modes
- Reducing redundant certificate chain traversals
- Improving error messages for certificate validation failures

This change ensures secure TLS connections with proper certificate checks,
preventing man-in-the-middle attacks when using the stricter SSL modes.

close #209


ymadd:
Thanks for tackling #209! I'm not a maintainer — just a user who read through the diff out of interest. Sharing a few observations below in case they're useful to you or the maintainers. Take or leave any of these.

## Things that stood out

### 1. `webpki` (rustls-webpki 0.101) looks unused
`src-tauri/Cargo.toml` adds `webpki = { package = "rustls-webpki", version = "0.101" }`, but I couldn't find a `webpki::` reference anywhere in `src-tauri/src/`. As a side effect, `Cargo.lock` now carries both `rustls-webpki 0.101.7` and `0.103.9`. Might be worth dropping if it's truly unused.

### 2. The `verify-ca` fallback in `VerifyCaCertVerifier::verify_server_cert` is hard to reason about
`pool_manager.rs:382-408`:

```rust
Err(e) if is_hostname_error(&e) => {
    self.inner.verify_server_cert(
        end_entity, intermediates,
        &ServerName::try_from("localhost").unwrap(),
        ocsp_response, now,
    )
    .or(Ok(ServerCertVerified::assertion()))
}
```

A couple of concerns from a reader's perspective:
- `.or(Ok(ServerCertVerified::assertion()))` converts any error from the second call into success. In practice the chain was already validated by the first call so this is probably benign today, but it relies on a specific internal ordering of `WebPkiServerVerifier`. If that ever changes, an unverified chain could slip through silently.
- The `"localhost"` placeholder + a second full verification means the chain gets validated twice per handshake.

`rustls 0.23` exposes `rustls::client::verify_server_cert_signed_by_trust_anchor` and `rustls::client::verify_server_name` as public API (see `rustls-0.23.40/src/lib.rs:615-618`). A more direct shape might be:

```rust
struct VerifyCaCertVerifier {
    roots: Arc<RootCertStore>,
    supported: WebPkiSupportedAlgorithms,
}

impl ServerCertVerifier for VerifyCaCertVerifier {
    fn verify_server_cert(&self, end_entity, intermediates, _server_name, _ocsp, now)
        -> Result<ServerCertVerified, TlsError>
    {
        let cert = ParsedCertificate::try_from(end_entity)?;
        verify_server_cert_signed_by_trust_anchor(&cert, &self.roots, intermediates, now, self.supported.all)?;
        Ok(ServerCertVerified::assertion())
    }
    // verify_tls{12,13}_signature delegate to rustls::crypto::verify_tls{12,13}_signature
}
```

This makes "we intentionally skip hostname verification" explicit, avoids the `.or(Ok(...))`, and only validates the chain once.

### 3. `WebPkiServerVerifier::builder(...).build().expect(...)` can panic
`pool_manager.rs:367`. `builder().build()` returns `Err(VerifierBuilderError::NoRootAnchors)` for an empty `RootCertStore`, and `load_platform_roots()` can return empty on macOS if `rustls_native_certs::load_native_certs()` partially fails. Propagating via `Result` would avoid a hard panic.

### 4. The `verify-ca` default path may regress what PR #167 fixed on macOS
`load_platform_roots()` uses `rustls_native_certs::load_native_certs()` and feeds them straight into `WebPkiServerVerifier`. PR #167 explicitly moved away from this pattern because of two macOS-specific problems:
- Secure Transport's strict `id-kp-serverAuth` EKU check on user-supplied root anchors rejects valid CA certs (including the AWS RDS bundle).
- The macOS keychain doesn't trust regional Amazon RDS roots out of the box.

The existing docstring at `pool_manager.rs:188-214` mentions this design decision. With this change, `verify-ca` without a user CA might fail for RDS users on macOS who currently work with the `rustls-platform-verifier` path. One option could be to keep `rustls-platform-verifier` for the no-user-CA branch of `verify-ca`, or to require an explicit CA file for `verify-ca` (which is effectively how libpq is used anyway). Either way the docstring would want a refresh.

### 5. `require` semantics change is a behavior shift worth flagging
Pre-PR `require` did strict verification (PR #167 made that intentional); post-PR `require` skips verification (which matches libpq). That's the right direction IMO, but users who currently rely on `require + CA bundle` on RDS will silently lose verification on upgrade. A CHANGELOG note pointing them at `verify-full` (or `verify-ca`) might save some confusion.

---

A few smaller notes I'll keep out of this comment unless useful: only `en` + `zh` of the 7 locales got `verify-ca` / `verify-full` strings; `CryptoProvider::get_default().unwrap()` is called per-handshake inside `NoCertVerifier::supported_verify_schemes`; the `test_tls_connector_*` tests assert `is_ok()` rather than verifier semantics (a `NoCertVerifier` in every branch would still pass them).

Hope some of this helps — and thanks again for the work on this.

VincentZhangy:
@ymadd Thank you again for the extremely thorough review – it was incredibly helpful.  
I’ve gone through every point you raised, and most of them have been addressed exactly as you suggested. Below is a quick summary of the fixes already in place, followed by one small remaining issue I spotted while double‑checking the code.

## What has been fixed (matching your feedback)

1. **Unused `webpki 0.101`** – removed from `Cargo.toml`; `Cargo.lock` now only carries `rustls-webpki 0.103.x`. ✅  
2. **`VerifyCaCertVerifier` rewritten** – now uses `verify_server_cert_signed_by_trust_anchor` directly, drops the `"localhost"` placeholder, validates the chain once, and makes the “skip hostname check” intent explicit. No more `.or(Ok(...))`. ✅  
3. **No more panics** – `VerifyCaCertVerifier::new()` returns `Result<Self, String>` and errors propagate through `build_postgres_tls_connector`. `WebPkiServerVerifier::builder().build()` uses `map_err` with `?` instead of `expect`. ✅  
4. **`verify-ca` now requires an explicit CA file** – we no longer fall back to platform roots, matching libpq’s `sslmode=verify-ca` and avoiding macOS Secure Transport EKU issues. The docstring has been updated accordingly. ✅  
5. **`require` semantics change** – documented in `CHANGELOG` as a breaking change, pointing users toward `verify-ca`/`verify-full` for validation. ✅  
6. **Smaller notes** – i18n strings added for all five missing locales, `NoCertVerifier` caches `WebPkiSupportedAlgorithms` at construction, dead code (`is_hostname_error`, `load_platform_roots`) removed, tests updated to cover the new `verify-ca` behaviour. ✅  

## One minor issue I noticed while reviewing the code

In `load_roots_from_pem` (pool_manager.rs), the current implementation writes:

```rust
let mut cursor = std::io::Cursor::new(&pem[..]);
for cert in rustls_pemfile::certs(&mut cursor) {
    let cert = cert.map_err(...)?;
    ...
}
```

However, `rustls_pemfile::certs` returns a `Result<impl Iterator<...>, ...>`. The `for` loop cannot iterate directly over a `Result` – the function call should be followed by `?` and then the iterator should be used, e.g.:

```rust
let cert_iter = rustls_pemfile::certs(&mut cursor)
    .map_err(|e| format!("Failed to parse ssl_ca '{}': {}", path, e))?;
for cert in cert_iter {
    let cert = cert.map_err(|e| format!("..."))?;
    roots.add(cert).map_err(|e| format!("..."))?;
}
```

I believe this is just a copy‑paste oversight in the snippet you shared – the actual code in your repository may already be correct. Could you double‑check it? Apart from that, everything else looks solid.

## Final note

Your review was exceptionally detailed and saved us from several subtle pitfalls. Thank you again for taking the time to write it up. If you spot anything else, please don’t hesitate to point it out – I’ll be happy to follow up.

Best regards

ymadd:
@VincentZhangy Thanks for the thorough revisions! All the Critical items look cleanly addressed, and the rewrite of `VerifyCaCertVerifier` on top of `verify_server_cert_signed_by_trust_anchor` is exactly the right shape. Nicely done.

debba:
Looks a very good work!
Thanks @VincentZhangy and @ymadd  for the revision :)
I will take a look and if no problems I will merge it asap.

debba:
Hi @VincentZhangy,

Thanks for the detailed write-up — the changes look good to me, everything you listed lines up with the diff.

About the `load_roots_from_pem` snippet: if I'm not mistaken, the code in the repo is actually fine. The signature you're describing (`Result<impl Iterator<...>, ...>`) is the old `rustls-pemfile` 1.x API. Since 2.0, `certs()` was reworked to return the iterator directly:

```rust
pub fn certs<R: BufRead>(rd: &mut R)
    -> impl Iterator<Item = Result<CertificateDer<'static>, io::Error>>
```

So each item yielded by the loop is already a `Result`, and the inner `cert.map_err(...)?` is the idiomatic way to handle it. Cargo.toml pins `rustls-pemfile = "2"` (currently resolving to 2.2.0), which is why the current code compiles and CI is green. Applying `?` to the `certs()` call itself would actually fail to compile against 2.x, since the return value isn't a `Result` anymore.

@ymadd could you confirm this on your side as well? If we're aligned, I'd say no further changes are needed and we can move toward merging.

By the way, if you'd like to chat with the maintainers or get more involved with the project, feel free to join our Discord: https://discord.com/invite/K2hmhfHRSt — would be great to have you there.

ymadd:
Confirmed on my side as well — the current code is correct.

  I checked the local rustls-pemfile 2.2.0 source and the signature is:
```rust
pub fn certs(rd: &mut dyn io::BufRead)
    -> impl Iterator<Item = Result<CertificateDer<'static>, io::Error>>
```
So the iterator is returned directly and each item is a `Result`. The existing
`for cert in rustls_pemfile::certs(&mut cursor) { let cert = cert.map_err(...)?; ... }`
pattern is the idiomatic shape against 2.x. Applying `?` to `certs()` itself would actually break against this version, as @debba pointed out. 
The pattern @VincentZhangy was thinking of is the rustls-pemfile 1.x API.

No further changes needed from my side — this looks good to merge.

Issue #96, opened 2026-03-17
[Bug]: Does not pull FK depending on rights in postgre


Mazwak:
### Describe the bug

I have a Postgre DB with a read-only user for most queries.

If I open it with this user, I don’t see the FK.
If I open it with postgre, I see them.

### To Reproduce

Connect to postgre with a realonly user.

User setup with:
GRANT USAGE ON SCHEMA your_schema TO readonly_user;
GRANT SELECT ON ALL TABLES IN SCHEMA your_schema TO readonly_user;

Open schema with tabularis, no FK are showed.

### OS Version

Windows 10

### Tabularis Version

0.9.9

### Relevant Log Output

```shell

```

Mazwak:
Your query is (from mitm):
SELECT c.column_name, c.data_type, c.is_nullable, c.column_default, c.is_identity, c.character_maximum_length, (SELECT COUNT(*) FROM information_schema.table_constraints tc JOIN information_schema.key_column_usage kcu ON tc.constraint_name = kcu.constraint_name AND tc.table_schema = kcu.table_schema WHERE tc.constraint_type = 'PRIMARY KEY' AND tc.table_schema = c.table_schema AND kcu.table_name = c.table_name AND kcu.column_name = c.column_name) > 0 as is_pk FROM information_schema.columns c WHERE c.table_schema = $1 AND c.table_name = $2 ORDER BY c.ordinal_position

But I have 0 rows in information_schema.table_constraints with the read-only user.
I don’t know why.

But I do have access to pg_constraint (and some other tables), so the following does work:

```
SELECT
    con.conname AS constraint_name,
    nsp.nspname AS source_schema,
    rel.relname AS source_table,
    att.attname AS source_column,
    fnsp.nspname AS target_schema,
    frel.relname AS target_table,
    fatt.attname AS target_column
FROM pg_constraint con
JOIN pg_class rel ON rel.oid = con.conrelid
JOIN pg_namespace nsp ON nsp.oid = rel.relnamespace
JOIN unnest(con.conkey) WITH ORDINALITY AS cols(attnum, ord) ON TRUE
JOIN pg_attribute att ON att.attrelid = rel.oid AND att.attnum = cols.attnum
JOIN pg_class frel ON frel.oid = con.confrelid
JOIN pg_namespace fnsp ON fnsp.oid = frel.relnamespace
JOIN unnest(con.confkey) WITH ORDINALITY AS fcols(attnum, ord) ON fcols.ord = cols.ord
[...]

Issue #95, opened 2026-03-16
[Feat]: Add support for DB2


ahmednrana:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Although the client is promising but the number of database it can connect to is limited.
Please add support for DB2.

### Motivation

It is heavily used in corporate apps.

debba:
Hi @ahmednrana , 
I started working on a db2 integration, is it already on plugin registry and available in Tabularis, but consider it as a work in progress.
Never had experience with DB2 so If you want contribute or you want invite other developers is more than welcome.

Repo for reference:

https://github.com/debba/tabularis-db2-plugin

PR #212, opened 2026-05-18
refactor(sqlserver): use bridge preview.5 Value ToSql, drop value_to_sql_param workaround

### LINKED ISSUE #145: [Phase 2] CRUD composite primary keys (trait default methods)
# [Phase 2] CRUD composite primary keys (trait default methods)

Part of the SQL Server Phase 2 epic #150.

The driver isn't on `main`. Code lives on [`feat/sql-server`](https://github.com/TabularisDB/tabularis/tree/feat/sql-server) and isn't in any release. Check that branch out, branch off it, target your PR at it — not at `main`. The whole thing squashes into `main` when Phase 2 closes (#149).

Phase 1 CRUD commands in `src-tauri/src/commands.rs` (`update_record`, `delete_record` around lines 1788–1855) assume a single PK column. SQL Server tables routinely have composite PKs — the driver needs a way to target rows by `(col1, col2, ...)` without breaking MySQL / Postgres / SQLite.

## Task

**1. Trait evolution (backward-compatible).** In `src-tauri/src/drivers/driver_trait.rs`, add two default methods to `DatabaseDriver`:

```rust
async fn delete_record_composite(
    &self,
    params: &ConnectionParams,
    table: &str,
    pk_cols: &[String],
    pk_vals: Vec<serde_json::Value>,
    schema: Option<&str>,
) -> Result<u64, String> {
    if pk_cols.len() == 1 {
        let v = pk_vals.into_iter().next().unwrap_or(serde_json::Value::Null);
        return self.delete_record(params, table, &pk_cols[0], v, schema).await;
    }
    Err("Composite primary keys not supported by this driver".into())
}

// Same pattern for update_record_composite (adds col_name / new_val / max_blob_size).
```

The default forwards to the legacy single-key method when `pk_cols.len() == 1`, which keeps MySQL / Postgres / SQLite drivers working without any change.

**2. SQL Server override.** In `src-tauri/src/drivers/sqlserver/mod.rs`, override both methods to build `WHERE [c1]=@P1 AND [c2]=@P2 AND ...` using tiberius parameter binding (ordinal markers).

**3. Tauri command extension.** In `src-tauri/src/commands.rs`, extend `update_record` / `delete_record` with optional parallel parameters:

```rust
pk_cols: Option<Vec<String>>,
pk_vals: Option<Vec<serde_json::Value>>,
```

If present → call the `*_composite` method. Otherwise → legacy path. Tauri deserializes missing JSON keys as `None`, so existing frontend callers are unaffected.

## Rules

- Existing frontend CRUD flows (MySQL / Postgres / SQLite) keep working byte-identically — verify with regression test
- Unit tests for the `*_composite` methods: one covering the single-key fallback, one covering a real composite `WHERE` clause (can use mock params for the `sql_statement_for_...` builders)
- No breaking changes to existing trait signatures — only new default methods

## Reference

- Plan: `docs/sql-server-implementation-plan.md` § Phase 2 — 2.3 + 2.4
- Existing assumption in MySQL: `src-tauri/src/drivers/mysql/mod.rs::update_record` uses a single `pk_col: &str`

saurabh500:
## Summary

Upgrades `mssql-tiberius-bridge` from `preview.3` to `preview.5` and removes the `value_to_sql_param()` workaround that was added in the issue #145 composite PK implementation.

## Background

The bridge's `impl ToSql for serde_json::Value` previously called `self.to_string()` on every variant, which:
- JSON-quoted strings: `Value::String("alice")` was bound as `"\"alice\""` — matching zero DB rows
- Stringified nulls: `Value::Null` was bound as `"null"` instead of SQL NULL

A local workaround `value_to_sql_param()` dispatched by variant to fix this. The fix was filed as [bridge issue #95](https://github.com/saurabh500/mssql-tiberius-bridge/issues/95) and shipped in preview.5 ([PR #96](https://github.com/saurabh500/mssql-tiberius-bridge/pull/96)).

## Changes

- **`Cargo.toml`**: bump `mssql-tiberius-bridge = "=0.1.0-preview.5"`
- **`sqlserver/helpers.rs`**: remove `value_to_sql_param()`, its `ToSql` import, and 6 now-redundant unit tests
- **`sqlserver/mod.rs`**: simplify parameter binding — no more `Box<dyn ToSql + Send + Sync>` intermediate vec or `&**b` double-deref:
  - Delete: `pk_vals.iter().map(|v| v as &dyn ToSql)`
  - Update: `once(&new_val).chain(pk_vals.iter()).map(|v| v as &dyn ToSql)`

## Tests

All 112 SQL Server driver tests pass.

Closes #145

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -14,7 +14,7 @@ version = "0.5.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d122413f284cf2d62fb1b7db97e02edb8cda96d769b16e443a4f6195e35662b0"
 dependencies = [
- "crypto-common",
+ "crypto-common 0.1.7",
  "generic-array",
 ]
 
@@ -26,7 +26,7 @@ checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
- "cpufeatures",
+ "cpufeatures 0.2.17",
 ]
 
 [[package]]
@@ -112,9 +112,9 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
 dependencies = [
  "anstyle",
  "anstyle-parse",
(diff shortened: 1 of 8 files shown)


saurabh500:
@debba  THis is ready for review

PR #213, opened 2026-05-18
fix(drivers): show pagination for SELECTs with leading SQL comments

ymadd:
## Summary

- `is_select_query` now strips leading `--` / `/* */` comments before checking for `SELECT`, matching `returns_result_set` and `is_explainable_query`. Without this, a `SELECT` preceded by a header-comment block was routed through the result-set fetch path but bypassed the pagination branch in `exec_on_*_conn`, returning `pagination: None` and hiding the UI controls entirely — so a 500-row result looked like a fixed slice with no way to advance.
- Fixed a latent bug in `strip_limit_offset` that the classifier change exposed: it tokenised then rejoined with `" "`, collapsing newlines. For a query like `-- header\nSELECT ... LIMIT 50` the rejoined output became `-- header SELECT ... LIMIT 50 OFFSET 0`, which MySQL/Postgres/SQLite all parse as a single `--` comment line — silently no-op. Switched the tokenizer to track byte offsets and made `strip_limit_offset` slice the original input, preserving comments and whitespace verbatim.
- Factored the three identical quoted-token scanner branches (`'`, `"`, `` ` ``) into a `read_quoted` helper while refactoring the tokenizer.

## Repro (before this PR)

```sql
-- ============================================
-- Any header comment block
-- ============================================
SELECT * FROM big_table ORDER BY id;  -- 500+ rows
```

Result: rows show, but the pagination bar is gone — only the first page is reachable.

After: pagination renders and pages can be navigated as expected.

## Test plan

- [x] `cargo test --lib drivers::common` — 42 passed, including 4 new regression tests:
  - `test_is_select_query_with_leading_comments`
  - `test_build_paginated_query_with_leading_comments`
  - `test_build_paginated_query_multiline_comments_with_user_limit`
  - `test_strip_limit_offset_preserves_leading_comments`
- [ ] Manual: paste a comment-headered `SELECT` returning >page_size rows against MySQL, Postgres, SQLite; verify pagination bar renders and Next/Prev work.

### DIFF
--- a/src-tauri/src/drivers/common/query.rs
+++ b/src-tauri/src/drivers/common/query.rs
@@ -1,6 +1,14 @@
-/// Check if a query is a SELECT statement
+/// Check if a query is a SELECT statement.
+///
+/// Leading SQL comments are stripped before checking, matching
+/// [`returns_result_set`] and [`is_explainable_query`] — drivers rely
+/// on this to route commented-header SELECTs through the pagination
+/// path.
 pub fn is_select_query(query: &str) -> bool {
-    query.trim_start().to_uppercase().starts_with("SELECT")
+    strip_leading_sql_comments(query)
+        .trim_start()
+        .to_uppercase()
+        .starts_with("SELECT")
 }
 
 /// Strip leading SQL comments (`-- …` line comments and `/* … */` block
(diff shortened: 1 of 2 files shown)


debba:
Fix looks good as-is. Separately, I was poking at the splitting dropdown you can see in the screenshot — those leading `--` comment blocks showing up as their own entries — to see if I could get the comments stripped out, but the splitting itself comes from `dbgate-query-splitter`, which we wired up in #119. Different rabbit hole, so I'm leaving it out of this PR and tracking it in #223 as a follow-up.

immagine

PR #215, opened 2026-05-18
feat(sqlserver): implement insert_record with IDENTITY_INSERT handling (#147)

### LINKED ISSUE #147: [Phase 2] IDENTITY_INSERT handling for inserts on identity tables
# [Phase 2] `IDENTITY_INSERT` handling for inserts on identity tables

Part of the SQL Server Phase 2 epic #150.

The driver isn't on `main`. Code lives on [`feat/sql-server`](https://github.com/TabularisDB/tabularis/tree/feat/sql-server) and isn't in any release. Check that branch out, branch off it, target your PR at it — not at `main`. The whole thing squashes into `main` when Phase 2 closes (#149).

SQL Server refuses any `INSERT` that provides a value for an `IDENTITY` column unless the session has `SET IDENTITY_INSERT [schema].[table] ON`. Phase 2 needs to detect this situation and wrap the insert in the right boilerplate.

## Task

In `src-tauri/src/drivers/sqlserver/mod.rs::insert_record` (currently a Phase 1 read-only stub that returns an error), replace the stub with a real implementation:

**1. Detect the identity column** via a metadata round-trip:

```sql
SELECT c.name
FROM sys.columns c
WHERE c.object_id = OBJECT_ID(@P1) AND c.is_identity = 1
```

(Or reuse the existing `Q_GET_COLUMNS` + filter on `is_identity`.)

**2. Two branches.** If the caller-provided `data: HashMap<String, serde_json::Value>` contains a value for the identity column, wrap the `INSERT` in an explicit transaction:

```sql
BEGIN TRAN;
SET IDENTITY_INSERT [schema].[table] ON;
INSERT INTO [schema].[table] ([c1], [c2], ...) VALUES (@P1, @P2, ...);
SET IDENTITY_INSERT [schema].[table] OFF;
COMMIT;
```

Otherwise plain `INSERT`.

**3. Error safety.** If the transaction fails, an implicit rollback leaves `IDENTITY_INSERT` OFF — but add an integration test that asserts this, so regressions are caught.

## Rules

- Integration test via `testcontainers` + `mcr.microsoft.com/mssql/server:2022-latest`, gated `#[ignore]` so CI without Docker skips it
- Test matrix: insert without identity value / insert with identity value / failed insert (e.g. NULL violation on NOT NULL column) / verify `IDENTITY_INSERT` state is OFF at the end of each

## Reference

- Plan: `docs/sql-server-implementation-plan.md` § Phase 2 — 2.4 (INSERT / IDENTITY handling)
- `SET IDENTITY_INSERT` docs: <https://learn.microsoft.com/en-us/sql/t-sql/statements/set-identity-insert-transact-sql>

saurabh500:
## Summary

Implements `Driver::insert_record` for the SQL Server driver with proper IDENTITY column handling. Closes #147.

## Changes

- **`sqlserver/helpers.rs`** — new pure helper `build_insert_sql(qualified, columns, wrap_identity_insert)` that emits either a plain `INSERT … VALUES (@P1, @P2, …)` or, when an IDENTITY column is being supplied, wraps the insert in `BEGIN TRY … BEGIN TRAN; SET IDENTITY_INSERT … ON; INSERT …; SET IDENTITY_INSERT … OFF; COMMIT TRAN; END TRY BEGIN CATCH IF @@TRANCOUNT > 0 ROLLBACK TRAN; SET IDENTITY_INSERT … OFF; THROW; END CATCH`. `IDENTITY_INSERT` is session-scoped and *not* reverted by rollback, so the CATCH block explicitly turns it back OFF before re-raising.
- **`sqlserver/helpers.rs`** — new helper `value_to_sql_param(&Value) -> Box<dyn ToSql>` that dispatches on the JSON variant to the bridge's typed primitive `ToSql` impls (`bool`, `i64`, `f64`, `String`, `Option::<String>::None`). The bridge's blanket `impl ToSql for serde_json::Value` JSON-stringifies *every* variant into `NVARCHAR(4000)` (filed upstream), which silently breaks inserts into `bit`/`int`/`float`/etc columns.
- **`sqlserver/introspection.rs`** — new `detect_identity_column(conn, schema, table)` async fn + `Q_GET_IDENTITY_COLUMN` constant. Queries `sys.columns` filtered by `is_identity = 1` for the resolved `object_id`.
- **`sqlserver/mod.rs`** — replaces the read-only stub with a real implementation:
  1. Validate non-empty `data`
  2. `detect_identity_column` to learn the table's IDENTITY column (if any)
  3. Sort columns deterministically and check whether the user-provided data includes the IDENTITY column → derive `needs_identity_insert`
  4. `build_insert_sql` with the right wrapping
  5. Bind params via `value_to_sql_param`, execute, return `execute_result.total()`

## Tests

- **9 new unit tests** in `helpers.rs`:
  - 4 covering `build_insert_sql` (plain INSERT, identifier escaping, identity wrapping shape, custom target name)
  - 1 covering `value_to_sql_param` dispatch across `Bool`/`Number(i64)`/`Number(f64)`/`String`/`Null`/`Array`/`Object`
- **1 new unit test** in `introspection.rs` covering `Q_GET_IDENTITY_COLUMN` query shape
- **Updated** `write_operations_error_out_in_phase1` — `insert_record` no longer errors out with "read-only" since the stub is gone; now asserts the empty-data validation message. Delete + create_view still assert "read-only".

All 102 driver tests pass (`cargo test --lib drivers::sqlserver`).

## Bridge limitation note

`preview.3`'s `execute()` returns 0 for DML row counts due to an mssql-tds DONE token limitation (bridge issue #1). Callers should verify via SELECT, not the return value. This PR returns `.total()` as-is — accurate row count handling is tracked by the bridge upgrade work (PR for #145).

## Out of scope (followups)

- Testcontainers-gated integration tests covering plain INSERT, identity-provided INSERT, and CATCH-path rollback.
- Removing manifest `readonly: true` once update/delete are also implemented.

Closes #147

### DIFF
--- a/src-tauri/src/drivers/sqlserver/helpers.rs
+++ b/src-tauri/src/drivers/sqlserver/helpers.rs
@@ -60,6 +60,68 @@ pub fn qualify(schema: Option<&str>, object: &str) -> String {
     format!("{}.{}", bracket_quote(schema), bracket_quote(object))
 }
 
+/// Build a parameterized SQL Server `INSERT` statement.
+///
+/// `qualified` is expected to already be a `[schema].[table]` produced by
+/// [`qualify`]. `columns` is the in-order list of column names that will be
+/// bound to `@P1, @P2, ...` (callers must bind values in the same order).
+///
+/// When `wrap_identity_insert` is `Some(target)`, the resulting batch toggles
+/// `SET IDENTITY_INSERT <target> ON` around the insert and is wrapped in
+/// `BEGIN TRY / BEGIN CATCH` so the session-scoped setting is always cleared,
+/// even if the insert fails. `target` should also be a `[schema].[table]`
+/// reference (typically the same as `qualified`); accepting it as a parameter
+/// keeps the helper pure and easy to unit-test.
+///
+/// Returns the SQL batch. The number of placeholders always matches
+/// `columns.len()`.
+pub fn build_insert_sql(
+    qualified: &str,
+    columns: &[String],
+    wrap_identity_insert: Option<&str>,
+) -> String {
+    let col_list = columns
+        .iter()
+        .map(|c| bracket_quote(c))
+        .collect::<Vec<_>>()
+        .join(", ");
+    let placeholders = (1..=columns.len())
+        .map(|i| format!("@P{}", i))
+        .collect::<Vec<_>>()
+        .join(", ");
+    let insert = format!(
+        "INSERT INTO {} ({}) VALUES ({})",
+        qualified, col_list, placeholders
+    );
+
+    match wrap_identity_insert {
+        None => format!("{};", insert),
+        Some(target) => {
+            // SET IDENTITY_INSERT is session-scoped and is *not* reverted by
+            // ROLLBACK, so the CATCH block must explicitly turn it OFF before
+            // re-raising. Setting OFF on a table that is already OFF is a
+            // no-op in SQL Server, so this is safe even if the failure occurs
+            // before the ON statement executes.
+            format!(
+                "BEGIN TRY\n\
+                     BEGIN TRAN;\n\
+                     SET IDENTITY_INSERT {target} ON;\n\
+                     {insert};\n\
+                     SET IDENTITY_INSERT {target} OFF;\n\
+                     COMMIT TRAN;\n\
+                 END TRY\n\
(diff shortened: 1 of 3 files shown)


saurabh500:
@debba this is ready for review

PR #216, opened 2026-05-18
perf: eliminate per-query disk I/O and unblock result display from me…

thomaswasle:
Problem

  Two independent issues were adding latency to every query execution:

  1. Connections file read on every query (backend)

  find_connection_by_id — called for every query, export, explain, schema command, etc. — unconditionally read and
  parsed the entire connections JSON file from disk using blocking std::fs::read_to_string on a Tokio worker thread.
  On SSH connections, expand_ssh_connection_params triggered a second file read for the SSH config. This happened even
   when the connection hadn't changed.

  2. Column/FK metadata fetch blocking result display (frontend)

  After each query completed, runQuery awaited fetchPkColumn — which fires parallel get_columns + get_foreign_keys
  backend calls — before calling updateTab with the query result. The result grid was invisible until both metadata
  requests finished. On a remote database, those two information_schema round-trips could add 100–500 ms of perceived
  query latency.

  Changes

  src-tauri/src/connection_cache.rs (new)

  Introduces ConnectionCache: a Mutex<Option<HashMap<String, SavedConnection>>> registered as Tauri state. Exposes
  three operations:
  - lookup(id) — returns Hit, Miss, or Cold in a single lock acquisition
  - populate(&[SavedConnection]) — fills the map on a cold miss
  - invalidate() — drops the map after any write to the connections file

  src-tauri/src/commands.rs

  - find_connection_by_id now checks the cache first. A cache hit is an O(1) HashMap lookup with no disk I/O. A cold
  miss reads the file once and warms the cache for all subsequent calls.
  - New save_connections_and_invalidate helper wraps every persistence::save_connections_file call and drops the cache
   afterwards, so stale data is never served. Replaces all 12 call sites.

  src/pages/Editor.tsx

  Reorders the post-query flow so updateTab with the result fires immediately after the query returns. fetchPkColumn
  is then started without await — it updates pkColumn, columnMetadata, and foreignKeys in the background via its own
  updateTab call. Both updates target independent tab keys, so the React functional state updater merges them
  correctly without races.

  Notes

  - The connection cache is intentionally whole-map invalidated (not per-key updated) on writes. Connections are saved
   infrequently; the simplicity is worth it.
  - fetchPkColumn already handles its own errors and calls updateTab({ pkColumn: null }) in its catch block, so
  removing the await does not affect error handling.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -310,17 +310,25 @@ pub fn find_connection_by_id<R: Runtime>(
     app: &AppHandle<R>,
     id: &str,
 ) -> Result<SavedConnection, String> {
-    let path = get_config_path(app)?;
-    if !path.exists() {
-        return Err("Connection not found".into());
-    }
-    // Use persistence module to properly load connections (handles both old and new formats)
-    let conn_file = persistence::load_connections_file(&path)?;
-    let mut conn = conn_file
-        .connections
-        .into_iter()
-        .find(|c| c.id == id)
-        .ok_or_else(|| "Connection not found".to_string())?;
+    let conn_cache =
+        app.state::<std::sync::Arc<crate::connection_cache::ConnectionCache>>();
+
+    let mut conn = match conn_cache.lookup(id) {
+        crate::connection_cache::CacheLookup::Hit(c) => c,
+        crate::connection_cache::CacheLookup::Miss => {
+            return Err("Connection not found".to_string())
+        }
+        crate::connection_cache::CacheLookup::Cold => {
+            let path = get_config_path(app)?;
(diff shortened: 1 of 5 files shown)


debba:
Nice work on this PR. I only added a few unit tests on `connection_cache.rs`: the cache itself was already clean and well thought out, what was missing was just coverage for the three `Hit`/`Miss`/`Cold` states and the `populate` → `invalidate` cycle.

It's nine small tests, but they pin down the transitions that actually matter — especially the distinction between a populated-but-empty cache (`Miss`) and a never-populated one (`Cold`), which is the exact thing that avoids the disk refetch and makes this PR interesting in the first place. Having a test that locks it down means whoever touches the cache later won't be able to break it silently.

Otherwise the perf win is solid and the diff is surgical. Nice one.

PR #220, opened 2026-05-18
fix(drivers): preserve BIGINT precision past JS Number.MAX_SAFE_INTEGER

### LINKED ISSUE #210: [Bug]: When using values above 2^53, the accuracy of the interface display is lost
### Describe the bug

When using values above 2^53, the accuracy of the interface display is lost, due to js limitations

### To Reproduce

1. Create a table with a 64-bit long integer number field using any type of database
2. Create a new record entry and store a large integer in this field, such as a snowflake ID:844197938335842304
3. Query it will be displayed as 844197938335842300 on the interface, but in fact the database is 844197938335842304

### OS Version

Windows 11

### Tabularis Version

0.10.3

### Relevant Log Output

```shell
It's easy to reproduce, should not need logs, there may not be this issue in the logs
```

NewtTheWolf:
## Summary

- Closes #210. BIGINT values above `2^53 - 1` (e.g. snowflake ids like `844197938335842304`) lost their last digits in the UI because the read path emitted them as JSON numbers and the frontend's `JSON.parse` rounds to the nearest IEEE 754 double.
- New `drivers::common::safe_int` helper emits `i64`/`u64` as a JSON **string** only when the value falls outside JS's safe integer range; smaller values stay native numbers, so the wire format for the 99% case is byte-identical to before.
- Wired through MySQL (`u64`/`i64` scalar fallback), Postgres (`INT8`, `XID8`, `MONEY`), and SQLite (`INTEGER`).
- Mirror change on the write-back side: a JSON string that parses cleanly as an i64 outside the safe range is bound as a native integer, so editing or deleting by a large bigint PK still matches. Postgres wraps the bind in `CAST AS bigint` to satisfy tokio-postgres' strict type checking; MySQL and SQLite rely on their implicit numeric coercion.
- 17 new unit tests covering boundary cases and the snowflake roundtrip. Full driver test suite: 297 passed.
- Demo seed `bigint_demo` added to MySQL and Postgres init scripts so the boundary cases are reproducible without ad-hoc SQL (mirrors the `text_demo` / `json_demo` pattern).

## Behaviour notes

- **Sort and filter are unaffected** because both run server-side. `handleSort` in `Editor.tsx` builds an `ORDER BY` clause and the filter toolbar builds a `WHERE` clause; the comparison happens in the database against the native BIGINT column, never in JS. The grid has no client-side global/quick filter that could trip over a mixed number/string column.
- Cells that now arrive as strings are still rendered identically (the renderer already accepts either shape), and the row editor sends them back as strings, which the write-back path translates back to `i64` before binding.

## Test plan

- [x] `cargo test --lib drivers` → 297 passed
- [x] `cargo build --lib` clean, no new clippy warnings on touched files
- [x] **MySQL manual test** (`bigint_demo` table):
  - [x] Read: snowflake `844197938335842304` displays exactly (was `…842300`)
  - [x] Read: i64::MAX `9223372036854775807` displays exactly
  - [x] Read: u64 `18446744073709551614` (BIGINT UNSIGNED above i64::MAX) displays exactly
  - [x] Boundary: `42` and `9007199254740991` stay JSON numbers (sortable numerically)
  - [x] Write UPDATE of non-PK column on the snowflake row matches exactly one row
  - [x] INSERT new row with `id = 7777777777777777777` succeeds
- [x] **Postgres manual test** (`bigint_demo` table):
  - [x] All read cases pass for `BIGINT`, `XID8` and `MONEY` columns
  - [x] UPDATE `note` on snowflake PK row → exactly one row updated
  - [x] UPDATE `amount` on snowflake row → bigint outside safe range round-trips through SET
  - [x] INSERT new row with `id = 7777777777777777777` succeeds

### DIFF
--- a/demo/init/mysql/05-bigint-demo.sql
+++ b/demo/init/mysql/05-bigint-demo.sql
@@ -1,0 +1,78 @@
+-- =============================================================
+-- Tabularis Demo — BIGINT precision showcase (MySQL 8)
+-- Database: tabularis_demo
+-- Purpose: exercise the JS-safe-integer fix for issue #210.
+--   JavaScript loses precision for integers outside ±(2^53 - 1).
+--   Values above that boundary must arrive in the UI as strings
+--   that preserve every digit; values inside the boundary must
+--   stay as native JSON numbers so sorting/filtering still work.
+-- =============================================================
+
+SET NAMES utf8mb4;
+
+USE tabularis_demo;
+
+DROP TABLE IF EXISTS bigint_demo;
+CREATE TABLE bigint_demo (
+  id          BIGINT       PRIMARY KEY,
+  unsigned_id BIGINT UNSIGNED,
+  amount      BIGINT,
+  label       VARCHAR(80)  NOT NULL,
+  note        VARCHAR(255)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+INSERT INTO bigint_demo (id, unsigned_id, amount, label, note) VALUES
+  -- Row 1: small signed bigint. Must stay a real JSON number so the
+  -- UI keeps numeric sort, range filters and inline arithmetic.
+  (42,
+   42,
+   1000,
+   'small (stays JSON number)',
+   'Below 2^53 — round-trips as number, no behaviour change.'),
+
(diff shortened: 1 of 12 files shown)


debba:
Looks good and thanks for demo seed :)

PR #221, opened 2026-05-18
feat: Delete selected rows with keyboard shortcut

### LINKED ISSUE #218: DataGrid: support Delete/Backspace key to delete selected rows
## Summary

Row deletion in the DataGrid is currently only reachable via right-click → context menu. Pressing the `Delete` or `Backspace` key while one or more rows are selected does nothing. A keyboard shortcut would match common spreadsheet/data-grid conventions (Excel, DBeaver, TablePlus, DataGrip) and reduce the friction of cleaning up test data.

## Current behaviour

- Selecting one or more rows in the grid and pressing `Delete` has no effect.
- The deletion logic already exists at `src/components/ui/DataGrid.tsx:845` (`deleteSelectedRow`) but is wired exclusively to the context menu callback.

## Proposed behaviour

- When the grid (or a cell within it) has focus and at least one row is selected, pressing `Delete` (and `Backspace` on macOS) marks the selected rows for deletion — same code path as the existing context-menu action.
- Guard against firing while a cell editor is open (so users can still delete characters inside a cell).
- Optional but nice: a small toast / status-bar hint announcing "N rows marked for deletion" so the action is discoverable.

## Out of scope

- No change to the actual deletion semantics — uses the existing `onMarkMultipleForDeletion` / `onMarkForDeletion` handlers and the user still has to commit pending changes.
- No multi-key combinations (Shift+Delete, etc.).

## Implementation pointer

Add a `keydown` handler on the grid root similar to the patterns already used in `src/components/notebook/NotebookView.tsx:519` and `src/components/layout/Sidebar.tsx:51`. Bail out early when:

- `event.target` is an `input`/`textarea`/`contenteditable` (a cell editor),
- `selectedRowIndices.size === 0`,
- a modifier key other than Shift is pressed.

thomaswasle:
## Delete selected rows with keyboard shortcut

  Pressing `Delete` or `Backspace` now marks all selected rows for deletion, matching the behaviour already available
  via the right-click context menu.

  ### Changes

  **`src/components/ui/DataGrid.tsx`**

  - Extracted the deletion logic from `deleteSelectedRow` into a new `deleteRowsByIndices(indicesToDelete: number[])`
  callback. The original context-menu handler now delegates to it, keeping behaviour identical.
  - Added `Delete` / `Backspace` handling to the document-level `keydown` listener. The shortcut fires only when all
  three conditions are true: rows are selected, no cell is currently being edited, and the grid is not in readonly
  mode.

  ### Behaviour

  | Scenario | Result |
  |---|---|
  | One or more rows selected, press `Delete` or `Backspace` | Rows are marked for deletion (pending commit) |
  | Cell is being edited, press `Backspace` | Normal text editing — no rows deleted |
  | Grid is readonly | Key is ignored |
  | No rows selected | Key is ignored |

  Existing rows follow the same pending-deletion flow as the context menu (committed on save). Pending insertions are
  discarded immediately, also matching existing behaviour.

Closes #218

### DIFF
--- a/src/components/ui/DataGrid.tsx
+++ b/src/components/ui/DataGrid.tsx
@@ -846,17 +846,7 @@ export const DataGrid = React.memo(
       pendingDeletions,
     ]);
 
-    const deleteSelectedRow = useCallback(() => {
-      if (!contextMenu) return;
-
-      // If the right-clicked row is part of a multi-selection, delete all selected rows.
-      // Otherwise fall back to deleting just the right-clicked row.
-      const rightClickedIsSelected = selectedRowIndices.has(contextMenu.rowIndex);
-      const indicesToDelete =
-        rightClickedIsSelected && selectedRowIndices.size > 1
-          ? Array.from(selectedRowIndices)
-          : [contextMenu.rowIndex];
-
+    const deleteRowsByIndices = useCallback((indicesToDelete: number[]) => {
       const pkVals: unknown[] = [];
       for (const idx of indicesToDelete) {
         const mergedRow = mergedRows[idx];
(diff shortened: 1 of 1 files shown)


NewtTheWolf:
tested. lgtm. Thanks @thomaswasle !

so i cant approve or merge it :laughing: but you have my go! really love it

debba:
Looks good! Thanks @thomaswasle !

PR #225, opened 2026-05-19
feat(sql): first-party splitter + per-driver dialect

### LINKED ISSUE #223: [Bug]: Splitter dropdown lists leading SQL comments as standalone statements
### What happened

When a SQL editor tab contains leading `--` or `/* */` comment blocks before a `SELECT` (or any other statement), the statement-splitting dropdown (next to the **Run** button) lists each leading comment line as its own entry alongside the actual statement.

### Repro

Paste the following into a Console tab and open the run-selection dropdown:

```sql
-- ============================================
-- Any header comment block
-- ============================================
SELECT * FROM big_table ORDER BY id;  -- 500+ rows
```

The dropdown shows entries like `-- ===…` and `-- 500+ rows` in addition to the `SELECT`, even though they are not executable statements.

### Expected

Leading/standalone comments should be merged into the following statement or filtered out — the dropdown should only list runnable statements.

### Context

Splitting is delegated to [`dbgate-query-splitter`](https://www.npmjs.com/package/dbgate-query-splitter), wired up in #119 (`src/utils/sql.ts`). The fix is either:

1. Post-process the splitter output in `splitQueries` to drop entries that are comment-only / merge them into the next statement, or
2. Configure / patch `dbgate-query-splitter` upstream if it exposes the right option.

Surfaced while working on #213 (pagination fix for SELECTs with leading comments).

Image

ymadd:
Closes #223.

## Summary

Replace `dbgate-query-splitter` with a first-party module under `src/utils/sqlSplitter/`. Comment-only fragments are now folded into the adjacent meaningful statement instead of surfacing as their own dropdown entry, which is the #223 bug. Per-driver SQL dialect is declared on `DriverCapabilities` and threaded through to the splitter so MySQL backticks, MSSQL `[...]`, PG dollar-quoting, etc. are honoured at split time (option B from the issue discussion, suggested by @debba).

## What landed

- `src/utils/sqlSplitter/` (index, tokenizer, splitter, classify): full first-party splitter
  - Strings (`''` / `\'` / `E'...'` / dollar-quoted / backtick / bracket), `--` and `/* */` comments (with PG nested-comment support), `;` / `DELIMITER` / `GO` delimiters
  - MySQL conditional `/*! ... */` emitted as its own statement so dump scripts execute correctly
  - Classifier mirror of `src-tauri/src/drivers/common/query.rs`: `isSelect`, `returnsResultSet`, `isExplainable`, `stripLeadingComments`, plus `MERGE` on both sides
- `DriverCapabilities.sql_dialect` (Rust enum, serde-validated at install) with built-in postgres / mysql / sqlite presets and a defensive `generic` fallback for unknown plugin manifests
- `src/utils/sql.ts` is a thin façade so existing call sites compile unchanged; old `splitQueries(sql)` calls keep working
- Editor wiring: `activeConnection.capabilities.sql_dialect` flows into every run / explain / dropdown path
- `dbgate-query-splitter` removed from dependencies and the open-source-libraries list

## Comment-fold rule

1. Leading comment-only span: folded into the NEXT meaningful statement (matches the previous splitter for header comments).
2. Trailing comment-only span: folded into the PREVIOUS meaningful statement. This is the #223 fix.
3. Entire input is comments: empty array. Callers already had a `queries[0] || fullText` fallback.

`Statement.range` points at the meaningful SQL body only, excluding any folded comments and any leading whitespace, so `sql.slice(range.start, range.end)` is always valid SQL.

## Test plan

- `pnpm vitest tests/utils/sqlSplitter tests/utils/sql.test.ts`: 104 passing
- `pnpm typecheck`: clean
- `cargo test drivers::common::tests::test_is_explainable_query`: 4 passing (covers `MERGE`)
- `pnpm test` full TS suite: 2376 / 2377 passing. The one failure is a pre-existing date-dependent test on `main`.
- Manual: pasted the #223 repro into a Console tab and confirmed one dropdown entry.

## Out of scope (follow-ups)

- Oracle `/` block terminator, Firebird PSQL `BEGIN...END`, MSSQL `adaptiveGoSplit`
- Postgres `COPY FROM stdin`
- Unifying `extractTableName` on the new tokenizer (still regex-based, pre-existing)

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
     "buffer": "^6.0.3",
     "clsx": "^2.1.1",
     "dagre": "^0.8.5",
-    "dbgate-query-splitter": "^4.12.0",
     "i18next": "^25.10.10",
     "i18next-browser-languagedetector": "^8.2.1",
     "json-edit-react": "^1.29.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -53,9 +53,6 @@ importers:
       dagre:
         specifier: ^0.8.5
         version: 0.8.5
-      dbgate-query-splitter:
-        specifier: ^4.12.0
-        version: 4.12.0
       i18next:
         specifier: ^25.10.10
         version: 25.10.10(typescript@5.9.3)
@@ -1560,9 +1557,6 @@ packages:
   data-urls@7.0.0:
     resolution: {integrity: sha512-23XHcCF+coGYevirZceTVD7NdJOqVn+49IHyxgszm+JIiHLoB2TkmPtsYkNWT1pvRSGkc35L6NHs0yHkN2SumA==}
     engines: {node: ^20.19.0 || ^22.12.0 || >=24.0.0}
-
-  dbgate-query-splitter@4.12.0:
-    resolution: {integrity: sha512-gHxH+QdFg488X8XXDHsWM0k1LwA+K5xcgp0YaA9wWye8A/zMCqaSDc4U6PMUu4LbKNchkIoTvFD4DJNXSsYVcw==}
 
   debug@4.4.3:
     resolution: {integrity: sha512-RGwwWnwQvkVfavKVt22FGLw+xYSdzARwm0ru6DhTVA3umU5hZc28V3kO4stgYryrTlLpuvgI9GiijltAjNbcqA==}
@@ -4238,8 +4232,6 @@ snapshots:
       whatwg-url: 16.0.1
     transitivePeerDependencies:
       - '@noble/hashes'
-
-  dbgate-query-splitter@4.12.0: {}
 
   debug@4.4.3:
     dependencies:
--- a/src-tauri/src/drivers/common/query.rs
+++ b/src-tauri/src/drivers/common/query.rs
@@ -70,8 +70,13 @@ pub fn returns_result_set(query: &str) -> bool {
 ///
 /// MySQL/MariaDB support EXPLAIN for DML statements only:
 /// SELECT, INSERT, UPDATE, DELETE, REPLACE, and WITH (CTE).
+/// PostgreSQL 15+ and Oracle also support EXPLAIN MERGE.
 /// DDL statements (CREATE, DROP, ALTER, TRUNCATE, etc.) are not supported.
 /// Leading SQL comments are stripped before checking.
+///
+/// Kept in sync with the TypeScript classifier in
+/// `src/utils/sqlSplitter/classify.ts` (EXPLAINABLE_KEYWORDS) so the
+/// editor's Explain UI cannot offer a statement the backend will refuse.
 pub fn is_explainable_query(query: &str) -> bool {
     let upper = strip_leading_sql_comments(query).to_uppercase();
     upper.starts_with("SELECT")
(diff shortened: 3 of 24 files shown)


debba:
One thing I'd like to throw out before the implementation lands — would it make sense to declare the dialect at the driver level rather than threading it from the frontend?

`DriverCapabilities` in `src-tauri/src/drivers/driver_trait.rs` already holds the dialect-specific knobs (`identifier_quote`, `auto_increment_keyword`, `serial_type`, `inline_pk`, `alter_column`). Adding a `sql_dialect` field there would let each driver — built-in and plugin — declare it once:

```rust
/// SQL dialect for the statement splitter/classifier.
#[serde(default = "default_dialect")]
pub sql_dialect: String, // "postgres" | "mysql" | "mssql" | "sqlite" | "oracle" | "generic"
```

Built-in drivers hardcode it in their `impl`, plugin drivers expose it in the manifest the same way they expose `identifier_quote` today. The frontend just reads `capabilities.sqlDialect` from the active connection and hands it to `splitQueries`.

A few things this would unlock:

- Removes the `postgres`-default as a long-term state. The "thread dialect into call sites" follow-up basically becomes "read it from the connection capabilities," which is already the pattern for everything else dialect-flavored.
- Plugin drivers get it for free. A ClickHouse / DuckDB plugin declares `"sql_dialect": "generic"` in the manifest, no PR to this repo needed.
- The Rust classifier in `drivers/common/query.rs` and the new TS splitter end up consulting the same source of truth instead of each guessing independently.

Open questions on my side:

- Default for existing plugins — `"generic"` feels safer than `"postgres"` (no `$$`, `DELIMITER`, `GO` surprises for non-postgres plugins already in the wild).
- Whether to model it as a typed enum on the Rust side and serialize to string, so a bad value in a manifest fails at install time rather than at the first split.
- Where on the frontend the capability is read — probably `DatabaseProvider` alongside the other `capabilities.*` fields the editor already consumes.

Curious if this fits the direction you had in mind, or if you'd rather keep the dialect a pure splitter concern decoupled from the driver layer. Either's fine by me, just wanted to raise it before the API signature settles.

ymadd:
@debba thanks, this fits well. The rails look like they're already
there end-to-end (`DriverCapabilities` → `DatabaseProvider` →
`useDatabase` → `Editor.tsx`), and since the splitter API in the
design is already dialect-aware, wiring it through should be
mechanical (rough guess: around 30 LOC across the manifest schema,
the trait, the 3 built-in driver impls, and the destructure in the
Editor). Sounds good to fold this into the same PR.

A couple of thoughts on your open questions:

- **Default for missing `sql_dialect`**: leaning `"postgres"` over
  `"generic"`. A `generic`-style dialect (no `$$`, no `DELIMITER`,
  no backticks) would silently strip dollar-quoting support from
  PG-compat plugins already in the wild that currently get postgres
  semantics via the existing dbgate-with-postgres-options path.
  `postgres` preserves today's behavior; non-PG plugins opt out
  explicitly.
- **Typed enum on the Rust side**: yes, worth it. A string-keyed
  dialect on the TS side risks an unknown value coming through from
  a plugin manifest and hitting a runtime trap. Validating at the
  trait boundary (or via serde) plus a `GENERIC` fallback on the
  splitter side covers both ends.
- **Frontend read location**: agreed, `DatabaseProvider` /
  `useDatabase` is the natural place. `capabilities` is already in
  scope at every call site that touches `splitQueries`.

Will start on it and push when there's something to look at.

PR #230, opened 2026-05-20
Add related records panel for foreign key values

### LINKED ISSUE #228: [Feat]: Feature Request: Related Records Panel for FK Values
### Prerequisites

### 🚀 Feature Proposal

Add a panel to the table editor that opens when you click on a foreign key (FK) value. 
The panel would show the referenced record(s) from the related table — similar to a peek view, but inline.

### Proposed Behavior
- Click any FK value → panel slides open below the current table
- Panel displays the referenced record(s) from the target table
- Main table stays visible and interactive
- Clicking a different FK updates the panel content without closing/reopening

### Motivation

I added a small prototype of this for my own workflow and found it very useful when quickly checking linked values (for example status enums, labels, metadata, etc.).

The inspiration came from DBeaver’s reference/dictionary-like panel, although this implementation would be simpler and more inline-focused.

### Possible Benefits
- Faster FK navigation
- Less context switching
- Easier inspection of related records
- Improved table editing workflow

### Example

_No response_

m-tonon:
## Summary
Adds an inline panel that opens when clicking a foreign key value in the table editor, showing the referenced record from the related table.

## Implementation
- Click any FK value → panel slides open below the current row
- Panel displays the referenced record(s) from the target table
- Main table remains visible and interactive
- Clicking a different FK updates the panel content dynamically

## Example
Before: Click `user_id = 123` in `orders` table → manually open `users` table → search for ID 123 → check email/role → navigate back to `orders`

After: Click `user_id = 123` → panel shows `users | id: 123 | email: john@example.com | role: admin` → glance at info → continue editing

## Testing
- [x] Click FK value → panel opens with correct data
- [x] Click different FK → panel updates
- [x] Panel closes when clicking outside

Closes #228


debba:
Really nice feature, love it!

A couple of UX thoughts to make it shine a bit more:

1. **Auto-close the panel when clicking a non-FK cell.** Right now it stays open even when focus moves to an unrelated column, which feels a bit sticky.
2. **Surface the action in the right-click context menu** (e.g. "Preview related record"). The cell-click trigger is great but a bit hidden — a menu entry alongside "Open referenced table in new tab" would help discoverability.

Optional extras worth considering: a `cursor-pointer` on FK cells as a visual affordance, a hover tooltip, or a keyboard shortcut on the focused cell. WDYT?

m-tonon:
Love the suggestions, thanks, let me improve it.

debba:
Small follow-up: when triggering **Preview related record** from the cell context menu, the underlying cell now becomes the focused/selected cell (matching the behavior of a direct left-click on an FK cell). Previously only the panel opened, leaving no visual indication of which cell the preview belonged to.

Ready to merge from my side.

PR #232, opened 2026-05-20
Fix filters for case-sensitive column names on PostgreSQL

### LINKED ISSUE #231: [Bug]: PostgreSQL filters fail with case-sensitive column names
### Describe the bug

Filters currently do not quote column identifiers when generating SQL clauses. This causes issues on PostgreSQL databases where identifiers are case-sensitive and require double quotes for columns containing uppercase or mixed-case names.

As a result, filters may fail or generate invalid queries when applied to PostgreSQL tables using non-lowercase column names.

### Expected Behavior

When connected to a PostgreSQL database:
- Column names should be wrapped with double quotes when generating filter SQL
- Generated queries should preserve the original column casing

Example:
"Status" = 'ACTIVE'

Instead of:
Status = 'ACTIVE'

### Proposed Fix
- Use the existing quoteIdentifier utility when building filter clauses
- Apply quoting conditionally for PostgreSQL drivers only
- Ensure all filter builders consistently receive the active driver context

### Additional Notes
This should cover both single and structured filter builders and include unit tests validating PostgreSQL vs non-PostgreSQL behavior.

### To Reproduce

1. Connect to a PostgreSQL database
2. Open a table containing mixed-case or uppercase column names (example: Status or UserName)
3. Apply a filter using the filter toolbar
4. Execute the filter/query

### OS Version

CachyOS (Kernel 7.0.9)

### Tabularis Version

v0.11.0

m-tonon:
**Fix: quote Postgres column names in structured filters**

Postgres is case-sensitive: `userId` was becoming `userid` and filters failed.
Now when driver = postgres, columns get double quotes via `quoteIdentifier` (MySQL/SQLite unchanged).

Closes #231

### DIFF
--- a/src/components/ui/TableToolbar.tsx
+++ b/src/components/ui/TableToolbar.tsx
@@ -18,6 +18,8 @@ import {
   createEmptyFilter,
 } from "../../utils/filterBar";
 import type { StructuredFilter } from "../../utils/filterBar";
+import { formatSqlIdentifier } from "../../utils/identifiers";
+import { formatSortClause } from "../../utils/tableToolbar";
 import { FilterRow } from "./FilterRow";
 import { SlotAnchor } from "./SlotAnchor";
 import { useDatabase } from "../../hooks/useDatabase";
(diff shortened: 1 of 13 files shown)


debba:
Hi! I fixed the failing tests in this PR. Could you please check that everything looks good before merging?

The changes only touch tests, so the implementation behavior should be unchanged.

m-tonon:
I noticed the `ORDER BY` had the same issue as well, so I created a helper to handle the column identifiers in a cleaner way and reused it wherever needed.

debba:
@m-tonon  That’s great.
I can merge it if you agree

m-tonon:
Now it's good to go :+1:

PR #236, opened 2026-05-21
feat(sqlserver): add Encrypt dropdown to connection dialog

saurabh500:
## Summary

Adds an **Encrypt** dropdown to the New Connection modal when the driver is **SQL Server**. Three options: **Yes** (default), **No**, **Strict**.

The backend already accepts the `encrypt` field on `ConnectionParams` and maps it correctly in `pool.rs::build_config` (`strict` → `Required`, `no` → `NotSupported`, default → `On`). This PR only adds the missing UI control.

## Changes

- `src/components/modals/NewConnectionModal.tsx` — local interface field, formData seed, driver-change reset, and conditional `<Select>` rendered for `driver === "sqlserver"` between the host/port grid and credentials block.
- `src/i18n/locales/en.json` — `newConnection.encrypt` label + `encryptModes.{yes,no,strict}` option labels (mirroring the existing `sslModes` pattern).

## Why draft

The new `encrypt` value is not yet threaded into the connection **save/load** path (i.e., the `Config` mapping where saved connections are reloaded into the dialog). Marking draft until that round-trip is wired so a saved connection re-opens with the previously chosen Encrypt setting.

## Manual test

1. Open New Connection → choose SQL Server → confirm the Encrypt dropdown appears between Host/Port and User/Password.
2. Switch driver away from SQL Server → confirm the dropdown disappears and `encrypt` is reset.
3. Pick **Strict** → connect to a server requiring TLS → confirms the value reaches the backend (existing `build_config` test coverage already validates the mapping).

## Out of scope

- `trust_server_certificate` UI control (field exists on the interface, no surface yet).
- Persisting encrypt on saved connections (the "why draft" item above).
- Other locale files; English-only for now, falls back via `defaultValue`.

### DIFF
--- a/src/components/modals/NewConnectionModal.tsx
+++ b/src/components/modals/NewConnectionModal.tsx
@@ -129,7 +129,7 @@ export const NewConnectionModal = ({
     username: "",
     database: "",
     ssl_mode: "",
-    encrypt: "",
+    encrypt: undefined,
     ssh_enabled: false,
     ssh_port: 22,
   });
@@ -369,7 +369,7 @@ export const NewConnectionModal = ({
       password: "",
       database: "",
       ssl_mode: "",
-      encrypt: "",
+      encrypt: undefined,
       ssh_enabled: false,
       ssh_connection_id: undefined,
       ssh_host: undefined,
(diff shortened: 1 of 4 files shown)


saurabh500:
This still needs work

PR #242, opened 2026-05-21
fix(schemas): surface get_schemas failure with error + retry

verbaux:
## Problem

When `get_schemas` rejects, the structure tree silently renders an empty flat layout — `TABLES (0)` / `VIEWS (0)` / `ROUTINES (0)`, no SCHEMAS header, no error. The connection looks open (the test connection succeeds on a separate code path), so it appears connected but shows nothing, with no indication anything failed.

image

Real-world example that exposed it: a Postgres database reached over an SSH tunnel failed the TLS handshake (`InvalidCertificate(UnknownIssuer)`). The schema fetch was rejected, but the UI showed only an empty tree (I still see this behavior in my app v0.10.3, but the fix for this issue has been implemented).

This PR addresses the missing UX safety net, which covers the whole class of `get_schemas` failures — TLS, permissions, dropped tunnel, network.

8oMogZ4eAAa


debba:
Hey @verbaux — quick UX adjustment I made on top of your branch before merging.

The raw error from a failed `get_schemas` (TLS, perms, dropped tunnel) lands in the sidebar as `"{Display}\n\n{Debug}"` produced by `map_pg_err` / `format_pg_error` in `drivers/postgres/client.rs`, so showing it as-is made the column either very tall or visually noisy. I tucked it behind a Details chevron:

- Preview under the title: only the part before `\n\n` (Display), `line-clamp-2`. Non-Postgres drivers don't use that separator, so the split is a no-op and the clamp handles overflow.
- Expanded: full raw text in a scrollable `<pre>` with a copy button (briefly toggles to a ✓ on click).

Also filled the new sidebar keys (`schemaLoadError`, `errorDetails`, `copyError`, `retry`) across the remaining 6 locales (de, es, fr, it, ja, zh) so they don't fall back to English.

I'm ready to merge unless you'd prefer a different layout — let me know.

verbaux:
Hello @debba!  
Thanks for your addition — I checked how it looks, and it really much better.

PR #245, opened 2026-05-22
Fix FK metadata loading

### LINKED ISSUE #244: [Bug]: FK redirection only works on some schemas
### Describe the bug

The FK redirection button, the action that opens the referenced record/table from a foreign key value cell, works in some PostgreSQL connections but not others.

In affected databases:
* no FK button is rendered
* no related-record panel opens
* no visual feedback or error is shown
* foreign keys appear to not be loaded at all

### Root Cause

The PostgreSQL FK discovery query based on `information_schema` was not returning FK metadata consistently in some environments/connections.

Because the application could not detect the table foreign keys correctly, the FK navigation button and related-record preview actions were never rendered in the UI.

### Proposed Fix

Replace the existing `information_schema` FK query with a `pg_catalog` implementation that reads PostgreSQL internal system tables directly (`pg_constraint`, `pg_attribute`, `unnest`).

This approach is more reliable and less dependent on `information_schema` behavior, restoring consistent FK discovery across environments/connections.

### Changes

* `src/drivers/postgres/mod.rs` — replaced FK metadata query

### To Reproduce

1. Connect to a PostgreSQL database containing valid foreign keys
2. Open a table that contains FK columns
3. Observe that FK metadata is not loaded in some environments/connections
4. Notice the FK navigation button and related-record preview actions are missing from FK value cells

#### Notes
The issue appears to depend on how the `information_schema` FK query behaves for specific PostgreSQL environments/connections. Replacing the query with a `pg_catalog` implementation consistently restores FK discovery.

### OS Version

CachyOS (Kernel 7.0.9)

### Tabularis Version

v0.11.0

m-tonon:
### Summary

Fix PostgreSQL foreign key metadata loading by swapping the `information_schema` FK query for a `pg_catalog` one.

The old query sometimes failed to load certain foreign keys — depending on how PostgreSQL exposed things through `information_schema`. When that happened, the FK buttons just disappeared from the UI.

The new query reads directly from PostgreSQL system catalogs (`pg_constraint`, `pg_attribute`, `pg_class`, `pg_namespace`) and correctly handles:

- foreign key relationships
- composite foreign keys
- update/delete rules

### Changes

- Replaced `information_schema` FK query with `pg_catalog` query
- Added integration test covering:
  - single-column foreign keys
  - composite foreign keys
  - cross-schema references
  - batch FK loading

### Testing

Ran:
```bash
cargo test test_postgres_foreign_keys_via_pg_catalog -- --ignored
```

Also manually checked that FK buttons and navigation work in the editor UI.

Closes #244


debba:
Thanks for this PR :)
I'll merge it now.

PR #252, opened 2026-05-25
feat: integrate Quick Navigator search overlay

debba:
# Description
This PR integrates a new **Quick Navigator** (search overlay/dialog) for database schema elements in the sidebar of an active connection, taking inspiration from Beekeeper Studio's Quick Search.

## Features
- **Customizable Shortcut**: The shortcut (`Cmd+P` / `Ctrl+P` by default) is fully customizable in settings under the "Navigation" category. We prevent browser print behavior when it triggers.
- **Dynamic Search Overlay**: Implemented `QuickNavigatorModal.tsx` that autofocuses a search input and dynamically filters tables, views, routines, and triggers.
- **All-Database Background Search**: Automatically resolves and indexes all databases/schemas configured for the active connection in the background when the modal is opened.
- **Section Grouping & Separators**: Results are grouped by database/schema under clean, high-readability headers, removing repetitive pills on individual rows.
- **Interactive Quick Actions**: Added action icons next to search items on hover (Inspect Structure, Generate SQL, Run Query).

## Implementation Details
- Extracted item listing and search filtering logic into a separate unit-tested utility module (`src/utils/quickNavigator.ts`).
- Added comprehensive unit tests in `tests/utils/quickNavigator.test.ts`.

## Verification
- Standard Vitest unit tests pass successfully (2432/2432 tests passed).
- TypeScript compiler passes with 0 type-check errors.


debba:
Hey @Lenobre, pushed a few fixes to the Quick Navigator branch, here's what changed:

The big one was the freeze when opening a result on connections with lots of tables (was testing with ~300). Turned out the sidebar table item wasn't memoized, so picking a table re-rendered every single row instead of just the two that actually change. Wrapped it in `React.memo` with a comparator that only looks at the data that matters and whether that specific row is the active one. Now switching tables only touches the two affected rows.

Second, on multi-database connections selecting a table in a different (collapsed) database didn't open that database. The schema item already auto-expanded when it became active, but the database item only did it on mount, so it never reacted afterwards. Made it behave like the schema one: auto-expand and lazy-load its tables when it becomes the active db.

Third, scrolling to the table worked going down but not up. The scroll ran once, before the freshly expanded db/schema had mounted its tables (they load async), so it found nothing and gave up. Now it retries over a few frames until the row is actually in the DOM. Also moved the selector into a small helper that escapes table/schema names, so weird names don't break the lookup.

Pulled the testable bits into `utils/sidebarTableItem.ts` and added unit tests for them. Also resolved the conflicts with main.

PR #256, opened 2026-05-26
fix(mcp): dispatch plugin drivers via the registry + harden the --mcp subprocess

### LINKED ISSUE #255: [Bug]: MCP tools fail with "Unsupported driver" for plugin-based connections
### Describe the bug

All MCP tools (`list_tables`, `describe_table`, `run_query`) and the `resources/read` schema endpoint return `MCP error -32000: Unsupported driver` for any connection that uses a **plugin driver** (e.g. the Hacker News, Redis, or Firestore plugins). Only the three built-in drivers (mysql/postgres/sqlite) work over MCP, even though the same connections work fine in the GUI.

Root cause: the MCP server hardcodes a `match` on the driver string for mysql/postgres/sqlite and falls through to `Unsupported driver` for everything else, instead of going through the shared driver registry the GUI uses. The standalone `--mcp` subprocess also never registers the plugin drivers (it has no Tauri `AppHandle`).

### To Reproduce

1. Install a plugin driver (e.g. the Hacker News plugin) and create a connection with it.
2. Connect an MCP client (Claude, etc.) to Tabularis.
3. Call `list_tables` (or `run_query`) on that connection via MCP.
4. See `MCP error -32000: Unsupported driver`.

### OS Version

macOS (Darwin 25.5.0)

### Tabularis Version

v0.12.0

### Relevant Log Output

```shell
MCP error -32000: Unsupported driver
```

ymadd:
## Summary

MCP could only reach built-in mysql/postgres/sqlite connections — every plugin-driven connection (Hacker News, Redis, …) failed with `Unsupported driver`. This routes all MCP driver operations through the shared driver registry (the same path the GUI uses) and registers the built-in + installed plugin drivers when the standalone `--mcp` subprocess starts.

Closes #255.

## What changed

- **Registry dispatch:** schema resource, `list_tables`, `describe_table`, `run_query`, and pre-flight `EXPLAIN` now resolve the driver via `registry::get_driver` instead of a hardcoded `match`.
- **Subprocess driver registration:** `--mcp` mode registers the three built-ins + installed plugins (`load_plugins_with_configs`), honoring `active_external_drivers`.

## Hardening surfaced by reaching plugins from MCP

- Bound plugin RPC calls with a timeout (120 s call / 15 s `initialize`) so a wedged plugin can't block the single-threaded request loop forever, and cancel the pending entry on timeout so it doesn't leak.
- `kill_on_drop` the plugin child so it isn't orphaned when the subprocess exits on stdin EOF.
- Refuse plugins that claim a built-in driver id (`mysql`/`postgres`/`sqlite`).
- Resolve `resources/read` through the keychain/SSH-aware path.
- Initialize the logger in `--mcp` mode (stderr only) so plugin-load errors are visible.
- Stop cleanly instead of panicking when a stdout write fails (BrokenPipe).
- Run `describe_table`'s column/FK/index fetches concurrently.

## Test plan

- [x] `cargo build` clean (no warnings) on latest `main`
- [x] `cargo test --lib plugins` passes
- [x] Verified end-to-end against a Hacker News plugin connection via the `--mcp` binary: `list_tables` → `["stories"]`, `run_query` → real rows, `resources/read` → schema (previously all `Unsupported driver`)
- [x] Confirmed no orphaned plugin process remains after the `--mcp` subprocess exits

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -122,6 +122,10 @@ pub fn run() {
     let args = cli::parse();
 
     if args.mcp {
+        // Initialize the logger so plugin-loading and driver RPC errors (which
+        // use the `log` crate) are visible. The custom logger writes to stderr
+        // only, leaving the stdout JSON-RPC stream clean.
+        init_logger(create_log_buffer(1000), log::LevelFilter::Info);
         let rt = tokio::runtime::Runtime::new().expect("Failed to create Tokio runtime");
         rt.block_on(mcp::run_mcp_server());
         return;
--- a/src-tauri/src/mcp/mod.rs
+++ b/src-tauri/src/mcp/mod.rs
@@ -7,13 +7,17 @@ use crate::config::{
     DEFAULT_MCP_APPROVAL_TIMEOUT_SECONDS, DEFAULT_MCP_PREFLIGHT_EXPLAIN,
 };
 use crate::credential_cache;
+use crate::drivers::driver_trait::DatabaseDriver;
+use crate::drivers::registry as driver_registry;
 use crate::drivers::{mysql, postgres, sqlite};
 use crate::heartbeat;
 use crate::models::{ConnectionParams, SshConnection};
 use crate::paths;
 use crate::persistence;
+use crate::plugins;
 use serde_json::{json, Value};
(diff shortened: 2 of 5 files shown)


debba:
Good catch! I will check it ASAP

PR #258, opened 2026-05-27
fix(plugins): unify plugin data directory under `tabularis`

### LINKED ISSUE #257: [Bug]: Plugins stored under inconsistent `com.debba.tabularis` directory instead of unified `tabularis`
### Describe the bug

The plugin subsystem resolves its storage directory with `ProjectDirs::from("com", "debba", "tabularis")` (`src-tauri/src/plugins/installer.rs`, `manager.rs`), while the rest of the application — config, themes, AI activity, heartbeat — uses `ProjectDirs::from("", "", "tabularis")` (`src-tauri/src/paths.rs`), and the Tauri bundle identifier is `tabularis`.

Because of the extra qualifier/organization, plugins are installed and loaded from a **different** folder than every other piece of app data:

| OS | App data dir | Plugins dir (actual) |
|----|--------------|----------------------|
| macOS | `~/Library/Application Support/tabularis/` | `~/Library/Application Support/com.debba.tabularis/plugins/` |
| Windows | `%APPDATA%\tabularis\` | `%APPDATA%\debba\tabularis\data\plugins\` |
| Linux | `~/.local/share/tabularis/` | `~/.local/share/tabularis/plugins/` (unaffected) |

This also contradicts the documentation (`plugins/PLUGIN_GUIDE.md`, `plugins/README.md`, `plugins/PLUGIN_TUTORIAL.md`) and the scaffolder template (`packages/create-plugin/templates/rust-driver/justfile.tmpl`), which all declare the plugins folder under `tabularis` (e.g. `%APPDATA%\tabularis\plugins\`). A plugin placed at the documented path — or dev-installed via `just dev-install` — is therefore not discovered by the loader on macOS and Windows.

**Expected:** all app data, including plugins, lives under a single `tabularis` directory on every platform, matching the documentation.

### To Reproduce

1. On macOS or Windows, install a plugin from **Settings → Plugins** (or scaffold one with `@tabularis/create-plugin` and run `just dev-install`).
2. Look inside the documented app data folder — macOS: `~/Library/Application Support/tabularis/plugins/`.
3. The plugin is not there: it landed in `~/Library/Application Support/com.debba.tabularis/plugins/` instead, split from the rest of the app data and not matching the docs.

### OS Version

macOS 14.x (also affects Windows; Linux unaffected)

### Tabularis Version

v0.12.0

### Relevant Log Output

```shell
(no error logged — plugins silently load from / save to the wrong directory)
```

debba:
## Summary

Plugins were resolved with `ProjectDirs::from("com", "debba", "tabularis")`, while the rest of the app data (config, themes, AI activity, heartbeat) uses `ProjectDirs::from("", "", "tabularis")` and the Tauri identifier is `tabularis`. On macOS and Windows this stored and loaded plugins from a separate `com.debba.tabularis` directory that also contradicted the documented path, so a plugin placed at the documented location was never discovered.

This unifies plugin storage under the same `tabularis` directory as all other app data, with no code duplication, plus a one-time migration of any existing plugins.

## Changes

- **`paths.rs`** — extract shared `project_dirs()` + a pure, unit-testable `unnested_app_dir()` helper; add `get_app_data_dir()`. `get_app_config_dir()` is refactored onto the same helpers with **identical output**.
- **`plugins/installer.rs`** — `get_plugins_dir()` now resolves via `paths::get_app_data_dir()`; add one-time `migrate_legacy_plugins_dir()` and a pure `migrate_plugins_between()` core.
- **`plugins/manager.rs`** — reuse `installer::get_plugins_dir()` instead of a duplicate `ProjectDirs` call.
- **`lib.rs`** — run the migration once at startup, before loading plugins.
- **Tests** — new `paths_tests.rs` and migration cases in `plugins/tests.rs`.

## Resulting plugin directory

| OS | Before | After (matches docs) |
|----|--------|----------------------|
| Linux | `~/.local/share/tabularis/plugins/` | `~/.local/share/tabularis/plugins/` (unchanged) |
| macOS | `~/Library/Application Support/com.debba.tabularis/plugins/` | `~/Library/Application Support/tabularis/plugins/` |
| Windows | `%APPDATA%\debba\tabularis\data\plugins\` | `%APPDATA%\tabularis\plugins\` |

The documented paths (`plugins/PLUGIN_GUIDE.md`, `plugins/README.md`, `plugins/PLUGIN_TUTORIAL.md`, scaffolder `justfile.tmpl`) already declared `tabularis`, so no doc changes are needed — the code now matches them.

## Migration

`migrate_legacy_plugins_dir()` runs once at startup and is idempotent. It moves plugin folders from the legacy `com.debba.tabularis` location into the new one only when the target is still empty (never clobbering an existing install), is a no-op on Linux (the two paths are already identical), and a no-op when there is nothing to migrate.

## Testing

- `cargo test` — `paths_tests` (4) and `plugins::tests` (6, including 4 new migration cases) pass.
- `cargo check --tests` — clean.

Closes #257


jonaspm:
tested on Windows 11 and the libsql plugin is detected successfully
image

debba:
That's great, so I will merge it :)

PR #262, opened 2026-05-28
Support for ssh authentication with security keys (like Yubikey)

robertpenz:
With this patch Tabularis allows ssh to prompt for a PIN to unlock the security key. It's tested it on Fedora 43.


debba:
Thanks for this PR — the feature is very welcome!

While testing it on Arch Linux + GNOME we hit a gap: `SSH_ASKPASS_REQUIRE=force` relies on a system askpass helper being installed, and many setups don't have one (the compiled-in default `/usr/lib/ssh/ssh-askpass` doesn't exist unless a package like `openssh-askpass`/`ksshaskpass` provides it — and macOS/Windows ship none at all). On those systems enabling the option made the connection fail with no visible prompt, which is hard to diagnose.

Since most users won't have an askpass helper installed, the simplest and most portable approach seemed to be reimplementing the askpass prompt as an in-app modal, so I pushed a commit (dacbdb7b) on top of your work that makes Tabularis act as its own askpass helper:

- ssh re-executes the Tabularis binary as `SSH_ASKPASS`; a thin client mode forwards the prompt to the running app over a private local socket (unix socket `chmod 600`, TCP loopback + auth token on Windows)
- the app shows a native modal for the three OpenSSH prompt types: secret (passphrase / security-key PIN), confirm (yes/no), and the `SSH_ASKPASS_PROMPT=none` notification ("Confirm user presence for key …"), which auto-dismisses once ssh detects the touch
- the tunnel readiness timeout is paused while a prompt is pending, so typing the PIN can't time the connection out
- if the app context isn't available it falls back to your original `SSH_ASKPASS_REQUIRE=force` behaviour

Tested here with a passphrase-protected key (prompt → passphrase → tunnel up), plus unit tests for the wire protocol, a server round-trip, and the frontend prompt queue.

I don't have a security key at hand for the full FIDO2 flow — could you confirm that with this commit your Yubikey still works as expected on Fedora 43 (PIN prompt in the app modal + touch notification)? Thanks!

robertpenz:
Nice, I can test it only in about 3 weeks - hope that's ok.

debba:
No problem @robertpenz keep me updated

robertpenz:
I'm back from hollidays :-). Are your changes already in main, or which version should I test?

debba:
Hi @robertpenz , no it's in this PR

robertpenz:
Works nicely with Yubikeys on my system.

debba:
That's great, I will merge it .

PR #267, opened 2026-05-28
Fix/postgres alias column autocomplete

### LINKED ISSUE #266: [Bug]: SQL autocomplete shows table names instead of columns after alias (PostgreSQL)
### Describe the bug

When writing a `SELECT` with a table alias and requesting completions after `alias.`, the SQL editor suggests **table names** instead of **column names**. This happens on PostgreSQL when the table is referenced with a double-quoted identifier.

- **Area:** SQL editor / Monaco autocomplete

## Expected behavior

Autocomplete should list **columns** from `"AccountEventLog"` (e.g. `id`, `created_at`, …).

## Actual behavior

Autocomplete lists **tables** in the schema (Monaco kind “Table”), not columns from the aliased table.

Image

## Additional cases

The same class of bug can affect:

- `SELECT u. FROM public.users u` (schema-qualified table + alias)
- `SELECT * FROM users u, orders o` then `o.` (comma-separated `FROM` — second alias not parsed)
- Any `alias.` where the alias was not resolved, which previously fell through to the default table suggestion list

Working case (before fix): unquoted simple tables, e.g. `SELECT u. FROM users u`.

## Root cause

1. **`parseTablesFromQuery`** (`src/utils/sqlAnalysis.ts`) only recognized unquoted identifiers and MySQL-style backticks. It did not parse PostgreSQL double-quoted table names (`"AccountEventLog"`).
2. The parser lowercased the entire SQL string before extraction, which broke mixed-case quoted identifiers.
3. When `alias.` could not be resolved, **`registerSqlAutocomplete`** (`src/utils/autocomplete.ts`) did not return early and fell through to the generic **table list** branch.

### To Reproduce

1. Connect to a PostgreSQL database that has a mixed-case or quoted table (example: `"AccountEventLog"`).
2. Open the SQL editor.
3. Type a query such as:

```sql
SELECT ael. FROM "AccountEventLog" ael
```

4. Trigger autocomplete immediately after `ael.` (type `.` or invoke suggestions).

### OS Version

CachyOS (Kernel 7.0.9)

### Tabularis Version

v0.12.0

### Relevant Log Output

```shell

```

m-tonon:
This PR implements a shared SQL autocomplete registration flow for both the editor and notebook views, improving lifecycle management and preventing stale or duplicate Monaco providers.

### Changes:
- Added `useSqlAutocompleteRegistration` hook to manage SQL autocomplete for active connections
- Updated `NotebookView` to utilize the new hook, passing the effective schema and active state
- Refactored `Editor` to replace direct SQL autocomplete registration with the hook
- Introduced `disposeSqlAutocomplete` to clean up resources on connection changes, reconnect, and failures
- Added support for schema-qualified table names, double-quoted identifiers, and alias resolution in PostgreSQL
- Added regression tests for schema-qualified tables, quoted identifiers, and autocomplete disposal
- Fixed false-positive table alias registration caused by commas inside function-call argument lists and AS alias column lists: replaced the comma alternative in `fromPattern` with a `splitTopLevelCommas()` pre-processing step that only splits at depth-0 commas

---
Closes #266

---
Built for NewtTheWolf by [Kilo](https://kilo.ai)


debba:
@m-tonon  Looks awesome!
I will check ASAP

NewtTheWolf:
@debba you can assign it to me

src/utils/autocomplete.ts
```diff
@@ -98,11 +99,32 @@ export const clearAutocompleteCache = (connectionId?: string) => {
...
+  tables.find((t) => t.name.toLowerCase() === name.toLowerCase())?.name;
+
+const tableInsertText = (
+  tableName: string,
+  driver?: string | null,
+  schema?: string | null,
+) =>
+  schema
+    ? quoteTableRef(tableName, driver, schema)
+    : formatSqlIdentifier(tableName, driver);
```
NewtTheWolf: The completion session is already scoped to the active schema (`getTableColumns(..., schema)`), so hard-qualifying every table as `"public"."table"` is redundant and noisy. Drop the schema prefix and quote only when needed (pairs with the call-site suggestion below + the `quoteIdentifierIfNeeded` helper from the review body):

```suggestion
const tableInsertText = (
  tableName: string,
  driver?: string | null,
) =>
  // Already scoped to the active schema — don't prefix "public". on every
  // table. Quote only when the name actually requires it.
  quoteIdentifierIfNeeded(tableName, driver);
```

If you want cross-schema qualification later, qualify only when `schema !== defaultSchema` rather than always.

src/utils/autocomplete.ts
```diff
@@ -259,7 +283,7 @@ export const registerSqlAutocomplete = (
         label: t.name,
         kind: monaco.languages.CompletionItemKind.Class,
         detail: "Table",
-        insertText: t.name,
+        insertText: tableInsertText(t.name, driver, schema),
```
NewtTheWolf: Match the simplified `tableInsertText` signature (no `schema` arg):

```suggestion
          insertText: tableInsertText(t.name, driver),
```

`quoteTableRef` then becomes unused in this file — drop it from the import on line 1.

src/utils/autocomplete.ts
```diff
@@ -163,13 +185,15 @@ export const registerSqlAutocomplete = (
             label: c.label,
             kind: monaco.languages.CompletionItemKind.Field,
             detail: c.detail,
-            insertText: c.label,
+            insertText: formatSqlIdentifier(c.label, driver),
```
NewtTheWolf: Columns get the same always-quote treatment — `"id"`, `"CreatedAt"`. Quote only when needed so plain lowercase columns stay clean while `"AccountId"` stays correctly quoted:

```suggestion
            insertText: quoteIdentifierIfNeeded(c.label, driver),
```

src/utils/autocomplete.ts
```diff
@@ -223,7 +247,7 @@ export const registerSqlAutocomplete = (
                 label: col.label,
                 kind: monaco.languages.CompletionItemKind.Field,
                 detail: `${col.detail} — ${table.name}${aliasHint}`,
-                insertText: col.label,
+                insertText: formatSqlIdentifier(col.label, driver),
```
NewtTheWolf: Same for the context-column insert path:

```suggestion
                insertText: quoteIdentifierIfNeeded(col.label, driver),
```

src/contexts/DatabaseProvider.tsx
```diff
@@ -691,6 +691,7 @@ export const DatabaseProvider = ({ children }: { children: ReactNode }) => {
     if (!targetId) return;
 
     clearAutocompleteCache(targetId);
+    disposeSqlAutocomplete();
```
NewtTheWolf: This is the root cause of point #1 above. The global provider mirrors the *active* editor, so only dispose it when the connection being torn down is actually the active one — otherwise disconnecting a background connection kills autocomplete in the editor you're using:

```suggestion
    if (targetId === activeConnectionId) {
      disposeSqlAutocomplete();
    }
```

The health-check handler at L786 has the same issue, but there `activeConnectionId` is a stale closure — you'd need an `activeConnectionIdRef` (or move disposal into the active-connection-change path). Cleanest long-term: let `useSqlAutocompleteRegistration` own disposal in its cleanup and drop both calls from here entirely.

NewtTheWolf (changes requested):
Hey @m-tonon — first off, this is genuinely cool work. 🙌 Bringing autocomplete to the notebook view, pulling the registration into a shared `useSqlAutocompleteRegistration` hook, and especially the #266 fix itself — resolving quoted / mixed-case / schema-qualified identifiers in `parseTablesFromQuery` — is solid, and the test coverage is great. Really nice.

Two things I'd love to see addressed before merge:

**1. (Blocking) Autocomplete dies when an *unrelated* connection disconnects.**
`disposeSqlAutocomplete()` tears down the single global completion provider, but it's called on *every* connection's disconnect / health-failure. Repro: open two connections, work in connection A's editor, disconnect B → A's autocomplete goes dead, because A's registration hook doesn't re-run (its deps didn't change). Confirmed locally. The provider should follow the *active* editor's lifecycle, not get torn down on arbitrary disconnects — see the inline note in `DatabaseProvider`.

**2. (Blocking, UX) Inserts are always fully quoted and always schema-prefixed.**
Before this PR autocomplete inserted bare names; now every Postgres insert becomes `"public"."users"`, `"id"`. That's noisier than every reference client: Postgres' own `quote_ident()`, DataGrip and DBeaver all quote **only when needed** (reserved word / mixed case / special char) and don't prefix the default schema. Crucially, the #266 fix is about *resolving* quoted identifiers for lookup — it doesn't require *emitting* quotes on every insert, so the two can be decoupled. The mixed-case correctness (`"AccountEventLog"` stays quoted) must stay; only the always-on quoting of plain lowercase names + the forced `public.` prefix should go.

To support the inline suggestions, I'd add a small helper to `src/utils/identifiers.ts` (it can't be a one-click suggestion since that file isn't in this diff):

```ts
// PostgreSQL folds unquoted identifiers to lowercase and only needs quotes for
// reserved words, mixed case, or special characters — mirroring quote_ident().
const PG_SAFE_IDENTIFIER = /^[a-z_][a-z0-9_$]*$/;
const PG_RESERVED = new Set([
  "select","from","where","table","user","order","group","join","and","or",
  "as","in","on","by","null","true","false","default","check","column","limit","offset",
]);

/** Like formatSqlIdentifier, but quotes only when the identifier actually needs it. */
export function quoteIdentifierIfNeeded(
  identifier: string,
  driver: string | null | undefined,
): string {
  if (!shouldQuoteIdentifiers(driver)) return identifier; // non-PG: unchanged
  if (PG_SAFE_IDENTIFIER.test(identifier) && !PG_RESERVED.has(identifier)) {
    return identifier;
  }
  return quoteIdentifier(identifier, driver);
}
```

The column/table inline suggestions below reference this helper, so they apply as a set. Net result: `SELECT id, name FROM users` for plain lowercase, `SELECT "AccountId" FROM "AccountEventLog"` for mixed-case — exactly what you'd want. 🚀

Thanks again — really nice contribution, this is the right direction!

m-tonon:
Thanks for the review! I'll take a look at the feedback, make the necessary improvements, and push an update soon. :rocket:

debba:
Hi @m-tonon !
any news about it?

m-tonon:
Hey, both blocking points are fixed:

**#1 — Autocomplete dying on unrelated disconnect**
- Autocomplete no longer shuts down when a background connection drops
- Only the active editor's cleanup controls disposal now

**#2 — Noisy inserts (always quoted + schema-prefixed)**
- Smart quoting: `users` stays plain, `"AccountEventLog"` gets quotes
- No more `"public"."table"` prefix on inserts
- Uses existing `formatSqlIdentifier` (no extra helper)

Ready for another look when you have time.

NewtTheWolf:
hey @m-tonon thanks for addressing the change request!

clould you fix the confilcts? after that i will rereview it!

src/pages/Editor.tsx
```diff
@@ -2186,6 +2187,12 @@ export const Editor = () => {
     });
   };
 
+  useSqlAutocompleteRegistration(activeConnectionId, {
+    monaco: monacoInstance,
+    schema: activeSchema,
+    enabled: !isNotebookTab,
+  });
+
   useEffect(() => {
```
NewtTheWolf: This leftover effect ([L2196-L2214](https://github.com/TabularisDB/tabularis/blob/7141e008f9362f5ead50c96ac8e00407bf2ed374/src/pages/Editor.tsx#L2196-L2214)) duplicates the `useSqlAutocompleteRegistration` hook added just above and — critically — calls `registerSqlAutocomplete` **without the `driver` argument**, so the provider ends up registered with `driver=undefined`. Since it's the last write to the single global provider, it overrides the hook and breaks smart-quoting (mixed-case identifiers insert unquoted).

Please remove this whole effect — the hook already handles registration with the driver. That also makes the `registerSqlAutocomplete` import (L86) unused, so it can go too.

NewtTheWolf (changes requested):
Re-reviewed on the rebased branch (`7141e00`), pulled locally and tested against a real Postgres schema with mixed-case tables.

The two original blockers are addressed — disposal is now owned by the hook (no more autocomplete dying on a background disconnect 👍), and inserts go through `formatSqlIdentifier` for smart quoting. But I hit a **regression** that blocks merge:

**On Postgres, mixed-case identifiers now insert unquoted** — e.g. autocompleting `AccountEventLog` yields `SELECT * FROM AccountEventLog` instead of `"AccountEventLog"`, which is invalid SQL (Postgres folds it to lowercase). Verified live: the completion provider receives `driver=undefined`.

**Root cause:** `Editor.tsx` registers the completion provider **twice** — once via the new `useSqlAutocompleteRegistration` hook (which passes `activeDriver`), and once via a leftover `useEffect` that calls `registerSqlAutocomplete(...)` **without** the `driver` argument. There's a single global provider (last registration wins), so the leftover effect clobbers the hook's registration → `driver=undefined` → `shouldQuoteIdentifiers` returns false → nothing gets quoted. Looks like the rebase onto the new `main` re-introduced the direct registration this PR had replaced.

**Fix:** remove that leftover effect — the hook already covers registration (with the driver). Details inline. After deleting it locally I re-confirmed `drv` becomes `"postgres"` and `AccountEventLog` inserts as `"AccountEventLog"`. ✅

NewtTheWolf:
Hey @m-tonon, thanks again for this — works great. While testing on Postgres I hit one edge case: typing an opening `"` before picking a suggestion (Monaco auto-closes it) produced `""AccountEventLog"`, and deleting the auto-closed quote left the identifier unclosed.

To avoid a review ping-pong round, I pushed a fix straight to your branch (`2daa3ec`): when an opening quote precedes the completion, the range now swallows the surrounding quote(s) and inserts a canonical quoted identifier, with a matching `filterText` so suggestions still show up. Added two tests for it; full suite is green.

Could you give it a quick validate on your side before we merge? 🙏

NewtTheWolf (dismissed):
LGTM 🚀 Postgres alias + quoted-identifier autocomplete works end to end (verified on a live demo DB). Full suite green. Approving — just give the quote-completion fix I pushed a quick validate on your side before merge.

src/utils/sqlAnalysis.ts
```diff
@@ -23,31 +31,31 @@ const extractFromSection = (sql: string): string => {
...
   if (!fromSection) return null;
 
   const tableMap = new Map<string, ParsedTableRef>();
-  // Groups: 1=first-id (schema when group 2 present, else table), 2=table (qualified), 3=alias.
-  // The negative-lookahead stops the optional alias group from swallowing a keyword that
-  // legally follows a table name (JOIN/LEFT/NATURAL/FOR/…). Without it the keyword is both
-  // mis-registered as an alias and consumed, dropping the table that follows it.
-  const fromPattern = /(?:from|join)\s+`?([a-z_][a-z0-9_]*)`?(?:\.`?([a-z_][a-z0-9_]*)`?)?(?:\s+(?:as\s+)?`?(?!(?:join|left|right|inner|outer|cross|natural|full|on|using|where|group|order|having|limit|offset|union|intersect|except|for|fetch|window|lateral|tablesample|qualify|straight_join)\b)([a-z_][a-z0-9_]*)`?)?/gi;
+  const fromPattern =
+    /(?:from|join|,)\s+("(?:[^"]|"")*"|`[^`]+`|[a-zA-Z_][a-zA-Z0-9_]*)(?:\.("(?:[^"]|"")*"|`[^`]+`|[a-zA-Z_][a-zA-Z0-9_]*))?(?:\s+(?:as\s+)?("(?:[^"]|"")*"|`[^`]+`|(?!(?:join|left|right|inner|outer|cross|natural|full|on|using|where|group|order|having|limit|offset|union|intersect|except|for|fetch|window|lateral|tablesample|qualify|straight_join)\b)[a-zA-Z_][a-zA-Z0-9_]*))?/gi;
```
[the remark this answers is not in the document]
NewtTheWolf: @kilocode-bot fix it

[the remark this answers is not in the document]
NewtTheWolf: @kilocode-bot fix it

debba (dismissed):

PR #269, opened 2026-06-01
fix(ai): cache AI key reads to stop repeated Keychain prompts on the AI tab

ymadd:
## Problem

Opening the **AI** settings tab repeatedly triggers the macOS Keychain
authorization prompt. A single tab open issues 8+ keychain reads:

- `SettingsProvider` calls `check_ai_key` per provider on mount
- `AiTab` calls `check_ai_key_status` per provider on mount
- `get_ai_models` reads the OpenAI / custom-openai keys

Each call hit `keychain_utils::get_ai_key` directly, so the keychain was
read on every render.

## Fix

`ca2e668` already added `CredentialCache` (with `get_ai_key_cached` /
`set_ai_key_cached` / `invalidate_ai_key`) to "cache db/ssh/ai creds",
but the AI read paths were never wired to it. This PR completes that:

- **config**: `get_ai_api_key` / `get_ai_api_key_status` read through the
  cache; `check_ai_key` / `check_ai_key_status` / `set_ai_key` /
  `delete_ai_key` take `AppHandle` to reach the managed cache state
  (write-through on set, invalidate on delete).
- **ai**: `&AppHandle` threaded through `dispatch_provider` and
  `get_ai_models`.

Result: the keychain is read **at most once per provider per session**,
matching how DB/SSH credentials already behave. Repeated prompts on tab
open are gone.

### Bonus: transient-error robustness

`keychain_utils::get_ai_key` now returns `Result<Option<String>, String>`
so the cache only records "absent" on a definitive `NoEntry`. Previously
a denied/timed-out prompt was cached as permanent absence, making a
configured key appear missing until the app restarted.

## Notes

- **No frontend changes**: Tauri injects `AppHandle`, so the existing
  `invoke(...)` call sites are unchanged.
- DB/SSH keychain readers still use `Result<String, String>` and conflate
  `NoEntry` with real errors; adopting the `Option` shape there is a
  sensible follow-up, out of scope here.

## Test plan

- [x] `cargo check --tests` passes
- [ ] Open the AI settings tab → no repeated Keychain prompts
- [ ] Set / delete an AI key → status updates without restart
- [ ] Generate / explain / model-list still works per provider

### DIFF
--- a/src-tauri/src/ai.rs
+++ b/src-tauri/src/ai.rs
@@ -285,7 +285,7 @@ pub async fn get_ai_models(
                 // Always refresh custom-openai as it depends on user configuration
                 if let (Some(base_url), Ok(api_key)) = (
                     app_config.ai_custom_openai_url.clone(),
-                    config::get_ai_api_key("custom-openai"),
+                    config::get_ai_api_key(&app, "custom-openai"),
                 ) {
                     if !base_url.is_empty() && !api_key.is_empty() {
                         let custom_models = fetch_custom_openai_models(&base_url, &api_key).await;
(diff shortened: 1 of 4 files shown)


debba:
Thanks a lot for this @ymadd !

PR #270, opened 2026-06-01
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade MiniMax model configuration to include the latest M3 model as the new default.

## Changes
- Add `MiniMax-M3` to the model selection list in `src-tauri/src/ai_models.yaml`
- Place `MiniMax-M3` first so it is auto-selected as the default when only the MiniMax key is configured
- Retain `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` as alternatives
- Update the existing Rust (`ai.rs`) and TypeScript (`minimax.test.ts`, `settings.test.ts`) test assertions for the new default
- Update README provider section to reflect the new default

## Why
MiniMax-M3 is the new flagship model and is now generally available. This keeps Tabularis on the latest model by default while preserving access to the previous generation for users that prefer it.

## Testing
- TypeScript: `pnpm exec vitest run tests/utils/minimax.test.ts tests/utils/settings.test.ts` — all assertions pass with the new default
- YAML parse verified

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,7 +366,7 @@ You can override or add custom models for AI providers by editing `config.json` 
   "aiCustomModels": {
     "openai": ["gpt-4o", "gpt-4-turbo", "gpt-3.5-turbo", "gpt-5-preview"],
     "anthropic": ["claude-3-opus-20240229", "claude-3-sonnet-20240229"],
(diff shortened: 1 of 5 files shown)


debba:
Thanks for this, will merge it now.

Issue #276, opened 2026-06-03
[Bug]: Explain Plan does not work for Postgres


GiteshDohare:
### Describe the bug

When I try to use "Explain Plan" with Postgres Database i get this error
**_Explain query failed: error deserializing column 0_**

Works fine on MySQL and SQLite, Does not work for Postgres (tested with Postgres v16 and v18)

<img width="1268" height="457" alt="Image" src="https://github.com/user-attachments/assets/1ccb3adf-76e2-404e-8e1b-857ba38b8256" />

### To Reproduce

use  Explain Plan feature with any query in Postgres

### OS Version

Ubuntu 24.04, Windows 10

### Tabularis Version

v0.12.0

### Relevant Log Output

```shell
[Logger] Capturing logger initialized successfully with level: Info
[LOG] [2026-06-03 11:53:41.153] [INFO] tabularis_lib - Tabularis application starting...
[LOG] [2026-06-03 11:53:41.153] [INFO] tabularis_lib - Debug mode enabled - verbose logging active
[LOG] [2026-06-03 11:53:41.211] [WARN] arboard::platform::linux - Tried to initialize the wayland data control protocol clipboard, but failed. Falling back to the X11 clipboard protocol. The error was: Unknown error while interacting with the clipboard: A required Wayland protocol (ext-data-control, or wlr-data-control version 1) is not supported by the compositor
[LOG] [2026-06-03 11:53:41.400] [INFO] tabularis_lib::drivers::registry - Registering driver: MySQL (mysql)
[LOG] [2026-06-03 11:53:41.400] [INFO] tabularis_lib::drivers::registry - Registering driver: PostgreSQL (postgres)
[LOG] [2026-06-03 11:53:41.400] [INFO] tabularis_lib::drivers::registry - Registering driver: SQLite (sqlite)
[LOG] [2026-06-03 11:53:41.401] [INFO] tabularis_lib::ai_approval_watcher - AI approval watcher started on /home/gitesh/.config/tabularis/pending_approvals
[LOG] [2026-06-03 11:53:41.401] [INFO] tabularis_lib::health_check - Health check: starting ping loop with interval 30s
[LOG] [2026-06-03 11:53:41.402] [INFO] tabularis_lib - DevTools opened (debug mode active)
[LOG] [2026-06-03 11:53:44.336] [INFO] tabularis_lib::commands - Testing connection to database: test-db
[...]

PR #279, opened 2026-06-03
fix(postgres): read EXPLAIN JSON output as json column

debba:
## What

Fixes the "Explain Plan" feature for Postgres, which always failed with `error deserializing column 0` (#276).

## Why it broke

`EXPLAIN (FORMAT JSON) ...` returns the plan in a single column, and on Postgres that column is typed as `json` (OID 114), not `text`. The code read it straight into a `String`:

```rust
let plan_json_str: String = rows[0].try_get(0)...;
```

`tokio-postgres` refuses to deserialize a `json` column into a `String`, so the call errored out before the plan was ever parsed. This is server-side behaviour that's been stable since the `FORMAT` option landed in PG 9.0, so the feature never worked on any Postgres version — matching the report against PG 16 and 18.

## Fix

Read the column as a `serde_json::Value` (the `with-serde_json-1` feature is already enabled) and re-serialize it for the existing parser. Postgres-compatible engines that hand the plan back as a plain `text` column are covered by a `String` fallback.

## Testing

- `cargo check` passes.
- Verified Explain Plan now renders against Postgres locally, with and without Analyze.

### DIFF
--- a/src-tauri/src/drivers/postgres/explain.rs
+++ b/src-tauri/src/drivers/postgres/explain.rs
@@ -29,8 +29,15 @@ pub async fn explain_query(
         return Err("EXPLAIN returned no output".into());
     }
 
-    // PostgreSQL returns a single row with a single text column containing JSON
-    let plan_json_str: String = rows[0].try_get(0).map_err(|e| format_pg_error(&e))?;
+    // `EXPLAIN (FORMAT JSON)` on real PostgreSQL returns a column of type
+    // `json` (not `text`), so reading it as `String` fails with
+    // "error deserializing column 0". Read it as a `serde_json::Value` and
+    // re-serialize. Some Postgres-compatible engines hand back a plain `text`
+    // column instead, so fall back to reading the raw string in that case.
+    let plan_json_str = match rows[0].try_get::<_, serde_json::Value>(0) {
+        Ok(value) => value.to_string(),
+        Err(_) => rows[0].try_get::<_, String>(0).map_err(|e| format_pg_error(&e))?,
+    };
 
     let mut plan = parse_postgres_json(&plan_json_str)?;
     plan.original_query = query.to_string();
--- a/src-tauri/src/drivers/postgres/explain.rs
+++ b/src-tauri/src/drivers/postgres/explain.rs
@@ -36,7 +36,10 @@ pub async fn explain_query(
     // column instead, so fall back to reading the raw string in that case.
     let plan_json_str = match rows[0].try_get::<_, serde_json::Value>(0) {
         Ok(value) => value.to_string(),
-        Err(_) => rows[0].try_get::<_, String>(0).map_err(|e| format_pg_error(&e))?,
+		Err(json_err) => rows[0].try_get::<_, String>(0).map_err(|e| {
+    	log::debug!("EXPLAIN json read failed ({json_err}); text read also failed");
+    	format_pg_error(&e)
+	})?,
     };
 
     let mut plan = parse_postgres_json(&plan_json_str)?;


src-tauri/src/drivers/postgres/explain.rs
```diff
@@ -29,8 +29,15 @@ pub async fn explain_query(
...
-    // PostgreSQL returns a single row with a single text column containing JSON
-    let plan_json_str: String = rows[0].try_get(0).map_err(|e| format_pg_error(&e))?;
+    // `EXPLAIN (FORMAT JSON)` on real PostgreSQL returns a column of type
+    // `json` (not `text`), so reading it as `String` fails with
+    // "error deserializing column 0". Read it as a `serde_json::Value` and
+    // re-serialize. Some Postgres-compatible engines hand back a plain `text`
+    // column instead, so fall back to reading the raw string in that case.
+    let plan_json_str = match rows[0].try_get::<_, serde_json::Value>(0) {
+        Ok(value) => value.to_string(),
+        Err(_) => rows[0].try_get::<_, String>(0).map_err(|e| format_pg_error(&e))?,
```
NewtTheWolf: Nit (non-blocking): the `Err(_)` discards the original `json`-deserialization error, so if *both* reads fail you only ever surface the `String` error — which can be misleading when debugging an unusual driver. Could keep it for context, e.g.

```rust
Err(json_err) => rows[0].try_get::<_, String>(0).map_err(|e| {
    log::debug!("EXPLAIN json read failed ({json_err}); text read also failed");
    format_pg_error(&e)
})?,
```

src-tauri/src/drivers/postgres/explain.rs
```diff
@@ -29,8 +29,15 @@ pub async fn explain_query(
...
 
-    // PostgreSQL returns a single row with a single text column containing JSON
-    let plan_json_str: String = rows[0].try_get(0).map_err(|e| format_pg_error(&e))?;
+    // `EXPLAIN (FORMAT JSON)` on real PostgreSQL returns a column of type
+    // `json` (not `text`), so reading it as `String` fails with
+    // "error deserializing column 0". Read it as a `serde_json::Value` and
+    // re-serialize. Some Postgres-compatible engines hand back a plain `text`
+    // column instead, so fall back to reading the raw string in that case.
+    let plan_json_str = match rows[0].try_get::<_, serde_json::Value>(0) {
+        Ok(value) => value.to_string(),
```
NewtTheWolf, line 38: Nit (non-blocking): this is a triple round-trip — PG sends JSON text → tokio-postgres parses into `serde_json::Value` → `.to_string()` re-serializes → `parse_postgres_json` parses it *again*. Totally fine for small EXPLAIN output, just flagging in case `parse_postgres_json` could grow a `&serde_json::Value` entry point to skip the re-parse. Not worth it unless it's trivial.

NewtTheWolf (commented):
@debba overall this looks good to me 👍 — correct root cause (`EXPLAIN (FORMAT JSON)` on real PostgreSQL returns a `json` column, OID 114, not `text`; I verified this locally against PG 16 with `\gdesc` → `QUERY PLAN | json`). The fix is minimal, the fallback for text-returning PG-compatible engines is a nice touch, and the comment explains the *why* well. `cargo test explain` is green and the Explain Plan now renders end-to-end.

Two optional, non-blocking nits inline — points 1 and 2 below. Happy to approve either way.

src-tauri/src/drivers/postgres/explain.rs
debba: Yes, it makes sense

NewtTheWolf: This is a test

```suggestion
		Err(json_err) => rows[0].try_get::<_, String>(0).map_err(|e| {
    	log::debug!("EXPLAIN json read failed ({json_err}); text read also failed");
    	format_pg_error(&e)
	})?,
```

NewtTheWolf (changes requested):
Test

PR #285, opened 2026-06-04
fix(grid): truncate large JSON/text cell previews to avoid UI freeze

### LINKED ISSUE #283: Truncate/lazy-render large JSON column values in the data grid for better performance
First of all, thank you for building Tabularis — it's a genuinely great tool and a pleasure to use. The speed, the design, and the overall experience are excellent.

I'd like to report a performance issue I ran into:

When a MySQL table has a `JSON` column whose values are very large, querying it makes the whole app very laggy — even when the query only returns a small result set (e.g. `LIMIT 500`). It seems the grid tries to fully render/process every large JSON value, which freezes the UI.

In practice, fully rendering such a large field directly in the grid isn't very useful anyway — you can't meaningfully read a huge JSON blob inline in a cell.

**Suggestion:** Could the grid render only a truncated preview of large cell values (e.g. the first N characters/bytes), and only load/expand the full value on demand (e.g. when the user clicks to open the JSON inspector)? This lazy/partial rendering should significantly improve performance for tables with large JSON columns.

**Environment:**
- Database: MySQL
- Column type: `JSON` (large values)
- Repro: run a query returning such a column, even with a small `LIMIT`

NewtTheWolf:
Large JSON column values (e.g. MySQL `JSON`) froze the grid: each visible cell tokenized and rendered its **full** stringified value into thousands of DOM nodes — despite cells being clipped to ~300px and the full value already being reachable on demand.

**Fix:** cap the inline preview (`truncateCellPreview`, 300 chars) before tokenization/render in `JsonCell` and `TextCell`, and cap the native `<td>` tooltip. The full value stays available via the inline expander and JSON viewer, which read the raw row value (not `displayText`) — so the cap is lossless.

Adds a ~1 MB big-JSON row to the MySQL and Postgres JSON demos to reproduce.

Verified: `tsc --noEmit`, `eslint`, 154 UI + utils tests green (incl. new `truncateCellPreview` and bounded-preview cases).

Closes #283

### DIFF
--- a/demo/init/mysql/03-json-demo.sql
+++ b/demo/init/mysql/03-json-demo.sql
@@ -43,3 +43,36 @@ INSERT INTO json_demo (label, payload, notes_text) VALUES
   ('unicode + emoji',
    '{"greeting":"こんにちは","emoji":"🦊🐾","math":"∑(α+β)²"}',
    '{"emoji":"⚙️"}');
+
+-- -------------------------------------------------------------
+-- Large payload — reproduces grid lag with big JSON cells (#283).
+-- ~3000 nested objects build a multi-hundred-KB / ~1MB JSON blob.
+-- The grid must stay responsive: only a truncated preview renders
+-- inline, the full value opens in the JSON viewer on demand.
+-- -------------------------------------------------------------
+SET SESSION cte_max_recursion_depth = 100000;
+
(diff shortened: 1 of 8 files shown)


NewtTheWolf:
@debba pr is tested https://github.com/TabularisDB/tabularis/issues/283#issuecomment-4621741785

PR #286, opened 2026-06-04
Derive pagination LIMIT/OFFSET from a SQL parser

debba:
Builds on #275.

### Background

PR #275 fixed the grid pagination rewriter to fold the user's `OFFSET` into the
per-page offset, but it did so by extending the **hand-rolled token scanner** in
`src-tauri/src/drivers/common/query.rs`, which only recognises the trailing
`… LIMIT <n> OFFSET <n>` shape.

That heuristic is fragile. It does **not** understand:
- MySQL's `LIMIT <offset>, <count>` syntax
- `OFFSET` before `LIMIT` (valid in Postgres)
- numeric expressions / placeholders

In those cases the values are read wrong (or dropped) and the stripped base can be
inconsistent with what was extracted, producing a malformed appended query.

### The change

- Add the `sqlparser` crate (v0.62) and read `Query.limit_clause` from the AST,
  parsed with the **correct per-driver dialect** (`PaginationDialect::{MySql,Postgres,Sqlite}`
  threaded through `build_paginated_query`). MySQL's `LIMIT <offset>, <count>` is
  normalised to the same `(limit, offset)` shape.
- Strip the trailing clause at its `LIMIT`/`OFFSET` keyword (reusing the existing
  position-aware tokenizer, which collapses parenthesised subqueries so inner
  `LIMIT`s are never touched), consistent with what the parser saw.
- Render the new pagination clause from a `LimitClause` AST node and concatenate it
  to the **verbatim sliced base**, so leading comments, inline `/*+ hints */`, and
  the body's formatting are preserved (no full-query reserialization).
- Keep the original token scanner as a **fallback** for inputs the parser rejects,
  so behaviour never regresses. `FETCH FIRST … ROWS` is out of scope and defers to
  the fallback.

### Tests

Existing `build_paginated_query` tests updated to pass a dialect; new cases added for
the MySQL comma form (pages 1 & 2), `OFFSET` before `LIMIT`, backtick identifiers,
inline-hint preservation, and the parse-error fallback. `cargo test drivers::common`
— 63 passed.


src-tauri/src/drivers/common/query.rs
```diff
@@ -281,22 +285,182 @@ pub fn extract_user_offset(query: &str) -> Option<u32> {
...
+
+/// Cut the query immediately before its trailing top-level LIMIT/OFFSET clause.
+///
+/// Reuses [`tokenize_sql_with_pos`], which collapses parenthesised groups into
+/// a single token, so a LIMIT/OFFSET inside a subquery is never seen and only
+/// the outer clause is removed. Unlike [`strip_limit_offset`], it cuts at the
+/// keyword regardless of the value shape, so MySQL's `LIMIT <offset>, <count>`
+/// is handled. Falls back to [`strip_limit_offset`] if no trailing clause is
+/// found (kept total; the parser having reported a clause makes this rare).
+fn strip_at_limit_keyword(query: &str) -> String {
```
NewtTheWolf, line 559: **Findings 2 & 4 — strip desync.** When the parser reported a clause (`has_limit_clause = true`) but it isn't plain numeric, this keyword scan *and* the `strip_limit_offset` fallback both fail to strip it, and `build_paginated_query` then appends pagination to the un-stripped query. Confirmed at runtime:

- `… LIMIT ALL` → executed `… LIMIT ALL LIMIT 501 OFFSET 0` → `syntax error at or near "LIMIT"`
- `… LIMIT 2000 -- note` → executed `… LIMIT 2000 -- note LIMIT 501 OFFSET 0` → the `--` comments out the appended `LIMIT 501 OFFSET 0`, so pagination is silently dropped (the same query without the comment strips cleanly to `… LIMIT 501 OFFSET 0`).

`is_pagination_tail_token` doesn't cover `ALL` / comment tokens, and the fallback only strips trailing numeric clauses — so any clause the parser accepts but these heuristics don't recognize desyncs. Since the parser already located the clause, consider cutting at its AST source span when `has_limit_clause` is true, instead of re-tokenizing.

src-tauri/src/drivers/common/query.rs
```diff
@@ -281,22 +285,182 @@ pub fn extract_user_offset(query: &str) -> Option<u32> {
...
+    let statements = Parser::parse_sql(dialect.parser_dialect().as_ref(), query).ok()?;
+    if statements.len() != 1 {
+        return None;
+    }
+    let Statement::Query(q) = &statements[0] else {
+        return None;
+    };
+
+    // FETCH FIRST … ROWS ONLY is out of scope; defer to the fallback path so
+    // its behaviour is unchanged rather than producing a mixed clause.
```
NewtTheWolf: Confirmed at runtime: `… FETCH FIRST 5000 ROWS ONLY` paginates to `… FETCH FIRST 5000 ROWS ONLY LIMIT 501 OFFSET 0` (DB error) — so the fallback **does** produce a mixed clause. Handling FETCH properly is out of scope, but the comment is misleading; fixing it to match:

```suggestion
    // FETCH FIRST … ROWS ONLY is out of scope; defer to the fallback path.
    // NB: the fallback does not strip FETCH, so a paginated FETCH query gets a
    // trailing LIMIT/OFFSET appended (a mixed clause the DB rejects).
```

NewtTheWolf (changes requested):
Validated the parser-based rewriter end-to-end against a seeded `perf_demo` (50k rows, MySQL + Postgres). The headline cases are solid — MySQL `LIMIT <off>, <cnt>`, `OFFSET` before `LIMIT`, backtick identifiers and inline leading hints all paginate correctly across pages. 👍

But I hit four issues where the rewriter mishandles inputs — three confirmed directly via the new "executed query" the UI shows. Requesting changes for the two correctness ones (CTE + the strip desync); the FETCH item is smaller.

**Confirmed at runtime** (executed query in brackets):

1. **CTEs are never paginated.** [`parse_pagination`](https://github.com/TabularisDB/tabularis/blob/3ccec1101bd53c3a99285bed905ff60b1c4c4cc2/src-tauri/src/drivers/common/query.rs#L339) parses the whole statement and matches `Statement::Query`, which already *includes* `WITH … SELECT`. But the call-site gate is still [`is_select_query()`](https://github.com/TabularisDB/tabularis/blob/3ccec1101bd53c3a99285bed905ff60b1c4c4cc2/src-tauri/src/drivers/common/query.rs#L11) — a `starts_with("SELECT")` string check — so a CTE falls through to the non-paginated branch and is hard-truncated at `page_size` with no paging controls. The new parser *could* handle it; the gate blocks it. (all three drivers)

2. **`strip_at_limit_keyword` desyncs from the parser for non-numeric clauses.** When the parser reports `has_limit_clause = true` but the clause isn't plain digits, both the token heuristic and the `strip_limit_offset` fallback fail to strip it, so pagination is appended to the *un-stripped* query:
   - `… LIMIT ALL` → `… LIMIT ALL LIMIT 501 OFFSET 0` → `syntax error at or near "LIMIT"`
   - `… LIMIT 2000 -- note` → `… LIMIT 2000 -- note LIMIT 501 OFFSET 0` → the `--` comments out the appended pagination, which is then **silently dropped** (the same query without the comment strips cleanly to `… LIMIT 501 OFFSET 0`).

   Root cause: once the parser has recognized the clause, stripping re-tokenizes heuristically instead of cutting at the span the parser already computed.

3. **`FETCH FIRST … ROWS` fallback produces a mixed clause.** `… FETCH FIRST 5000 ROWS ONLY` → `… FETCH FIRST 5000 ROWS ONLY LIMIT 501 OFFSET 0` → DB error. Out of scope per the PR — but the inline comment claims it defers "rather than producing a mixed clause", and it does produce one. (suggestion inline)

**Recommendations**
- **(1)** Derive paginatability from the parse, not the string prefix: paginate iff the query parses to a single `Statement::Query` — cleanly includes CTEs/`VALUES`, excludes `SHOW`/`EXPLAIN`/DDL. (Don't widen to `returns_result_set`, which would try to paginate `SHOW`/`EXPLAIN` and append an invalid `LIMIT`.)
- **(2)** When `has_limit_clause` is true, cut at the clause's source span from the parse instead of re-scanning tokens, so `LIMIT ALL`, trailing comments, etc. can't desync.

Details inline.

src-tauri/src/drivers/postgres/mod.rs
```diff
@@ -810,7 +810,12 @@ async fn exec_on_pg_client(
 
     let (final_query, pagination_meta) = if is_select && limit.is_some() {
         let l = limit.unwrap();
-        let data_query = crate::drivers::common::build_paginated_query(query, l, page);
+        let data_query = crate::drivers::common::build_paginated_query(
```
NewtTheWolf, line 816: **Finding 1 — CTEs are never paginated.** This new `build_paginated_query` call is only reached when `is_select` (line 811) is true, and `is_select` = `is_select_query()` = `starts_with("SELECT")`. So a `WITH … SELECT` CTE never reaches the parser path you just added — it falls through to the non-paginated branch and is hard-truncated at `page_size` (confirmed at runtime: no AUTO-PAGINATED badge, can't reach page 2).

`parse_pagination` already matches `Statement::Query`, which *includes* CTEs — so the new parser would handle them; only the string-prefix gate blocks it. Suggest deriving the gate from the parse (`Statement::Query`) instead of `starts_with("SELECT")`. Same gate in `mysql/mod.rs:948` and the sqlite driver.

debba: Fixed in d96ec81f. The gate is now `is_paginatable_query()`: paginate iff the query parses to a single `Statement::Query` — CTEs and `VALUES` get in, SHOW/EXPLAIN/DDL stay out (per your note, it does not widen to `returns_result_set`). `is_select_query()` survives only as the fallback when the parser rejects the input, so dialect quirks sqlparser can't parse keep their old prefix-based routing. All three drivers migrated (postgres, mysql, sqlite).

src-tauri/src/drivers/common/query.rs
debba: Fixed in d96ec81f, taking your span suggestion. When the parser reports a clause, the strip now cuts at the clause's AST source span instead of re-tokenizing — with a walk-back over the leading keyword, since sqlparser's spans anchor at the clause's first *value* (`LIMIT`/`OFFSET` belong to no AST node).

Two wrinkles surfaced while wiring it up:

- A bare `LIMIT ALL` never reaches the AST at all — `parse_optional_limit_clause` folds it into "no limit clause" (sqlparser 0.62, parser/mod.rs:13104) — so there is no span to anchor to. It's located by a trailing-token check instead. `LIMIT ALL OFFSET n` does produce a clause and anchors to the OFFSET value, peeling the dangling `LIMIT ALL` on the way back.
- Queries with locking clauses defer to the fallback (like FETCH), because cutting span-to-end would silently drop `FOR UPDATE` along with the clause.

The heuristic tokenizer also skips `--`/`/* */` comments now, so the parse-error fallback path can't be shielded by a trailing comment either. Both runtime cases from the review are covered by tests, plus `LIMIT ALL`, `LIMIT ALL OFFSET n`, a trailing block comment, and a non-literal `LIMIT 10 + 5`.

src-tauri/src/drivers/common/query.rs
debba: Applied your suggested wording in d96ec81f — the comment now states the fallback does append a mixed clause for FETCH queries.

NewtTheWolf (approved):
Re-reviewed `3ccec110..d96ec81` — both correctness blockers resolved, excellent work @debba.

✅ **CTEs (#1)** — `is_paginatable_query` now derives paginatability from the parse (single `Statement::Query`, fallback to the prefix check), and all three drivers gate on it. Verified on live Postgres: `WITH recent AS (…) SELECT * FROM recent` now paginates instead of being hard-truncated.

✅ **Strip desync (#2)** — cutting at the clause's AST source span (`clause_start`) instead of re-tokenizing fixes `LIMIT ALL`, trailing `--`/`/* */` comments, and expression limits. Confirmed the old `… LIMIT ALL LIMIT 101 OFFSET 0` output errored on PG and the new `… LIMIT 101 OFFSET 0` is clean. The `LIMIT ALL` textual special-case (parser folds it away) and the comment-skipping tokenizer are nicely handled.

👍 Bonus: deferring `FOR UPDATE/SHARE` to the fallback so a span-cut can't silently drop a lock.

`drivers::common` 79/79 green. The corrected comment on the `FETCH FIRST` mixed-clause is the honest call — one optional follow-up: for `FETCH`/locked queries it'd be strictly safer to skip auto-pagination than to append a clause the DB rejects. Non-blocking. LGTM 🚀

PR #287, opened 2026-06-04
Speed up DataGrid scrolling on wide / large tables

debba:
Scrolling a table with ~500 rows and 30-40 columns was painfully slow. The grid only virtualized rows, so every scroll tick re-rendered the whole visible `<tbody>` (≈30 rows × 40 cells) from scratch, and each row was re-measured on the fly with a per-row ResizeObserver.

### What changed

- The per-row render now lives in a `React.memo` component (`DataGridRow`). All the stable, grid-wide dependencies are bundled into one memoized `rowCtx` object, and the volatile per-row bits (`isSelected`, editing/focused/expanded column, …) are passed as plain primitives. During a vertical scroll only the rows that actually changed re-render — the rest bail out of the shallow compare.
- `handleCellDoubleClick`, `handleEditCommit` and `handleKeyDown` are now `useCallback`s (they read the live `editingCell` through a ref), otherwise the memo would never hold across re-renders.
- `formatCellValue` is computed once per cell instead of twice (it was being run again just for the `title` tooltip).
- Rows now have a fixed height and no longer use `measureElement` — same approach `MiniResultGrid` already uses. No more layout thrashing while scrolling.
- The small default-cell render helper moved to `src/utils/dataGridCell.tsx` with unit tests.

Behaviour is unchanged: inline editing, JSON/text expansion, FK/blob buttons, selection highlighting, sticky `#` column and sticky header all work as before.

### Trying it out

There's now a `perf_demo` database in the demo stack with a `wide_table` of 50 columns × 50,000 rows on both MySQL and Postgres, with a mix of column types so the various cell renderers get exercised. Bring up the stack and open `perf_demo.wide_table`:

```
docker compose -f demo/docker-compose.yml up -d
```

The SQL is generated by `demo/generate-perf-sql.py` and the connections are already in `demo/connections.json`.

### Notes

- Column (horizontal) virtualization was intentionally left out for now — with row memoization in place scrolling is fluid, and that change is much riskier (sticky header, sticky `#` column, expansion `colSpan`). Easy follow-up if very wide tables still feel heavy.
- `tsc`, eslint and the full test suite (2508 tests) are green.

Worth profiling with React DevTools on the 50k×50 table to confirm rows that stay on screen don't re-render mid-scroll.


thomaswasle:
Tested with my own much larger table and it feels good. A great improvement. Also the codechanges look good to me.

debba:
@thomaswasle thanks a lot for your feedback 😃

demo/init/mysql/06-perf-wide.sql
```diff
@@ -0,0 +1,137 @@
...
+    id, json_data, long_text, created_at, updated_at, birth_date, is_active, is_verified, amount, balance, score, big_value, attr_01, attr_02, attr_03, attr_04, attr_05, attr_06, attr_07, attr_08, attr_09, attr_10, attr_11, attr_12, attr_13, attr_14, attr_15, attr_16, attr_17, attr_18, attr_19, attr_20, attr_21, attr_22, attr_23, attr_24, attr_25, attr_26, attr_27, attr_28, attr_29, attr_30, attr_31, attr_32, attr_33, attr_34, attr_35, attr_36, attr_37, attr_38
+)
+SELECT
+    g.g,
+    JSON_OBJECT('row', g.g, 'label', CONCAT('item-', g.g), 'tags', JSON_ARRAY('alpha','beta', g.g MOD 7)),
+    REPEAT(CONCAT('Lorem ipsum dolor row ', g.g, '. '), 6),
+    DATE_ADD('2000-01-01 00:00:00', INTERVAL g.g SECOND),
+    DATE_ADD('2010-01-01 00:00:00', INTERVAL (g.g*2) SECOND),
+    DATE_ADD('1970-01-01', INTERVAL (g.g MOD 18000) DAY),
+    (g.g MOD 2),
```
NewtTheWolf: `is_active` is inverted vs the Postgres seed: here `(g.g MOD 2)` is true for **odd** `g`, but `demo/init/postgres/06-perf-wide.sql:80` uses `(g % 2 = 0)` (true for **even** `g`), so the two demo DBs disagree on every row. Please also fix the source at `generate-perf-sql.py:32`.

```suggestion
    (g.g MOD 2 = 0),
```

NewtTheWolf (changes requested):
Nice refactor — scrolling is clearly smoother on a wide 50k-row table. A few cleanup items below, plus two things I want to flag that I **couldn't reproduce locally**, so treat them as "worth a look" rather than confirmed:

**Flagging (not reproduced):**
- **Expanded rows + fixed height** ([`DataGrid.tsx:802`](https://github.com/TabularisDB/tabularis/blob/c41f2b8791d8c8e4337ba54373d4fabef95bdf89/src/components/ui/DataGrid.tsx#L802)): the virtualizer now uses `estimateSize: () => 35` with `measureElement` removed, but an expanded `json_data`/`long_text` editor row is taller than 35px. Its extra height is never reported, so `getTotalSize()` and the `translateY` offsets are computed as if every row is 35px — I'd expect scroll drift / unreachable bottom rows while a row is expanded. `main` covered this via `measureElement`.
- **`rowCtx` rebuilt on selection** ([`DataGrid.tsx:417`](https://github.com/TabularisDB/tabularis/blob/c41f2b8791d8c8e4337ba54373d4fabef95bdf89/src/components/ui/DataGrid.tsx#L417) + [`:1184`](https://github.com/TabularisDB/tabularis/blob/c41f2b8791d8c8e4337ba54373d4fabef95bdf89/src/components/ui/DataGrid.tsx#L1184)): `handleRowClick` depends on `selectedRowIndices` and sits in `rowCtx`'s deps, so selecting a row hands every visible `DataGridRow` a new `ctx` and defeats `React.memo` for the whole viewport. Pure scrolling is unaffected (the headline win is real), but the memo doesn't hold during selection/editing. Reading selection through a ref (like `editingCellRef`) would keep `handleRowClick` stable.

**Cleanup / dedup:**
- **Dead duplicate cell renderer** ([`DataGrid.tsx:738-758`](https://github.com/TabularisDB/tabularis/blob/c41f2b8791d8c8e4337ba54373d4fabef95bdf89/src/components/ui/DataGrid.tsx#L738-L758)) — cells now render via `DataGridRow → renderDefaultCellContent`; `flexRender` only runs for headers, so the inline `cell` body never executes yet still duplicates `formatCellValue` + the null styling. See inline comment.
- **[`MiniResultGrid`](https://github.com/TabularisDB/tabularis/blob/c41f2b8791d8c8e4337ba54373d4fabef95bdf89/src/components/ui/MiniResultGrid.tsx)** still hand-rolls the same null-cell rendering instead of reusing `renderDefaultCellContent`/`formatCellValue` — the extraction is only half done.
- **Dead `isRawSql` plumbing** — `editingIsRawSql` is passed ([`DataGrid.tsx:1331`](https://github.com/TabularisDB/tabularis/blob/c41f2b8791d8c8e4337ba54373d4fabef95bdf89/src/components/ui/DataGrid.tsx#L1331)) into [`DataGridRow`](https://github.com/TabularisDB/tabularis/blob/c41f2b8791d8c8e4337ba54373d4fabef95bdf89/src/components/ui/DataGridRow.tsx) but never read there.

One inline suggestion below for an `is_active` mismatch between the MySQL and Postgres seeds.

NewtTheWolf:
Following up on my review — here are the two **dead-code / dedup** items from the "Cleanup / dedup" section, ready to drop in. Pure deletions (27 lines, no behavior change). `tsc --noEmit` and `eslint` are green and I ran it live — cell rendering and inline editing are unaffected:

- **Dead `cell:` renderer** in `tableColumns` — cells render through `DataGridRow` → `renderDefaultCellContent` and `flexRender` only runs for the header, so this `cell` body never executed (it just duplicated `formatCellValue` + the null styling). Removed it, plus the now-unused `columnTypeMap`/`columnLengthMap` entries in that `useMemo`'s deps. The `accessorFn` stays.
- **Dead `isRawSql` plumbing** — `editingIsRawSql` was passed into `DataGridRow` but never read there. The live write path (`setEditingCell` from the cell's `onChange`) is untouched — only the unused prop, its type, and the `editingCell` field are gone.

Sending it as a diff because I don't have write access to your branch, so I can't push to it directly. Happy to open a PR from my fork instead if that's easier for you.

**To apply** — save the block below as `cleanup.patch` in the repo root, then:
```bash
git apply cleanup.patch
```

````diff
diff --git a/src/components/ui/DataGrid.tsx b/src/components/ui/DataGrid.tsx
index 49dc0ce..a5bdd9e 100644
--- a/src/components/ui/DataGrid.tsx
+++ b/src/components/ui/DataGrid.tsx
@@ -735,27 +735,6 @@ export const DataGrid = React.memo(
                 </div>
               );
             },
-            cell: (info) => {
-              const val = info.getValue();
-              const colName = info.column.id;
-              const colType = columnTypeMap?.get(colName);
-              const formatted = formatCellValue(
-                val,
-                t("dataGrid.null"),
-                colType,
-                columnLengthMap?.get(colName),
-              );
-
-              // The <generated> placeholder logic for auto-increment columns is handled
-              // in the main render loop where we have full context (isInsertion, etc).
-
-              // Apply styling for null values
-              if (val === null || val === undefined) {
-                return <span className="text-muted italic">{formatted}</span>;
-              }
-
-              return formatted;
-            },
           }),
         ),
       [
@@ -764,8 +743,6 @@ export const DataGrid = React.memo(
         t,
         sortClause,
         onSort,
-        columnTypeMap,
-        columnLengthMap,
       ],
     );
 
@@ -1328,7 +1305,6 @@ export const DataGrid = React.memo(
                       pkVal={pkVal}
                       editingColIndex={isRowEditing ? editingCell!.colIndex : null}
                       editingValue={isRowEditing ? editingCell!.value : undefined}
-                      editingIsRawSql={isRowEditing ? editingCell!.isRawSql : undefined}
                       focusedColIndex={isRowFocused ? focusedCell!.colIndex : null}
                       expandedColIndex={isRowExpanded ? expandedCell!.colIndex : null}
                       expandedKind={isRowExpanded ? expandedCell!.kind : null}
diff --git a/src/components/ui/DataGridRow.tsx b/src/components/ui/DataGridRow.tsx
index b3fb096..61318a2 100644
--- a/src/components/ui/DataGridRow.tsx
+++ b/src/components/ui/DataGridRow.tsx
@@ -124,7 +124,6 @@ export interface MemoRowProps {
   pkVal: string | null;
   editingColIndex: number | null;
   editingValue: unknown;
-  editingIsRawSql?: boolean;
   focusedColIndex: number | null;
   expandedColIndex: number | null;
   expandedKind: "json" | "text" | null;
@@ -141,7 +140,6 @@ export const MemoRow = React.memo(function MemoRow(rowCtx: MemoRowProps) {
     pkVal,
     editingColIndex,
     editingValue,
-    editingIsRawSql,
     focusedColIndex,
     expandedColIndex,
     expandedKind,
@@ -190,7 +188,6 @@ export const MemoRow = React.memo(function MemoRow(rowCtx: MemoRowProps) {
           rowIndex,
           colIndex: editingColIndex,
           value: editingValue,
-          isRawSql: editingIsRawSql,
         }
       : null;
   const focusedCell =
````

NewtTheWolf (approved):
lgtm

PR #290, opened 2026-06-05
Make selection of SELECT part of view definition robust.

maacl:
Current selction of SELECT part of view fails in many scenarios, for instance if first AS is surrounded by carriage returns and nort spaces, which is a common way of formatting view definitions. Proposed PR should make this much more robust. Please note the PR is AI generated but reviwed for sanity.

### DIFF
--- a/src/components/modals/ViewEditorModal.tsx
+++ b/src/components/modals/ViewEditorModal.tsx
@@ -60,7 +60,7 @@ export const ViewEditorModal = ({
        * @param sql The full CREATE VIEW SQL string.
        * @returns The underlying SQL query defining the view.
        */
-      export function extractSelectFromView(sql: string): string {
+      function extractSelectFromView(sql: string): string {
         if (!sql || typeof sql !== 'string') return '';
 
         let i = 0;
--- a/src/components/modals/ViewEditorModal.tsx
+++ b/src/components/modals/ViewEditorModal.tsx
@@ -51,10 +51,125 @@ export const ViewEditorModal = ({
       });
       // Extract just the SELECT part for editing
       let selectPart = def;
-      if (def.toUpperCase().includes(" AS ")) {
-        const asIndex = def.toUpperCase().indexOf(" AS ");
-        selectPart = def.substring(asIndex + 4).trim();
+      
+      /**
+       * Reliably extracts the SELECT (or WITH ... SELECT) query from a CREATE VIEW definition.
+       * It uses a lexical scanner to ignore strings, comments, brackets, and parentheses,
+       * ensuring it only splits on the structural "AS" keyword.
+       *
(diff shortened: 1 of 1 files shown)


debba:
Hi @maacl !
Thanks for this contribution, I will review it ASAP.

debba (changes requested):
Hi @maacl, thanks again for this — and sorry for the wait!

I finally sat down with it. The bug is real and the scanner approach is the right way to go: the old `" AS "` search breaks on newline-formatted definitions (SQLite is the worst offender since `sqlite_master` stores the SQL exactly as the user wrote it), and skipping strings/comments/quoted identifiers is exactly what was missing. So the direction is good.

A few things I'd like to see changed before merging though.

The biggest one is the `throw` at the end: it actually breaks Postgres. Our driver uses `pg_get_viewdef()`, which returns *just the SELECT* — no `CREATE VIEW ... AS` prefix. So for a plain view like `SELECT id FROM t;` there's no top-level `AS` at all, and where the old code quietly fell back to the full definition, the new one throws and the user ends up with a load error and an empty editor. Please keep a graceful fallback (`return sql.trim()`) instead.

Related idea: only run the extraction when the input actually starts with `CREATE ... VIEW`, otherwise return it as-is. That also fixes Postgres aliases for free — with `pg_get_viewdef` output like `SELECT count(*) AS total FROM t`, the first depth-0 `AS` is a column alias and the scanner returns `total FROM t`. The old code had the same flaw, but since we're rewriting this anyway, might as well kill it.

Also, please don't strip `WITH CHECK OPTION` / `WITH READ ONLY`. Those clauses round-trip through `alter_view`: the old code kept them in the editable text so saving preserved them. With the strip, someone editing a view with `CHECK OPTION` silently loses it on save — that's the kind of thing nobody notices until it bites.

Last structural thing: the function shouldn't live inside the `useCallback` body (it gets re-created on every render and can't be tested there). In this repo utilities go in `src/utils/` (e.g. `src/utils/sql.ts`) and must come with unit tests in the same PR. This one is a perfect candidate for a table-driven test: MySQL `SHOW CREATE VIEW` with `DEFINER` and backticks, SQLite multiline `AS`, bare `pg_get_viewdef` output, `AS` inside strings/comments, SQL Server brackets, column lists... And honestly, since the code is AI-generated, tests are also the best way to lock in what you reviewed.

Couple of nits, no action needed: backslash escapes inside double-quoted identifiers aren't a thing in Postgres/ANSI (quotes are doubled there), `#` comments are MySQL-only, and there's a trailing-whitespace line right after `let selectPart = def;`. None of these matter in practice.

Happy to merge once the fallback, the utils extraction and the tests are in. Thanks for the patience!

PR #295, opened 2026-06-06
fix(autocomplete): suggest clause keywords and correct columns after …

### LINKED ISSUE #293: [Bug]: No keyword autocomplete (WHERE / ORDER BY / ...) after SELECT * FROM <table>
### Describe the bug

Once a `FROM <table>` is present in the statement, the SQL editor only suggests column names — clause keywords like `WHERE`, `ORDER BY`, `GROUP BY`, `LIMIT` are never offered. Typing the start of such a keyword shows no completion at all.

### To Reproduce

1. In the SQL editor, type `SELECT * FROM users ` (using a real table).
2. Type `w`, expecting a `WHERE` suggestion.
3. No suggestion appears.

**Expected:** clause keywords valid at this position (`WHERE`, `ORDER BY`, `GROUP BY`, `LIMIT`, …) are suggested — typing `w` completes to `WHERE`.
**Actual:** nothing is suggested. Because a table is in scope, the completion list is filled with that table's columns, none of which start with `w`, so Monaco filters everything out.

### Likely cause

Keyword suggestions are gated behind "only if there are no context columns", so as soon as a `FROM` table resolves they are dropped entirely:

- [`autocomplete.ts` — keyword suggestions gated on `contextColumnSuggestions.length === 0`](https://github.com/TabularisDB/tabularis/blob/4bd6e1dd4a7e7d5d5a3ebc24a930c19c3cfca4b3/src/utils/autocomplete.ts#L236-L247)

They should be offered alongside columns (ideally context-aware: clause keywords after a complete `FROM` clause).

### OS Version

Linux (CachyOS)

### Tabularis Version

v0.13.0

thomaswasle:
closes #293.

## Summary

  - **Keywords never suggested after FROM clause** — The completion provider only
  offered keyword completions when no context columns were available. Removed the
  `contextColumnSuggestions.length === 0` guard so clause keywords (WHERE, ORDER BY,
  GROUP BY, LIMIT, …) are always included alongside column and table suggestions.

  - **SQL keywords captured as table aliases** — `parseTablesFromQuery` matched `FROM
  users WHERE` and registered `WHERE` as the alias, poisoning the alias map. Added a
  `SQL_RESERVED` set; any captured alias that is a reserved word is discarded and the
  table name is used as its own alias.

  - **Wrong columns in MySQL multiDb mode** — When `activeSchema` is null (multiDb),
  `getTableColumns` was called without a schema, causing MySQL to fall back to
  `information_schema` and return system columns instead of the user's table columns.
  Fixed by adding an optional `schema?: string` field to `TableInfo` and populating it
  with the source database name when building `effectiveTables` in multiDb mode. Both
  the context-aware and dot-trigger paths now pass `t.schema ?? schema`.

  - **Column cache poisoning** — The fallback that attempts column fetches before the
  tables list loads could receive an empty response (schema mismatch / connection not
  yet ready) and cache it, blocking all subsequent fetches for that table. Empty results
  are now never written to the cache.

  ## Files changed

  | File | Change |
  |---|---|
  | `src/utils/autocomplete.ts` | Remove keyword gate; add schema-aware table lookup for dot-trigger; use `t.schema` in context-aware path; skip caching empty results |
  | `src/utils/sqlAnalysis.ts` | Add `SQL_RESERVED` set; filter captured aliases against it |
  | `src/contexts/DatabaseContext.ts` | Add optional `schema?: string` to `TableInfo` |
  | `src/pages/Editor.tsx` | Annotate each table with its source database when building  `effectiveTables` in multiDb mode |

  ## Test plan

  - [ ] MySQL multiDb: type `SELECT * FROM <table> ` → columns from the correct database
  appear
  - [ ] MySQL multiDb: type `SELECT * FROM <table> WHERE ` → clause keywords (AND, OR,
  …) and column names both appear
  - [ ] Type `SELECT * FROM <table> WHERE <partial>` → matching column names filter
  correctly
  - [ ] Dot-trigger `table.` → columns from the correct per-database table appear
  - [ ] Single-database connections: keywords still appear; dot-trigger still works
  - [ ] Non-MySQL drivers: no regression in keyword or column suggestions

### DIFF
--- a/src/utils/autocomplete.ts
+++ b/src/utils/autocomplete.ts
@@ -62,7 +62,7 @@ const getTableColumns = async (connectionId: string, tableName: string, schema?:
       tableName,
       ...(schema ? { schema } : {}),
     });
-    
+
     if (!Array.isArray(cols)) {
       console.warn(`get_columns returned non-array for table ${tableName}`, cols);
       return [];
@@ -74,12 +74,16 @@ const getTableColumns = async (connectionId: string, tableName: string, schema?:
       detail: c.data_type,
     }));
 
-    columnsCache.set(cacheKey, {
-      data: simpleCols,
-      timestamp: Date.now()
-    });
-    
-    cleanupCache();
+    // Don't cache empty results — they likely reflect a transient failure
+    // (connection not ready, schema mismatch) rather than a genuinely empty table.
+    // An uncached miss retries the backend on the next trigger, which is cheap.
+    if (simpleCols.length > 0) {
+      columnsCache.set(cacheKey, {
+        data: simpleCols,
+        timestamp: Date.now()
+      });
+      cleanupCache();
+    }
     return simpleCols;
   } catch (e) {
     console.error(`Failed to fetch columns for autocomplete: ${tableName}`, e);
@@ -110,7 +114,7 @@ export const registerSqlAutocomplete = (
       if (!connectionId) return { suggestions: [] };
 
       const wordUntil = model.getWordUntilPosition(position);
-      
+
       const range = {
         startLineNumber: position.lineNumber,
         endLineNumber: position.lineNumber,
@@ -129,6 +133,7 @@ export const registerSqlAutocomplete = (
       // Get current statement context
       const currentStatement = getCurrentStatement(model, position);
       const tableAliases = parseTablesFromQuery(currentStatement);
+
 
       // ============================================
       // 1. DOT TRIGGER (table.column or alias.column)
@@ -149,7 +154,8 @@ export const registerSqlAutocomplete = (
         }
         
         if (actualTableName) {
-          const columns = await getTableColumns(connectionId, actualTableName, schema);
+          const foundTable = tables.find(t => t.name.toLowerCase() === actualTableName.toLowerCase());
+          const columns = await getTableColumns(connectionId, actualTableName, foundTable?.schema ?? schema);
           
           // Calculate range for column name after dot
           const columnRange = {
@@ -187,8 +193,11 @@ export const registerSqlAutocomplete = (
       if (tableAliases && tableAliases.size > 0) {
         // User is inside a query with FROM/JOIN - suggest columns from those tables
         const tableNames = Array.from(new Set(tableAliases.values()));
+        // Prefer the TableInfo from the loaded list (original casing, full metadata).
+        // Fall back to a minimal synthetic entry from the parsed name so columns can
+        // still be fetched when the tables list hasn't finished loading yet.
         const matchingTables = tableNames
-          .map(name => tables.find(t => t.name.toLowerCase() === name.toLowerCase()))
+          .map(name => tables.find(t => t.name.toLowerCase() === name.toLowerCase()) ?? { name })
           .filter(Boolean) as TableInfo[];
         
         // Limit parallel fetches to prevent memory spikes
@@ -198,7 +207,7 @@ export const registerSqlAutocomplete = (
         }
         
         const results = await Promise.all(
-          matchingTables.map(t => getTableColumns(connectionId, t.name, schema))
+          matchingTables.map(t => getTableColumns(connectionId, t.name, t.schema ?? schema))
         );
         
         const seenColumns = new Set<string>();
@@ -233,24 +242,15 @@ export const registerSqlAutocomplete = (
       }
 
       // ============================================
-      // 3. KEYWORD SUGGESTIONS (only if no context columns)
+      // 3. KEYWORD SUGGESTIONS
       // ============================================
-      let keywordSuggestions: Array<{
-        label: string;
-        kind: number;
-        insertText: string;
-        range: { startLineNumber: number; endLineNumber: number; startColumn: number; endColumn: number };
-        sortText: string;
-      }> = [];
-      if (contextColumnSuggestions.length === 0) {
-        keywordSuggestions = SQL_KEYWORDS.map((kw) => ({
-          label: kw,
-          kind: monaco.languages.CompletionItemKind.Keyword,
-          insertText: kw,
-          range,
-          sortText: `2_${kw}`
-        }));
-      }
+      const keywordSuggestions = SQL_KEYWORDS.map((kw) => ({
+        label: kw,
+        kind: monaco.languages.CompletionItemKind.Keyword,
+        insertText: kw,
+        range,
+        sortText: `2_${kw}`
+      }));
 
       // ============================================
       // 4. TABLE SUGGESTIONS
--- a/src/utils/autocomplete.ts
+++ b/src/utils/autocomplete.ts
@@ -1,15 +1,17 @@
 import type { Monaco } from "@monaco-editor/react";
 import { invoke } from "@tauri-apps/api/core";
 import type { TableInfo } from "../contexts/DatabaseContext";
-import { getCurrentStatement, parseTablesFromQuery } from "./sqlAnalysis";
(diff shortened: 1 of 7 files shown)


NewtTheWolf:
uhhh thx @thomaswasle i will take a look today! @debba u can assign it to me

src/utils/sqlAnalysis.ts
```diff
@@ -1,29 +1,41 @@
 // SQL Analysis Utilities - Pure logic functions for parsing and analyzing SQL
 
+// SQL reserved words that are never valid unquoted table aliases
+const SQL_RESERVED = new Set([
+  'where', 'on', 'set', 'having', 'group', 'order', 'limit', 'offset',
+  'join', 'left', 'right', 'inner', 'outer', 'cross', 'union', 'intersect',
+  'except', 'select', 'from', 'into', 'values', 'update', 'delete', 'insert',
+  'create', 'drop', 'alter', 'and', 'or', 'not', 'is', 'in', 'between',
+  'like', 'as', 'distinct', 'case', 'when', 'then', 'else', 'end', 'by',
+  'asc', 'desc', 'null', 'with', 'exists', 'all', 'any', 'using',
```
NewtTheWolf: **Blocking (cheap):** this set misses a few keywords that can legally follow a table name, so they're still captured as aliases — the exact bug it's meant to prevent:

- `SELECT * FROM orders FOR UPDATE` → `for` registered as alias of `orders`
- `FROM t NATURAL JOIN u` → `natural` as alias of `t`
- also `FETCH` (FETCH FIRST), `WINDOW`, `LATERAL`, `STRAIGHT_JOIN` (MySQL), `TABLESAMPLE`, `QUALIFY`

Adding them completes the fix:

```suggestion
  'asc', 'desc', 'null', 'with', 'exists', 'all', 'any', 'using',
  'for', 'fetch', 'natural', 'window', 'lateral', 'straight_join', 'tablesample', 'qualify',
```

(Longer term, a negative-lookahead in the alias capture would remove the need to maintain this list at all — but that's non-blocking.)

src/utils/autocomplete.ts
```diff
@@ -74,12 +74,16 @@ const getTableColumns = async (connectionId: string, tableName: string, schema?:
...
-    columnsCache.set(cacheKey, {
-      data: simpleCols,
-      timestamp: Date.now()
-    });
-    
-    cleanupCache();
+    // Don't cache empty results — they likely reflect a transient failure
+    // (connection not ready, schema mismatch) rather than a genuinely empty table.
+    // An uncached miss retries the backend on the next trigger, which is cheap.
+    if (simpleCols.length > 0) {
```
NewtTheWolf: **Non-blocking** — the rationale in your comment is sound (empty ≈ transient/not-ready, retry is cheap) and the cache already has a TTL on success, so skipping empties is a reasonable trade-off.

One edge case to keep in mind: a *genuinely* empty relation (a 0-column view, or a table the backend can't introspect) is never cached, so it re-issues `get_columns` on every trigger that references it. Almost certainly fine given how rare that is — but if it ever shows up as IPC chatter, distinguishing "empty because not-ready" from "empty because actually-empty" (or a short negative-cache TTL) would bound it. Fine to leave as a follow-up issue.

NewtTheWolf (changes requested):
Tested this end-to-end on the demo stack and all four fixes work: in MySQL multiDb the columns now resolve to the real table instead of `information_schema`, clause keywords show up after `FROM`, `WHERE` is no longer captured as an alias, and partial-filtering + the dot-trigger both behave. No regression on Postgres. Nice fix 👍

One small thing I'd want completed before merge, plus a few non-blocking notes.

**Requesting changes**
- **`SQL_RESERVED` is incomplete** — it covers the common clause keywords but misses a few that can legally follow a table name, so they're still captured as aliases (the exact bug this set is meant to prevent): `SELECT * FROM orders FOR UPDATE` → `for` becomes an alias of `orders`; `FROM t NATURAL JOIN u` → `natural`. One-line addition — inline suggestion below.

**Non-blocking** (happy to spin any of these into separate issues instead of holding the PR):
- **Denylist vs. parse depth** — `SQL_RESERVED` patches a greedy regex that grabs the first word after the table as the alias. A negative-lookahead / clause-boundary parse would make the denylist unnecessary (no perpetual keyword maintenance). The completion above fixes the immediate case; this is the longer-term shape.
- **multiDb same-name tables** — `tables.find(t => t.name === …)` returns the first match, so if the same table name exists in two selected DBs the first DB's schema always wins, and `db.table.` qualifiers aren't parsed. Still strictly better than before (which returned no columns at all) — worth a follow-up for cross-DB disambiguation.
- **Empty-result cache skip** — deliberate per your comment (see inline); just one rare edge case to consider.
- **Keyword clutter / duplicate labels** — keywords now always render, so a column named like a keyword (`count`, `name`, `value`) shows twice (Field + Keyword). Relies entirely on `sortText` ordering. UX tuning.
- **Three keyword lists** — `SQL_RESERVED` overlaps `SQL_KEYWORDS` (`autocomplete.ts`) and the sets in `sqlSplitter/classify.ts`; a shared source would avoid drift.

Inline details below.

src/utils/sqlAnalysis.ts
thomaswasle: @NewtTheWolf  I decided to skip this and directly went for the long term solution.

NewtTheWolf:
Hey @thomaswasle, thanks for the quick turnaround — and the extra work on cross-DB disambiguation, the keyword/column dedup and the negative-cache TTL is genuinely nice, I want to keep all of that.

I think there was a small mix-up on the blocking point, though. The ask was to *complete* the keyword set (add `for`, `natural`, `fetch`, …). The "negative-lookahead, longer-term" note was a separate, non-blocking idea. The new `extractFromSection` approach goes in that spirit, but as implemented it **removed** the keyword guard entirely — so it doesn't actually solve the flagged cases, and it introduced a regression:

```
FROM orders FOR UPDATE   → { for: orders }      (still captured — the blocking case)
FROM t NATURAL JOIN u    → { natural: t, … }    (still captured)
FROM users JOIN orders   → { join: users }      (regression: "join" becomes the alias, "orders" dropped)
FROM a LEFT JOIN b        → { left: a, b: b }    (regression: "left" becomes an alias)
```

Root cause: the alias-capture group still has nothing stopping it from grabbing a keyword that directly follows a table name — and once it grabs e.g. `join`, that token is consumed and the next table is lost. A post-filter denylist can't fix that (token already eaten); the negative-lookahead can, because it stops the keyword from being matched at all.

To save us a review round-trip I went ahead and pushed a small commit directly to the branch: a negative-lookahead in the alias regex (the long-term shape) + regression tests for the keyword-as-alias and schema-qualified-JOIN cases. Everything you built stays — only the one regex line changed.

Verified locally on the demo stack: `vitest` 2510 passed, `tsc` clean, lint clean, `pnpm build` ok, and the four cases above now resolve correctly (unaliased and `db.table.` joins included). Hope that's alright — shout if you'd rather have driven it yourself.

NewtTheWolf (approved):
CI green (test + Kilo Code Review both pass). The alias-capture fix is in: keywords (JOIN/LEFT/NATURAL/FOR/…) are no longer captured as aliases, the unaliased/qualified-JOIN regression is gone, and the originally flagged cases now resolve correctly — with regression tests covering them. All your cross-DB / dedup / negative-cache work is preserved. Thanks @thomaswasle! 🚀

debba:
Very great work @thomaswasle and thanks @NewtTheWolf for the review.
We can go ahead on merging if you are ready!

thomaswasle:
Yes. Go ahead! Unless @NewtTheWolf has any concerns?

PR #299, opened 2026-06-07
feat(plugins): hosted Tabularium registry + connection catalogue (fully BC)

NewtTheWolf:
## What

Wires up the hosted Tabularium registry (registry.tabularis.dev) and rebuilds
the New Connection flow around a driver catalogue — without breaking the old
registry format.

70 files · +6311 / −314

## Backend

- **Hosted registry + merge** — new HTTP API catalogue merged with the legacy
  registry (union, API wins on id conflict), so old plugins (duckdb, …) stay
  visible. (`registry.rs`, `commands.rs`)
- **Release integrity** — Ed25519 JWS verification against the registry's JWKS,
  with release-identity guards; the verified hash is what the installer checks
  the downloaded asset against. Unsigned legacy releases warn rather than
  refuse, since no live release is signed yet. (`integrity.rs`)
- **Deep-link install** — `tabularis://install` handler, version-aware via
  semver: Install / Update / already-installed. (`deep_link.rs`)
- **Tracked downloads** — assets resolve through the registry redirect endpoint
  (`/latest` unpinned, `/releases/{version}` pinned) so installs get counted.
  (`tabularium.rs`)
- **Removable BC layer** — all compat code in `compat.rs`, tagged
  `COMPAT(registry-ga):` at every call site. Legacy `registry.json` install +
  listing and legacy `manifest.json` still work; deletable once all plugins
  have migrated.

## Frontend

- **Connection catalogue** as step 1 of New Connection: search, paradigm
  facets, engine cards, version picker, install-gated uninstalled drivers with
  inline install. (`ConnectionCatalogue`, `EngineCard`, `InstallGate`, …)
- **PluginInstallConfirmModal** renders the deep-link action (install / update /
  already-installed) and badges the release's signature status.
- **Plugins tab** — the Installed tab now offers the update it advertises,
  gated on the release actually being installable (platform + app version), not
  just on a newer version existing.
- i18n across all 8 locales.

## Registry 0.13 compatibility

0.13 gates `GET /api/plugins` on a manifest having resolved at ingest, while
`/{slug}` stays ungated — a plugin can be invisible in the catalogue but still
resolvable. Installed-but-unlisted plugins are re-resolved by slug so they keep
their update path; dropping out of the list is not a deletion. Unknown slugs are
skipped, and when the API is down we don't fire lookups that cannot succeed.

`manifest_raw` needs no client work: no forge round-trip exists to drop, and the
`null` (unsigned legacy) branch is already handled — which is the only branch
live data currently exercises.

## Fixes found while testing against the live registry

- Legacy-only plugins were stamped with the API base, so their card linked to a
  `<base>/plugins/<id>` that 404s instead of falling back to their homepage.
- `has_manifest` only knew `.tabularium` while `read_manifest` already had the
  legacy fallback, so `manifest.json` bundles were rejected on install *and*
  silently skipped by `list_installed`. Ten of the thirteen catalogue plugins
  still ship `manifest.json`.
- A picked non-latest version survived its own install and pinned the card:
  neither "up to date" nor updatable, with the picker hidden — no way out.

## Compatibility

Fully backwards-compatible: the legacy registry format and `manifest.json` are
still consumed. The in-repo `plugins/registry.json` seed is replaced by
`tabularium-extensions.schema.json`; the legacy format is read from the hosted
source via the BC layer.

## Tests

Backend 940 ✓ · Frontend 2867 ✓ · typecheck + lint clean

## Not covered

- **Signed installs are untested against real data.** No live plugin on
  registry.tabularis.dev returns a `jws` — every install currently takes the
  unsigned legacy branch. The JWS path has unit coverage (tampered payload,
  wrong identity, unknown kid, malformed JWS) but no live exercise. This is a
  re-ingest problem on the registry side, not a client one.
- Two published plugins ship inconsistent assets (`cloudflare-d1@0.1.0` has no
  manifest in its archive; `redis@0.4.1` ships a `0.4.0` manifest). Both are
  publisher-side; the client reports them correctly.


src-tauri/src/plugins/compat.rs
```diff
@@ -0,0 +1,384 @@
...
+
+use std::path::Path;
+
+use crate::plugins::registry::{self, PluginRegistry};
+
+/// Legacy default registry: the flat `registry.json` hosted on GitHub that
+/// `main` shipped before the Tabularium API cutover. Used only as a
+/// last-resort fallback (see `resolve_registry`).
+pub const LEGACY_REGISTRY_URL: &str =
+    "https://raw.githubusercontent.com/TabularisDB/tabularis/main/plugins/registry.json";
```
debba, line 27: This constant points at `plugins/registry.json` on `main`, but this PR deletes that file — once merged, the URL will 404. That breaks two things: (1) the legacy half of `resolve_registry`'s merge always fails, so plugins not yet migrated to the Tabularium API silently disappear from the catalogue (the exact scenario this compat layer exists to prevent); (2) already-shipped app versions use this file as their *only* registry source, with no fallback.

Let's keep the old `registry.json` in place for now and prune it only once all plugins have completed the migration (i.e. at the same time the whole `COMPAT(registry-ga)` layer is removed).

src/utils/connectionCatalogue.ts
```diff
@@ -0,0 +1,163 @@
...
+      displayName: representative.name,
+      primaryParadigm,
+      secondaryParadigms: [...allParadigms],
+      drivers: list,
+      installed: list.some((d) => d.installed),
+      verified: list.some((d) => d.verified),
+      downloads,
+    });
+  }
+  return groups.sort((a, b) => a.displayName.localeCompare(b.displayName));
```
debba: This sort is purely alphabetical, and since the section order in `ConnectionCatalogue.tsx` is just Map-insertion order over this list, the paradigm sections end up in an emergent order (currently Other → Document → SQL, because Clickhouse happens to sort first).

I'd show the **builtin drivers first** (MySQL, PostgreSQL, SQLite), then sort the remaining engines alphabetically. As a side effect the SQL section would also become the first one, which is the most common case for users — instead of opening the modal on a wall of "Other" entries (which are mostly legacy-registry plugins that simply lack `paradigms` metadata).

Something like:

```ts
return groups.sort((a, b) =>
  Number(b.drivers.some(d => d.isBuiltin)) - Number(a.drivers.some(d => d.isBuiltin)) ||
  a.displayName.localeCompare(b.displayName)
);
```

src-tauri/src/plugins/compat.rs
NewtTheWolf: Fixed in b705383. Restored `plugins/registry.json` instead of deleting it — `LEGACY_REGISTRY_URL` points at it on `main`, so the deletion would have 404ed post-merge (un-migrated plugins dropping silently out of `resolve_registry`'s merge, plus already-shipped apps losing their only registry source). It now stays until the whole `COMPAT(registry-ga)` layer is pruned. As a side effect this also resolved the `modify/delete` merge conflict with `main`, which had modified the same file.

src/utils/connectionCatalogue.ts
NewtTheWolf: Fixed in b705383. The sort is now builtins-first → other-last → alphabetical: builtin engines (MySQL/PostgreSQL/SQLite) lead so the SQL section comes first, and engines without `paradigms` metadata (the legacy-registry `'other'` bucket) sink to the end. Resulting section order is SQL → Document/NoSQL/Key-Value (alphabetical) → Other.

NewtTheWolf:
@debba small follow-up on this branch (2cbb015) — I aligned `packages/` with the new `.tabularium` host contract and added tooling so plugin authors can keep up:

**`@tabularis/create-plugin` (0.1.1 → 0.2.0)**
- Scaffold now emits `.tabularium` instead of `manifest.json`. `id`=slug / `name`=display are deliberately kept, because the host falls back to `name` as the identity when `id` is absent (otherwise the display name would become the slug).
- New command **`migrate [path]`**: converts existing `manifest.json` plugins → `.tabularium`, drops a *redundant* `id` (only when `id===name`), and refuses without a semver `version` (registry requirement). `justfile`/`README`/`release.yml` references are carried along.
- **`migrate --ci`**: additionally regenerates `release.yml` from the registry-ready template (BIN_NAME derived from `manifest.executable`). Without `--ci` the CI structure is left untouched — only its `manifest.json` reference is renamed so the build keeps working.

**Registry-ready release workflow (template)**
The previous template workflow only bundled `.tabularium` *inside* the platform zips. But the hosted Tabularium registry resolves the manifest from the **release assets** (since 0.8.0), not from the git ref. The workflow now stashes the artifacts and, in a separate `release` job, uploads the zips **plus `.tabularium` as a standalone asset** (modeled on my firestore plugin CI). I deliberately left out the Codeberg publish — that's a hosting choice, not generic.

**`@tabularis/plugin-api`**: doc comments updated to reference `.tabularium` only.

Tests: 32 green (8 new for `migrate`, incl. `--ci` regeneration + edge cases), `tsc`/build clean, generated YAML validated. Happy to hear if the `migrate` defaults (id handling, CI only with `--ci`) look right to you.

debba:
That's awesome! Thanks @NewtTheWolf !

PR #304, opened 2026-06-08
feat(notebook): manage saved notebooks per connection

### LINKED ISSUE #288: [Bug]: SQL Notebooks cannot be renamed.
### Describe the bug

SQL Notebooks are automatically created with default names/titles such as:

- Notebook
- Notebook 2
- Notebook 3
- etc.

However, I cannot find any way to rename them.

I tried:

- Double-clicking the notebook tab title.
- Right-clicking the notebook tab.
- Searching through the application menus.
- Pressing Enter after selecting the tab title.
- Creating additional notebooks to see if a naming prompt appears.

The context menu only shows tab management actions:

- Convert to Console
- Close Tab
- Close Other Tabs
- Close Tabs to Right
- Close Tabs to Left
- Close All Tabs

There is no Rename option and no apparent way to assign a custom name.

---

I also tried exporting a notebook, saving it with a custom filename and location, and then importing it back into Tabularis.

Even after importing the notebook, the tab title is still assigned a default generated name such as:

- Notebook
- Notebook 2
- Notebook 3

The filename used during export/import does not appear to be reflected in the notebook tab title.

This makes me wonder whether notebook names are currently unsupported or if the rename functionality is missing from the UI.

### To Reproduce

1. Open Tabularis.
2. Open any database.
3. Create a new SQL Notebook.
4. Create one or more additional SQL Notebooks.
5. Notice that they are automatically named:
   - Notebook
   - Notebook 2
   - Notebook 3
   - etc.
6. Attempt to rename a notebook by:
   - Double-clicking the notebook tab title.
   - Right-clicking the notebook tab.
   - Looking through the available menus.
7. Observe that no rename functionality is available.

Image

### OS Version

macOS 14.8.2

### Tabularis Version

v0.13.0

### Relevant Log Output

```shell

```

### LINKED ISSUE #294: [Feat]: Better management and reload of notebooks
### Prerequisites

### 🚀 Feature Proposal

After a notebook has been closed, there seems to be no easy way to get back to it, other then reimporting it from %APPDATA%\tabularis\notebooks (also differs per platform). In conjunction with #288 it would be great if a previously named notebook, say 'MyNotebook123' could simply be opened again in a new session by picking from a list of saved notebooks.

As it is, with the workaround if importing it again, the name of the file does not provide any clue as to what it was all about.

### Motivation

Better user experience. Notebooks are great, but it feels like they are not managed fully yet.

### Example

_No response_

### LINKED ISSUE #305: [Feat]: Undo/redo and an editable history for notebooks
### Prerequisites

### 🚀 Feature Proposal

Notebooks autosave on every change, so there's no way to take back an edit. If you delete a cell, rewrite a query or change parameters by mistake, the previous version is gone.

It would help to keep a per-notebook edit history with:

- **Undo / redo** for document changes (cell content, add/remove/reorder cells, names, parameters, stop-on-error), via toolbar buttons and the usual keyboard shortcuts.
- A **history view** listing the recorded steps, so you can see what changed and jump straight back to a specific point instead of pressing undo repeatedly.

Running a cell and its results shouldn't count as history steps — only actual edits to the document.

### Motivation

Autosave is convenient but unforgiving. A history with undo/redo (and the ability to jump to a point) makes editing notebooks much safer and matches what people expect from any editor.

### Example

_No response_

debba:
## What

Notebooks used to be write-only files under `notebooks/` — once a notebook tab was closed there was no way back to it except re-importing the file by its opaque name. This PR makes saved notebooks first-class and manageable, scoped to each connection, and adds editing safety on top of autosave.

## Changes

**Storage**
- Notebooks are stored per connection at `notebooks/<connectionId>/<id>.tabularis-notebook`, so listing a connection's notebooks is a single directory read.
- Legacy flat notebooks are migrated into the per-connection folder lazily, the first time they're opened.

**Sidebar**
- New **Notebooks** tab listing the active connection's saved notebooks (search + last-modified time).
- Double-click to open (focuses the tab if already open).
- Right-click context menu: **Rename**, **Export**, **Save as HTML**, **Import**, **Delete** (delete asks for confirmation).
- The list refreshes automatically when a notebook is created, imported, renamed or deleted — a newly created notebook shows up immediately.

**Rename**
- Notebooks can be renamed directly from their editor tab (double-click the title or use the tab context menu), and from the sidebar.

**History & undo/redo**
- Per-notebook edit history of document changes (cell content, add/remove/reorder, names, parameters, stop-on-error). Running a cell and its results never create history entries; rapid typing is coalesced into a single step.
- Undo/redo from the toolbar and the keyboard (Cmd/Ctrl+Z, Cmd/Ctrl+Shift+Z, Ctrl+Y). Shortcuts are scoped to the active tab and defer to a focused cell editor so its own text undo still works.
- A history panel lists the recorded steps so you can jump straight to a specific point instead of undoing one step at a time.

**Misc**
- "Convert to Console" is no longer offered on notebook or visual query tabs (it only makes sense for table tabs).
- New strings are translated across all locales.

## Backend

- New `list_notebooks` and `rename_notebook` commands; existing create/save/load/delete now take the owning connection id.
- Path-traversal guards on ids; rename patches only the title in place without touching cells.

## Tests

- Rust: unit tests for the notebook store covering round-trip, listing, rename, delete, legacy migration and id validation.
- Frontend: store tests for the per-connection save path and change notifications; pure unit tests for the undo/redo history (signature, coalescing, limit, undo/redo round-trips).

Closes #294
Closes #288
Closes #305


NewtTheWolf:
One small UX note on cell reordering (non-blocking, nice-to-have): right now when you drag a cell it goes to 40% opacity in place and the target cell gets a 2px blue top-border. It works, but it's a bit hard to tell *where* the cell will land — you don't really feel like you're holding anything.

Two cheap improvements that would make it much clearer:

1. **A "held" drag preview** — set a custom drag image (`e.dataTransfer.setDragImage(...)` in `handleDragStart`) so a small card/ghost of the cell follows the cursor. Makes it feel like you're carrying the cell rather than just fading it out in place.
2. **A clear insertion indicator** — instead of a top-border on the hovered cell, render a prominent "drop line" in the gap *between* cells at the exact landing position. That removes the ambiguity of "does it go above or below this one?".

Both are doable on top of the current native HTML5 DnD without a new dependency. If we ever want buttery animated reordering (cells sliding out of the way), `dnd-kit` would do it — but that's a heavier change and probably overkill for this PR; the two tweaks above get 90% of the feel.

Totally fine as a follow-up — just flagging it while it's fresh.

debba:
Done! Added the held drag preview so you actually feel like you're carrying the cell, and replaced the top-border with a proper drop line that sits in the gap and follows whether you're dropping above or below the target. While I was at it I also added edge auto-scroll during the drag, otherwise a very tall cell would trap you with no way to reach the edge to scroll past it.

Really good feedback, thanks for flagging it. I took the chance to also tidy up the notebook toolbar a bit: add and export are dropdowns now and the labels collapse gracefully on a narrow window.

All pushed to the branch.

PR #309, opened 2026-06-09
feat(connection): show SSL tab for plugin drivers via supports_ssl capability

Aditeya:
## Summary

Enables the SSL/TLS configuration tab in the connection modal for plugin drivers (e.g. ClickHouse), which was previously restricted to the built-in `mysql` and `postgres` drivers.

Fixes the app-side half of TabularisDB/tabularis-clickhouse-plugin#1. Companion plugin PR: TabularisDB/tabularis-clickhouse-plugin#2 (both are needed for end-to-end SSL).

## Problem

The SSL tab was gated by a hardcoded driver-ID allow-list in `NewConnectionModal.tsx`:

```tsx
...((driver === "mysql" || driver === "postgres") && isNetworkDriver
  ? [{ id: "ssl", label: "SSL" }] : []),
```

There was no capability a plugin could set to opt in, so external network drivers like ClickHouse could never show SSL configuration — even though `ConnectionParams` (`ssl_mode`/`ssl_ca`/`ssl_cert`/`ssl_key`) is already forwarded to plugins over JSON-RPC.

## Changes

- Add a `supports_ssl` capability to `DriverCapabilities` (Rust `driver_trait.rs` + TS `plugins.ts`), defaulting to `false` so existing manifests are unaffected (accepts `supportsSsl` alias for plugin authors).
- Set `supports_ssl: true` for built-in `postgres` and `mysql` (Rust manifests + `useDrivers.ts` fallbacks); `false` for `sqlite`.
- Gate the SSL tab on `activeDriver?.capabilities?.supports_ssl && isNetworkDriver` instead of the hardcoded ID list.
- Add ClickHouse-appropriate `ssl_mode` options (`disable`/`require`) and make the tab description driver-agnostic (no longer says "MySQL").
- Document the new capability in `plugins/manifest.schema.json` (required — `capabilities` uses `additionalProperties: false`) and `plugins/PLUGIN_GUIDE.md`.

A plugin opts in by adding `"supports_ssl": true` to its manifest `capabilities`.

## Testing

- `cargo test` (655 passed), `cargo check` clean
- `pnpm typecheck`, `pnpm lint` (no new warnings), `pnpm build`
- `vitest run` (2515 passed)


debba:
Hi @Aditeya ,
thanks for your contribution.

Tbh, I'd just like to avoid hardcoding plugin logic into the core, as I mentioned in my code review comments.

This could be solved in several ways, either by using UI slots in plugins or by defining it through capabilities, we can agree on the approach.

Newt has also drafted an RFC around core plugins here:

https://github.com/orgs/TabularisDB/discussions/306

Regarding the RFC, I'd like to clarify the main objective a bit better.

The goal is not necessarily to turn everything into plugins or to aggressively decouple every subsystem. The real objective is to prevent domain-specific knowledge from leaking into the core and to establish clear capability boundaries.

For example, database dialects, transports, exporters, and other feature-specific concerns should ideally expose capabilities through well-defined contracts, while the core remains as thin and generic as possible.

Whether those capabilities are implemented through modules, feature flags, UI slots, or actual plugins is a secondary concern. The important part is defining the seam first and avoiding situations where adding a new feature requires touching multiple unrelated parts of the core.

In the context of plugins, my concern is specifically avoiding hardcoded plugin logic inside the core. If a plugin needs to expose UI, commands, capabilities, or other integrations, I'd prefer those to be defined through extension points and capability contracts rather than special-case logic in the application itself.

I'm happy to discuss the best way to implement this. It's effectively a new architectural scenario for the project, so it makes sense to evaluate it together and reach a shared understanding before moving forward.

@NewtTheWolf, what's your take on this?

Aditeya:
> In the context of plugins, my concern is specifically avoiding hardcoded plugin logic inside the core. If a plugin needs to expose UI, commands, capabilities, or other integrations, I'd prefer those to be defined through extension points and capability contracts rather than special-case logic in the application itself.

I agree with this. In fact when I was looking to fix this I looked at the plugin repo for the UI code but couldn't find anything.

> I'm happy to discuss the best way to implement this. It's effectively a new architectural scenario for the project, so it makes sense to evaluate it together and reach a shared understanding before moving forward.

Its alright with me, I'll wait till Newt has finalized decoupling the plugin system. Then we can take it forward?

NewtTheWolf:
Thanks for the work here @Aditeya, and thanks @debba for the review.

This is a sensible step in the right direction: `supports_ssl` isn't special-case logic — it extends the existing `DriverCapabilities` contract, and gating the SSL tab on a capability instead of the hardcoded `mysql`/`postgres` allow-list keeps the core generic while letting plugins opt in.

One remaining rough edge: the SSL *mode options* are still picked by driver ID (now including a `clickhouse` branch) rather than declared by the driver. Not a blocker — we'll track that as a follow-up to make the SSL modes plugin-defined.

Good to merge. 🚀

NewtTheWolf (approved):
Approved — sensible capability-based approach. Follow-up tracked for making SSL modes plugin-defined.

debba:
Totally agree with @NewtTheWolf 
I will merge it @Aditeya 
Feel free to open other PRs and join on Discord for discussing other stuff.

Aditeya:
@debba don't forget to merge https://github.com/TabularisDB/tabularis-clickhouse-plugin/pull/2 because this pr depends on it

debba:
@Aditeya yes and also we need to create a new release of clickhouse plugin.

PR #311, opened 2026-06-09
feat(mcp): add approval attention controls and localized notifications

### LINKED ISSUE #307: [Feat]: MCP approval attention controls and localized notifications
## Prerequisites

## 🚀 Feature Proposal
Add MCP approval attention controls so Tabularis can actively surface pending approval requests instead of silently waiting in the background. When a new MCP approval arrives, the app should be able to bring the main window to the foreground, temporarily force always-on-top, request OS-level user attention, send a native notification, and play a short alert sound.

This also adds two persisted safety settings under the MCP safety section:
- mcpApprovalAlwaysOnTop (default 	rue)
- mcpApprovalNotifySound (default 	rue)

The approval gate should wait until the application language is settled before triggering modal copy and notification copy, so users see localized text consistently. If language application hangs, the UI should fail open instead of blocking approvals indefinitely.

## Motivation
Today MCP approvals can be easy to miss if Tabularis is behind another window or if startup language hydration is still in progress. That creates a bad AI approval workflow because the MCP client is waiting while the user has no clear signal that action is required.

These changes improve discoverability and safety by making approval prompts visible and configurable, while also protecting against regressions where temporary window state is not restored or localization races show inconsistent copy.

## Example
1. A Claude/Cursor/OpenCode MCP client sends a query that requires approval.
2. Tabularis shows the approval modal.
3. If enabled, Tabularis:
   - brings the main window to the front,
   - temporarily makes it top-most,
   - requests OS-level user attention,
   - sends a native notification,
   - plays a short alert sound.
4. After the approval is handled, the previous window state is restored.
5. Users can disable either always-on-top attention or notification/sound from MCP Safety settings.

Implementation notes covered by the branch:
- Tauri notification plugin integration
- extra Tauri window capabilities for attention/focus/minimize/show/hide behavior
- frontend utility for window snapshot/restore and alert playback
- tests for attention behavior, language-settlement timing, and notification/sound handling

Stiwar0098:
## Summary

Adds attention-grabbing controls for MCP approval gates so users never miss a pending approval, with fully localized notification text.

Closes #307

### Attention Flow (src/utils/mcpApprovalAttention.ts)
- **Window snapshot/restore**: captures and restores window position, size, and visibility
- **Focus request**: brings window to front with 
equestUserAttention(Informational)
- **System notification**: sends OS notification with localized title + body
- **Alert sound**: optional sound on notification

### Settings (McpSafetySection)
- mcpApprovalAlwaysOnTop — keep approval window on top while pending
- mcpApprovalNotifySound — play alert sound on notification
- Labels localized in 8 languages (en, es, de, fr, it, ja, ru, zh)

### Trigger (AiApprovalGate)
- Calls attention flow when a pending approval appears
- Respects both opt-out toggles

### Infrastructure
- Added Tauri capabilities: 
otification:default, is-visible, is-minimized, 
equest-user-attention, is-always-on-top, set-always-on-top
- Rust config structs for the two new settings
- isLanguageApplied() helper avoids spurious changeLanguage() on startup
- Unit + component test coverage for attention flow, opt-out paths, and language-settlement edge cases

## Test plan

- [x] 23 targeted tests pass (pnpm test)
- [x] Production build succeeds (pnpm build && pnpm tauri build)
- [x] Manual: approval gate appears on top when setting enabled
- [ ] Manual: notification fires with localized text in each locale
- [x] Manual: sound plays/disabled per setting
- [x] Manual: window restores to original state after approval


src/contexts/SettingsProvider.tsx
```diff
@@ -129,16 +234,42 @@ export const SettingsProvider = ({ children }: { children: ReactNode }) => {
     };
 
     loadSettings();
-  }, []);
+  }, [isLanguageApplied, queueLanguageApplication]);
```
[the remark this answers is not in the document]
Stiwar0098: Fixed in 941629b — removed both \isLanguageApplied\ and \queueLanguageApplication\ from the mount effect dependency array. The effect now uses \[]\ as deps, which was the original intent: \loadSettings()\ should only run once on mount. The callbacks are still referenced inside the effect body, but since they're consumed synchronously during mount (before any re-render can change them), this is safe.

src/utils/mcpApprovalAttention.ts
```diff
@@ -0,0 +1,269 @@
...
+  );
+}
+
+export async function focusWindowForApproval(approvalId: string): Promise<void> {
+  const appWindow = getCurrentWindow();
+
+  approvalAttentionOwnerId = approvalId;
+  const version = ++approvalAttentionVersion;
+
+  if (windowAttentionSnapshot === undefined) {
```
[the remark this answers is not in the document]
Stiwar0098: Fixed in 941629b — the snapshot capture condition now checks windowAttentionSnapshot === undefined || windowAttentionSnapshotApprovalId !== approvalId. When a new approval ID arrives, a fresh snapshot is captured unconditionally instead of reusing the stale residual from a previous partial restore. Added a test case that verifies this scenario: first approval partially restores (setAlwaysOnTop fails), then a second approval with a different ID correctly captures a fresh snapshot.

NewtTheWolf:
hey @Stiwar0098 thanks for your work!

didnt had time to review it yet but it is still on my todo! will do it tomorrow ASAP

NewtTheWolf (changes requested):
## Review — MCP approval attention controls

Nice work @Stiwar0098 — clean watcher→event→gate architecture and solid config tests. Tested locally on Linux; the core flow works when a language is set. Findings:

**🔴 Blocking — no approval shows without a configured language.** Fresh config (`language: null`, the default on first launch) → modal *and* notification never appear, the MCP call just times out. Backend emit is fine — it's the `isLanguageSettled` gate ([L47](https://github.com/TabularisDB/tabularis/blob/941629b0bc72eb0535241bb5a1caba44c1df36b7/src/components/modals/AiApprovalGate.tsx#L47), [L79](https://github.com/TabularisDB/tabularis/blob/941629b0bc72eb0535241bb5a1caba44c1df36b7/src/components/modals/AiApprovalGate.tsx#L79)). i18next has `fallbackLng: 'en'`, so decouple from the gate and translate with the fallback instead of suppressing.

**🟡 Alert sound — couldn't validate.** On my machine the sound never played (the notification did). I'm fairly sure that's my own OS notification-sound settings, so I can't confirm whether the code is fine or whether it's hitting the Web Audio autoplay policy (programmatic open → no user gesture → `AudioContext` stays `suspended`, [L252-266](https://github.com/TabularisDB/tabularis/blob/941629b0bc72eb0535241bb5a1caba44c1df36b7/src/utils/mcpApprovalAttention.ts#L252-L266)). @debba — could you validate the sound on a setup where notification sounds are known to work?

**🟡 React rule #2** — sync `setState` in `useEffect` ([L245-246](https://github.com/TabularisDB/tabularis/blob/941629b0bc72eb0535241bb5a1caba44c1df36b7/src/contexts/SettingsProvider.tsx#L245-L246), [L253-254](https://github.com/TabularisDB/tabularis/blob/941629b0bc72eb0535241bb5a1caba44c1df36b7/src/contexts/SettingsProvider.tsx#L253-L254)).

ℹ️ Window not raising on Linux is *not* a bug — the WM blocks focus-stealing; the code calls the right APIs.

Left two inline suggestions on the gate decoupling.

src/components/modals/AiApprovalGate.tsx
```diff
@@ -1,13 +1,82 @@
...
+    })();
+  }, [currentApprovalId, isLanguageSettled, isSettingsLoading, restoreWindowState, settings.mcpApprovalAlwaysOnTop, settings.mcpApprovalNotifySound, t]);
+
+  useEffect(() => {
+    return () => {
+      void restoreWindowState();
+    };
+  }, [restoreWindowState]);
+
+  if (!current || !isLanguageSettled) return null;
```
NewtTheWolf: Decouple the modal from `isLanguageSettled` — with i18next `fallbackLng: 'en'` the prompt renders in English until the language settles, which beats never showing it at all on a fresh install (no language set).

```suggestion
  if (!current) return null;
```

src/components/modals/AiApprovalGate.tsx
```diff
@@ -1,13 +1,82 @@
...
+    if (!currentApprovalId) {
+      void restoreWindowState();
+      notifiedApprovalIdRef.current = null;
+      return;
+    }
+
+    if (!isLanguageSettled) {
+      return;
+    }
+
```
NewtTheWolf: Same decoupling for the attention + notification effect — otherwise the OS notification still waits on the language settling. Dropping this guard lets it fire with the i18next fallback.

```suggestion

```

src/components/modals/AiApprovalGate.tsx
Stiwar0098: Fixed in 8601478. The modal is now decoupled from isLanguageSettled, so a pending approval renders immediately and i18next can fall back to English until the preferred language finishes settling.

src/components/modals/AiApprovalGate.tsx
Stiwar0098: Fixed in 8601478 as well. I removed the isLanguageSettled guard from the attention and notification effect, so the OS notification no longer waits for language settlement and uses the i18next fallback strings when needed.

Stiwar0098:
On the sound point, I tested the approval flow on Windows and the notification sound does play correctly on my side. That suggests the implementation is working at least on Windows, and the earlier no-sound result may have been environment-specific on Linux.

Stiwar0098:
On the setState note, this is now addressed in 8601478 as well. The language effect no longer does synchronous flag resets inside useEffect; isLanguageReady and isLanguageSettled are derived from the current i18n state plus the tracked result of the async language application for the active language, which preserves the same behavior without the extra post-render setState pattern.

NewtTheWolf:
Followed up on the alert sound, @Stiwar0098. Confirmed it's **not** a bug in your code — on Linux the synthesized Web Audio tone can't play because WebKitGTK's autoplay policy keeps the programmatically-started `AudioContext` suspended (no user gesture). Your Windows result was correct; WebView2 doesn't have that restriction. Verified end-to-end on KDE: system audio is fine, but the in-page tone never reaches the sink.

To avoid a review ping-pong round I pushed a fix straight to the branch (`ed9c12f`): on Linux the alert now rides the OS notification's XDG theme sound (`message-new-instant`) instead, and the in-page tone is skipped there to avoid a double alert. macOS/Windows keep using the Web Audio tone, so your platform is unchanged. Added a unit test; full suite green (2530).

Could you give it a quick check on Windows to confirm the tone still plays there as before? 🙏

NewtTheWolf (approved):
Re-reviewed `941629b..ed9c12f` — both blocking points resolved, nicely done @Stiwar0098.

✅ **Language gate** — `isLanguageSettled` fully decoupled from the effect, deps, and render gate. Verified end-to-end on a fresh `language: null` config (staged a pending approval): modal **and** notification now appear with English fallback text, where before nothing showed. The inverted test locks it in.

✅ **React rule #2** — `isLanguageReady`/`isLanguageSettled` are derived during render; the i18n effect only sets state post-`await` with a language guard.

🔊 **Sound** — addressed in the follow-up commit I pushed (Linux now uses the OS notification sound; see comment above). Works end-to-end on KDE.

Branch tests green, typecheck + lint clean (the `exhaustive-deps` warning on the `loadSettings` mount effect is the intentional run-once pattern). LGTM 🚀

Stiwar0098:
Sorry for the delay on this. I rechecked the approval flow on Windows, and the tone still plays as before after the Linux-specific follow-up, so Windows behavior looks unchanged on my side.

NewtTheWolf (approved):

Issue #310, opened 2026-06-09
[Feat]: Refresh and Add Table buttons are difficult to select when scroll bars appear


VincentZhangy:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Currently, when scroll bars appear in the table list box, the Refresh and Add Table buttons become difficult to click or tap. This may be caused by the scroll bar overlapping and partially obscuring the button area.

This proposal suggests adjusting the layout or z‑order so that both buttons remain fully accessible regardless of whether scroll bars are visible. Possible solutions include:
* Adding margin or padding to the buttons to prevent overlap with the scroll bar.
* Increasing the clickable area of the buttons (e.g., using a larger hit target) when scroll bars are detected.
* The goal is to ensure a smooth, frustration‑free user experience when managing table data.

### Motivation

Users need to refresh table list or add a new table. When scroll bars appear, the current UI behavior makes it hard to select the Refresh and Add Table buttons without accidentally scrolling or misclicking. This degrades productivity and creates unnecessary friction, especially for users with less precise mouse control or those on touch devices. Improving the button accessibility will reduce user errors, speed up common workflows, and enhance overall usability without adding visual clutter.

debba:
Hi @VincentZhangy thanks for the feedback!
Do you want to create a PR for it? I can assign this issue to you if you want.

VincentZhangy:
@debba ok

DhruvShah-Dev:
Opened draft PR #539 to address this. The change reserves a padded action lane in sidebar accordion headers, keeping the Refresh/Add controls away from the scrollbar edge while long section titles truncate instead of squeezing the buttons. Validation passed: focused accordion tests, typecheck, lint, production build, and GitNexus low-risk change detection.

PR #315, opened 2026-06-10
feat(frontend): Refresh and Add Table buttons are difficult to select when scroll bars appear

VincentZhangy:
issue #310


debba:
lgtm, thanks a lot for this contribution

Issue #316, opened 2026-06-10
External plugin host support gaps found while building a DM / Dameng driver


haos666:
## Context

Hi! We have been building a third-party DM / Dameng database driver as a Tabularis external plugin: https://github.com/haos666/tabularis-dameng-plugin

The plugin has reached a plugin-side `v1.0.0` baseline: https://github.com/haos666/tabularis-dameng-plugin/releases/tag/v1.0.0

It is implemented as a standalone Java process using Tabularis' JSON-RPC-over-stdin/stdout plugin contract. The Dameng JDBC driver is provided locally by the user and is not redistributed by the plugin.

The plugin side currently implements and validates:

- connect / ping / schemas / tables / columns
- indexes / foreign keys / schema snapshot / batch metadata
- views and view definitions
- routines and routine parameters
- triggers and trigger definitions, plus plugin-side create/drop trigger RPCs
- explain plans
- SQL execution for SELECT, DML, and DDL
- row editing CRUD
- DDL preview helpers
- `execute_query_batch`
- BLOB/CLOB handling, including a Tabularis-compatible BLOB wire format

We have validated it with:

- a local Docker DM instance
- direct JSON-RPC protocol smoke tests
- the Tabularis UI
- the Tabularis MCP connection path

While doing this, we found a few host-side gaps where the external plugin contract or host forwarding does not yet expose capabilities that the plugin can already provide. We would like to help with PRs, but wanted to open an issue first to align on scope and sequencing.

### 1. Manifest schema and plugin guide are slightly behind runtime capabilities

The TypeScript runtime model and UI already use `capabilities.triggers`, but `plugins/manifest.schema.json` does not declare it yet.

Also, `plugins/PLUGIN_GUIDE.md` still shows older column metadata field names in at least one example:

- `column_default`
- `is_primary_key`

The current runtime model appears to consume:

- `default_value`
- `is_pk`

The guide also does not currently document the trigger RPCs used by the UI.

### 2. SQL dump/import does not support external plugin drivers

When using the SQL/database import flow with an external plugin connection, Tabularis currently returns `Unsupported driver`. In `src-tauri/src/dump_commands.rs`, import/export paths appear to dispatch only built-in drivers such as MySQL, PostgreSQL, and SQLite.

For the DM plugin, this means a valid schema script can be executed through database-native tooling or plugin SQL execution paths, but not through Tabularis' import UI.

### 3. External plugin batch execution is implemented plugin-side but not fully host-forwarded

[...]

PR #317, opened 2026-06-10
Align external plugin trigger capability and metadata docs

haos666:
## Summary

Follow-up to #316.

This aligns the external plugin manifest schema and plugin guide with the current runtime model:

- Add `capabilities.triggers` to `plugins/manifest.schema.json`
- Add `capabilities.sql_dialect` to the manifest schema and document the supported splitter dialect values
- Document trigger-related external plugin RPCs in `PLUGIN_GUIDE.md`
- Update the `get_columns` example to use the current runtime field names: `default_value` and `is_pk`

## Notes

This PR intentionally does not implement trigger RPC forwarding or splitter behavior changes for external plugins. Runtime forwarding and Oracle-like block splitting will be handled separately so this schema/docs alignment remains low-risk and reviewable.

## Validation

- `python3 -m json.tool plugins/manifest.schema.json`
- `pnpm run typecheck`
- `pnpm run lint` (passes with one existing warning in `src/components/settings/LocalizationTab.tsx`)


debba:
Thanks for the follow-up! Since this documents trigger RPCs and a `sql_dialect` capability that the host doesn't act on yet, I'd prefer to land the runtime forwarding first, so the guide never describes behavior external plugins can't actually use.

Do you plan to open a PR for the trigger RPC forwarding (and the splitter changes)? If that's coming soon, we can simply keep this one open and merge it right after that lands.

Otherwise, I'd suggest extracting the `get_columns` example fix (`default_value` / `is_pk`) into its own PR — that part is already accurate today and can be merged right away — while the trigger/`sql_dialect` schema and docs changes stay here, sequenced after the runtime implementation.

haos666:
Thanks, that sequencing makes sense to me.

I'll keep this PR open for now rather than extracting the `get_columns` example fix separately. My plan is to follow up with runtime PRs first, then come back to this contract/docs PR once the behavior it documents is available in the host:

1. External plugin trigger RPC forwarding (`get_triggers`, `get_trigger_definition`, `create_trigger`, `drop_trigger`)
2. Oracle/DM splitter behavior for `sql_dialect: "oracle"` (`/` terminator, PL/SQL-style blocks, DM `CLASS` / `JAVA CLASS` forms, q-quoting)

That should avoid merging guide/schema text ahead of runtime support, while keeping this PR as the docs/schema alignment companion once the runtime pieces land.

PR #319, opened 2026-06-11
Kubernetes Selection Dialog Improvements

metalgrid:
### Summary

 The Kubernetes connection dialogs had two usability issues:

 1. Non-searchable dropdowns — Context, namespace, and resource name selectors
    required scrolling through potentially long lists with no way to filter.
 2. Hard-coded MySQL port — The K8s "Container Port" always defaulted to 3306
    regardless of the selected database driver (Postgres 5432, ClickHouse 8123, etc.)
    or the service's actual exposed port.

 ### Changes

 Searchable dropdowns
 - Enabled search on K8s context, namespace, saved connection, and resource name
   selectors in both K8sConnectionsModal and the K8s tab of NewConnectionModal. Uses
   the existing Select component's built-in search with common.search /
   common.noResults i18n keys.
 - Resource type (service/pod) kept non-searchable since it only has two options.

 Driver-aware port defaults
 - Replaced hard-coded 3306 with activeDriver.default_port from the driver manifest,
   so:
     - MySQL → 3306, Postgres → 5432, plugins (e.g. ClickHouse) → their manifest
       default
 - K8sConnectionsModal accepts an optional defaultPort prop from the parent modal.
 - Port placeholder and initial value reflect the selected driver.

 Single-service-port auto-prefill
 - New additive Tauri command get_k8s_resource_ports_cmd discovers ports for a
   selected K8s service via kubectl get service -o jsonpath={.spec.ports[*].port}.
 - TypeScript utility getK8sResourcePorts() wraps the command.
 - When a service exposes exactly one port and the user hasn't manually edited the
   port, it is auto-prefilled.
 - Port auto-prefill is restricted to services only (pods skipped).
 - Guarded by isPortOverridden / isK8sPortOverridden flags so manual edits are never
   overwritten.

### DIFF
--- a/src/components/modals/K8sConnectionsModal.tsx
+++ b/src/components/modals/K8sConnectionsModal.tsx
@@ -55,8 +55,9 @@ export function K8sConnectionsModal({
   const [namespace, setNamespace] = useState("");
   const [resourceType, setResourceType] = useState<string>("service");
   const [resourceName, setResourceName] = useState("");
-  const effectiveDefaultPort = defaultPort ?? 3306;
-  const [port, setPort] = useState<number>(effectiveDefaultPort);
+  const effectiveDefaultPort = defaultPort ?? undefined;
+  const [port, setPort] = useState<number | undefined>(undefined);
+  const effectivePort = port ?? effectiveDefaultPort;
   const [isPortOverridden, setIsPortOverridden] = useState(false);
 
   // Discovery state
@@ -145,7 +146,7 @@ export function K8sConnectionsModal({
           resourceName,
         );
         if (!cancelled && ports.length === 1) {
-          setPort(ports[0]);
+          setPort((prev) => (prev === ports[0] ? prev : ports[0]));
         }
       } catch {
         // Best-effort convenience only: keep the current/default port.
@@ -163,7 +164,7 @@ export function K8sConnectionsModal({
     setNamespace("");
     setResourceType("service");
     setResourceName("");
-    setPort(effectiveDefaultPort);
+    setPort(undefined);
     setIsPortOverridden(false);
     setTestStatus("idle");
     setTestMessage("");
@@ -198,22 +199,31 @@ export function K8sConnectionsModal({
   };
 
   const handleSave = async () => {
-    const input: K8sConnectionInput = {
+    const validation = validateK8sConnection({
       name,
       context,
       namespace,
       resource_type: resourceType,
       resource_name: resourceName,
-      port,
-    };
-
-    const validation = validateK8sConnection(input);
+      port: effectivePort,
+    });
     if (!validation.isValid) {
       setValidationError(
         validation.error ?? t("k8sConnections.validationFailed"),
       );
       return;
     }
+
+    if (effectivePort == null) return;
+
+    const input: K8sConnectionInput = {
+      name,
+      context,
+      namespace,
+      resource_type: resourceType,
+      resource_name: resourceName,
+      port: effectivePort,
+    };
 
     try {
       if (isCreating) {
@@ -304,9 +314,9 @@ export function K8sConnectionsModal({
                   setResourceType={setResourceType}
                   resourceName={resourceName}
                   setResourceName={setResourceName}
-                  port={port}
+                  port={effectivePort}
                   setPort={(value) => {
-                    setIsPortOverridden(true);
+                    setIsPortOverridden(value != null);
                     setPort(value);
                   }}
                   defaultPort={effectiveDefaultPort}
@@ -366,9 +376,9 @@ export function K8sConnectionsModal({
                 setResourceType={setResourceType}
                 resourceName={resourceName}
                 setResourceName={setResourceName}
-                port={port}
+                port={effectivePort}
                 setPort={(value) => {
-                  setIsPortOverridden(true);
+                  setIsPortOverridden(value != null);
                   setPort(value);
                 }}
                 defaultPort={effectiveDefaultPort}
@@ -413,9 +423,9 @@ interface EditFormProps {
   setResourceType: (v: string) => void;
   resourceName: string;
   setResourceName: (v: string) => void;
-  port: number;
-  setPort: (v: number) => void;
-  defaultPort: number;
+  port: number | undefined;
+  setPort: (v: number | undefined) => void;
+  defaultPort?: number;
   contexts: string[];
   namespaces: string[];
   resources: string[];
@@ -524,10 +534,12 @@ function EditForm({
         <label className={LabelClass}>{t("k8sConnections.port")}</label>
         <input
           type="number"
-          value={port}
-          onChange={(e) => setPort(Number(e.target.value))}
+          value={port ?? ""}
+          onChange={(e) =>
+            setPort(e.target.value === "" ? undefined : Number(e.target.value))
+          }
           className={InputClass}
-          placeholder={String(defaultPort)}
+          placeholder={defaultPort != null ? String(defaultPort) : undefined}
         />
       </div>
 
--- a/src/components/modals/K8sConnectionsModal.tsx
+++ b/src/components/modals/K8sConnectionsModal.tsx
@@ -213,9 +213,6 @@ export function K8sConnectionsModal({
       );
       return;
     }
-
-    if (effectivePort == null) return;
-
     const input: K8sConnectionInput = {
       name,
       context,
--- a/src/components/modals/K8sConnectionsModal.tsx
+++ b/src/components/modals/K8sConnectionsModal.tsx
@@ -19,6 +19,7 @@ import {
   getK8sContexts,
   getK8sNamespaces,
   getK8sResources,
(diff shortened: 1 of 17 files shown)


src/components/modals/NewConnectionModal.tsx
```diff
@@ -1771,8 +1860,11 @@ export const NewConnectionModal = ({
                   defaultValue: "Container Port",
                 })}
                 value={formData.k8s_port ?? ""}
-                onChange={(v) => updateField("k8s_port", Number(v))}
-                placeholder="3306"
+                onChange={(v) => {
+                  setIsK8sPortOverridden(true);
+                  updateField("k8s_port", Number(v));
+                }}
```
NewtTheWolf, line 1869: `Number("")` is `0`, so clearing this field stores an invalid port `0` **and** flips `isK8sPortOverridden` — which then blocks both the driver-default and the single-port auto-prefill for the rest of the session. Mapping empty input back to `undefined` lets the defaults still kick in:

```suggestion
                onChange={(v) => {
                  setIsK8sPortOverridden(true);
                  updateField("k8s_port", v === "" ? undefined : Number(v));
                }}
```

The same `Number(value)` → `0`-on-empty pattern also exists at [`K8sConnectionsModal.tsx` L528](https://github.com/TabularisDB/tabularis/blob/6c15048d087da38eb424628aad15eb095c870fdc/src/components/modals/K8sConnectionsModal.tsx#L528).

src/components/modals/NewConnectionModal.tsx
```diff
@@ -377,6 +380,76 @@ export const NewConnectionModal = ({
...
+    setFormData((prev) =>
+      prev.k8s_port != null ? prev : { ...prev, k8s_port: k8sDefaultPort },
+    );
+  }, [
+    formData.k8s_enabled,
+    formData.k8s_port,
+    isK8sPortOverridden,
+    k8sDefaultPort,
+    k8sMode,
+  ]);
```
NewtTheWolf: [`.rules/react.md`](https://github.com/TabularisDB/tabularis/blob/main/.rules/react.md) #2 says *never call `setState` synchronously inside `useEffect`* — this effect seeds the default port via a synchronous `setFormData`. It's also redundant with the `k8s_enabled` toggle and the inline/existing-mode toggle, which already set `k8s_port` to `k8sDefaultPort`.

Can this effect be dropped (relying on those handlers), or is it covering a path they miss — e.g. loading an existing connection that has k8s enabled but no port? If it has to stay, a one-line comment on *why* would help. Practical impact is low (the functional update returns `prev` unchanged once the port is set), but it's an explicit repo-rule violation.

src/components/modals/K8sConnectionsModal.tsx
```diff
@@ -52,7 +55,9 @@ export function K8sConnectionsModal({
   const [namespace, setNamespace] = useState("");
   const [resourceType, setResourceType] = useState<string>("service");
   const [resourceName, setResourceName] = useState("");
-  const [port, setPort] = useState<number>(3306);
+  const effectiveDefaultPort = defaultPort ?? 3306;
```
NewtTheWolf: Nit: the PR's goal is to remove the hard-coded `3306`, but it survives here as the fallback when a driver reports [`default_port: null`](https://github.com/TabularisDB/tabularis/blob/6c15048d087da38eb424628aad15eb095c870fdc/src/types/plugins.ts#L62) — a standalone K8s connection for such a driver would silently default to MySQL's port. Not blocking, just flagging the irony; a named constant or deriving from the active driver would read clearer.

NewtTheWolf (changes requested):
@metalgrid — pulled this onto a live minikube cluster to check the new flow end to end: the added `kubectl … -o jsonpath={.spec.ports[*].port}` returns `5432` for a single-port service and `8123 9000` for a multi-port one, and auto-prefill fires only in the single-port case as intended. Branch tests are green (25 TS + 4 Rust parser), and the backend command is solid — kubectl args go through an array (no shell injection) and `u16` parsing drops overflow.

Two things I'd like resolved before merge, plus a few nits.

**Blocking**
1. [`NewConnectionModal.tsx` L383–403](https://github.com/TabularisDB/tabularis/blob/6c15048d087da38eb424628aad15eb095c870fdc/src/components/modals/NewConnectionModal.tsx#L383-L403) violates [`.rules/react.md`](https://github.com/TabularisDB/tabularis/blob/main/.rules/react.md) #2 — synchronous `setFormData` inside `useEffect`.
2. [`NewConnectionModal.tsx` L1863–1866](https://github.com/TabularisDB/tabularis/blob/6c15048d087da38eb424628aad15eb095c870fdc/src/components/modals/NewConnectionModal.tsx#L1863-L1866): clearing the port field stores an invalid `0` and disables auto-prefill for the session.

**Nits**
- Hard-coded `3306` fallback survives in [`K8sConnectionsModal.tsx` L58](https://github.com/TabularisDB/tabularis/blob/6c15048d087da38eb424628aad15eb095c870fdc/src/components/modals/K8sConnectionsModal.tsx#L58) — the very thing this PR removes elsewhere.
- Same `Number(value)` → `0`-on-empty at [`K8sConnectionsModal.tsx` L528](https://github.com/TabularisDB/tabularis/blob/6c15048d087da38eb424628aad15eb095c870fdc/src/components/modals/K8sConnectionsModal.tsx#L528).
- No tests for the trickiest new logic — the auto-prefill effects and the `isK8sPortOverridden` guard. The Rust parser and invoke wrapper are covered, but not the React behavior.

src/components/modals/NewConnectionModal.tsx
metalgrid: Changed override to verify if the user actually passed values. This allows to remove the overridden value even if there was one.

Additionally, modified the input to be numeric, as it is in the General tab (prevents typing a letter and ending up with `NaN` as the value).

src/components/modals/NewConnectionModal.tsx
metalgrid: I'm not very good with React, so I may state something incorrect and possibly wrong.

The way I understand it, is that without the effect we rely on users interactions to progress the setup. It's necessary in the event of a user editing an existing k8s-enabled connection in order to set values in the form data, or to fill fields after we fetch connection parameters.

For now, I've wrapped it into an async function, which (to my understanding) doesn't violate the sync setState in an effect rule, but this causes the values to change depending on latency. So when you manage a cluster across  the world, it takes a moment to get the correct port filled, and before that it shows the driver default. Not sure if that can be made better (perhaps a loading state where fields get disabled while fetching? feels complicated). I'm open to suggestions on this one.

src/components/modals/K8sConnectionsModal.tsx
metalgrid: Dumb oversight on my end. :grin:

metalgrid:
My intent was to have this as a QoL PR because I extensively use it, and while great that it's now a first-class citizen, it did feel rather clunky. Apologies for having y'all point out the mess. :smile:

NewtTheWolf:
hey @metalgrid no wories! thats why we do reviews ;p

Thanks for your work, i was a bit busy the last few days. I will review your changes tomorrow (CET)

src/components/modals/K8sConnectionsModal.tsx
```diff
@@ -158,23 +199,32 @@ export function K8sConnectionsModal({
...
+    });
     if (!validation.isValid) {
       setValidationError(
         validation.error ?? t("k8sConnections.validationFailed"),
       );
       return;
     }
 
+    if (effectivePort == null) return;
+
```
NewtTheWolf: Optional code-quality cleanup: this guard is now unreachable — `validateK8sConnection` just above already rejects a missing/zero port (`"Port must be between 1 and 65535"`), so `effectivePort` can never be `null` by the time we get here. Safe to drop:

```suggestion

```

NewtTheWolf (approved):
Re-reviewed `36d4d1f..65a9074` — all three points addressed, nicely done @metalgrid. Pulled the branch and ran the new suites: 5/5 green.

✅ **`Number("")` → port `0`** — fixed in both spots (empty input maps to `undefined`, and `setIsK8sPortOverridden(v !== "")` re-enables the defaults). Covered by the new "clearing a manual port…" tests.

✅ **React rule #2** — you went further than just async-wrapping it: the synchronous seeding effect is gone entirely and the port is now derived during render (`resolveK8sPort`), with `setK8sAutoPort` running post-`await`. Deps are exhaustive. The default→fetched transition you flagged is fine as-is — showing the driver default until discovery returns is reasonable UX and no longer a rule violation. Good call.

✅ **Hardcoded `3306`** — gone (`defaultPort ?? undefined`), placeholder handles the no-default case.

One optional code-quality suggestion inline (a now-unreachable guard) — non-blocking.

LGTM 🚀

metalgrid:
Fixed a type issue with `number | undefined` for the k8s port. Extended the validation with localized messages instead of the hardcoded English strings.

PR #320, opened 2026-06-12
fix(view-editor): robustly extract the SELECT body from view definitions

maacl:
Replaces #290 (closed as a duplicate).

The View editor needs the bare `SELECT` body so people can edit a view without touching the `CREATE VIEW ... AS` wrapper that each engine emits. Until now we found that body by looking for the first `" AS "` with `indexOf` and cutting there. That works on the happy path but falls apart quickly:

- View definitions are often formatted across multiple lines, so the leading `AS` ends up surrounded by newlines rather than plain spaces and never matches.
- An `AS` can show up well before the real one inside a column alias list, a string literal, a comment, or a quoted/bracketed identifier, and `indexOf` happily cuts on the wrong one.

This swaps that out for a small left-to-right scan (`extractEditableViewDefinition` in `src/utils/sql.ts`) that walks the definition while tracking paren depth and skipping over single quotes, double quotes, backticks, `[bracket]` identifiers, and `--`, `#` and `/* */` comments. It cuts on the first top-level `AS` keyword that actually follows the view header, so the column list, literals and comments no longer throw it off. When there's no such `AS` — e.g. Postgres `pg_get_viewdef` already hands back a bare `SELECT` — the input is returned untouched.

`ViewEditorModal` just calls the helper now instead of carrying the inline string logic.

Covered by tests in `tests/utils/sql.test.ts`: MySQL `SHOW CREATE VIEW` output, multiline SQLite, bare Postgres `SELECT` (with and without aliases), `AS` hidden in strings/comments/quoted identifiers, SQL Server bracket identifiers, `AS` inside a column list, `WITH CHECK OPTION` / `WITH READ ONLY`, and a malformed-input fallback.


debba:
Hi @maacl !
This PR replaces #290 right?

maacl:
It does indeed, sorry.

debba:
Took this for a spin locally and it's a solid cleanup. The old `indexOf(" AS ")` approach was fragile, it'd happily cut on an `AS` inside a column alias, a string literal or a comment. The new tokenizer handles all of that properly (quotes, backticks, bracket identifiers, comments, paren depth) and the Postgres bare-`SELECT` path is left untouched, which is the right call.

Ran `tests/utils/sql.test.ts` here and all green (40/40). Coverage looks good too, the column-list and "AS inside strings" cases were exactly the ones the old code got wrong.

One small thing: the PR description is empty : mind adding a line or two on the motivation for posterity?

@newtthewolf could you take a look as well? Would like a second pair of eyes on the dialect edge cases before merging.

maacl:
Does not appear that I can update the PR description? Do I convert to a draft, or?

debba:
No worries @maacl, contributors can't edit the PR body once it's opened against the upstream repo, so I've filled it in for you, mostly so there's a paper trail of what changed and why for whoever looks at this later. No need to convert to a draft. Thanks again for the cleanup!

tests/utils/sql.test.ts
```diff
@@ -196,4 +196,63 @@ describe('sql utils', () => {
...
+      {
+        name: 'preserves WITH READ ONLY',
+        sql: 'CREATE VIEW reporting_users AS SELECT * FROM users WITH READ ONLY',
+        expected: 'SELECT * FROM users WITH READ ONLY',
+      },
+      {
+        name: 'falls back to trimmed input when no top-level AS follows VIEW',
+        sql: 'CREATE VIEW broken_view SELECT 1',
+        expected: 'CREATE VIEW broken_view SELECT 1',
+      },
```
NewtTheWolf, line 253: Optional: add cases for the common Postgres `MATERIALIZED` / `OR REPLACE` forms — I verified the scanner already handles both, this just guards against regressions:

```suggestion
      },
      {
        name: 'extracts the SELECT from a Postgres MATERIALIZED VIEW',
        sql: 'CREATE MATERIALIZED VIEW mv AS SELECT 1 AS x FROM t',
        expected: 'SELECT 1 AS x FROM t',
      },
      {
        name: 'extracts the SELECT from CREATE OR REPLACE VIEW',
        sql: 'CREATE OR REPLACE VIEW v AS SELECT a AS b FROM t',
        expected: 'SELECT a AS b FROM t',
      },
```

NewtTheWolf (approved):
@debba second pair of eyes as requested 👇 — solid cleanup, thanks @maacl.

Pulled the branch and went after the dialect edge cases specifically. Beyond the PR's own suite (40/40 green) I stress-tested the tokenizer against:
- `CREATE MATERIALIZED VIEW` / `CREATE OR REPLACE VIEW` / `TEMP VIEW`
- SQL Server `[bracket]` identifiers, MySQL backtick + `DEFINER` headers, schema-qualified names
- `AS` hidden in column-list parens, string literals, comments, quoted identifiers
- multiline SQLite, lowercase DDL, parenthesised bodies, leading comments
- Postgres bare `pg_get_viewdef` passthrough

All extract correctly. Also confirmed the old `indexOf(" AS ")` logic isn't duplicated anywhere else and the helper is cleanly isolated to `ViewEditorModal`.

The left-to-right scanner with paren-depth + quote/comment skipping is the right approach here. LGTM 🚀

debba:
That's great, I will merge it.

PR #321, opened 2026-06-12
Forward external plugin trigger RPCs

haos666:
## Summary

Follow-up to #316 and the runtime companion change for #317.

- Forward external plugin trigger RPCs through RpcDriver: get_triggers, get_trigger_definition, create_trigger, and drop_trigger.
- Add RpcDriver unit tests covering method names, params, and result shapes.
- Keep built-in driver trigger behavior unchanged; this only fills the external plugin adapter methods that currently fall back to the default unsupported implementation.

## Validation

- pnpm run typecheck
- pnpm run lint
- cargo test plugins::driver::tests -- --nocapture
- cargo test --lib
- Local DM plugin smoke: listed triggers, viewed definitions, created a trigger, and dropped a trigger through the Tabularis UI.

### DIFF
--- a/src-tauri/src/plugins/driver.rs
+++ b/src-tauri/src/plugins/driver.rs
@@ -13,7 +13,7 @@ use tokio::sync::{mpsc, oneshot};
 use crate::drivers::driver_trait::{DatabaseDriver, PluginManifest};
 use crate::models::{
     ColumnDefinition, ConnectionParams, DataTypeInfo, ExplainPlan, ForeignKey, Index, QueryResult,
-    RoutineInfo, RoutineParameter, TableColumn, TableInfo, TableSchema, ViewInfo,
+    RoutineInfo, RoutineParameter, TableColumn, TableInfo, TableSchema, TriggerInfo, ViewInfo,
 };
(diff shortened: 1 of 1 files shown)


debba:
LGTM.
@NewtTheWolf if you want to do a check for new registry compliance.

NewtTheWolf:
> LGTM. @NewtTheWolf if you want to do a check for new registry compliance.

will do it tomorrow!

debba:
@saurabh500 following #268,  this could be interesting for SQL Server integration too.

NewtTheWolf:
@debba — registry-compliance check done 👍

This is fine from the registry side: the new `tabularium-extensions.schema.json` (in #299) already defines the `triggers` capability (`capabilities.triggers` — *"Enable trigger listing and management"*), so a plugin exposing these forwarded trigger RPCs can declare support in a schema-compliant way. **Nothing to change in #321** for compliance — it's good to go.

One cross-PR note: the only registry-side piece here is that schema, which lives in #299. So as long as it's just the schema change, #321 is fine — but the hosted registry needs to be brought in line with it **before #299 is merged**.

NewtTheWolf (approved):
LGTM

PR #324, opened 2026-06-15
fix(editor): update/delete use full composite PK in WHERE clause

thomaswasle:
Summary

  - Editing or deleting a row in a table with a composite primary key (e.g. PRIMARY KEY
  (profile_id, phone_type, key)) previously affected all rows sharing the partial key,
  because only the first PK column was sent to the backend
  - pkColumn: string | null replaced by pkColumns: string[] | null throughout the
  frontend; the row identity is now a pkMap: Record<string, unknown> object containing all
  PK column values
  - All Tauri commands (update_record, delete_record, save_blob_to_file,
  fetch_blob_as_data_url) and all three drivers (MySQL, SQLite, PostgreSQL) now receive
  pk_map: HashMap<String, Value> and build a compound WHERE col1 = ? AND col2 = ? AND col3
  = ? clause

  Test plan

  - [ ] Open a table with a composite PK (e.g. two columns that together form the PK)
  - [ ] Edit a cell value in one row — confirm only that row is updated
  - [ ] Delete a row — confirm only that row is deleted
  - [ ] Verify single-PK tables still work correctly for edit and delete
  - [ ] Verify BLOB fetch/save works on a composite-PK table


debba (commented):
Thanks for taking this on,
it's a genuinely nasty bug and I'm glad someone finally chased it down. Sending only the first PK column meant composite-key tables were quietly updating or deleting every row that shared that partial value, which is about as bad as it gets for a data tool. The `pk_map` approach is the right one, and I appreciate that you carried it all the way through: all four Tauri commands, the trait, the three drivers, the RPC plugin path, and the whole frontend. I went looking for leftover `pkCol`/`pkVal` references and couldn't find any, so the migration is complete.

Two things in particular made me happy. Bailing out of edit mode entirely when a PK column is missing from the result set — instead of building a partial WHERE — is exactly the kind of defensive choice I'd want here. And slipping `escape_identifier` onto the identifiers that used to be interpolated raw is a nice bonus on top of the actual fix.

A few things I'd like sorted before we merge:

- **Tests.** Right now everything in the test diff is just the `pkColumn` → `pkColumns` rename — the new helpers have no coverage. `serializePkKey` and `buildPkMap` on the TS side, and `build_mysql_pk_where` / `build_pk_map_predicate` / `sqlite_push_pk_where` on the Rust side. I'd especially want `serializePkKey` tested for key ordering and the single vs. composite cases. We ask for tests alongside new utilities, so this is the main blocker for me.

- The doc comment on `serializePkKey` says single-column PKs fall back to plain `String(value)` for backward compat, but the code always `JSON.stringify`s — so `id=5` ends up as `{"id":5}`, not `"5"`. It's harmless since these keys never leave memory, but the comment will trip up whoever reads it next, so I'd just fix the comment.

- `build_mysql_pk_where` builds the predicate string and then all three callers throw it away (`let (_, pairs)`) and rebuild it by hand with `push_bind`. The `join(" AND ")` is dead work — I'd have it return just the sorted pairs. Postgres and SQLite are fine; it's only the MySQL helper that's out of step.

- Minor, take it or leave it: `RowEditorSidebar` rebuilds `pkMap` with `Object.fromEntries` on every render, so `BlobInput`'s fetch effect re-fires each time now (it used to lean on stable primitives). A `useMemo` would settle it.

The composite WHERE logic itself reads correctly to me, so really it's the tests and that comment standing between this and a merge. Nice work overall.

thomaswasle:
I tried to address your requested changes. But I am not shure why the kilo code review is failing.Can you have a look and help @debba ?

debba:
Hi @thomaswasle, it looks that kilo failed because of API error, so nothing related to the code .
I tried relaunch it.
Btw I will review ASAP the code.
Thanks a lot

NewtTheWolf (dismissed):
Solid fix — thank you! 🙏 Right at the root: the frontend now derives **all** PK columns (`cols.filter(c => c.is_pk)`), and the backend builds a compound `col1 = $1 AND col2 = $2 …` predicate with properly escaped identifiers and parameter-bound values (no injection). Verified locally against a composite-PK Postgres table: a partial `WHERE a = 1` hits 2 rows (the old bug), the compound `WHERE a = 1 AND b = 'x'` hits exactly 1 (the fix). Tests green on the branch: 324 driver tests + 250 frontend.

One thing before merge: the branch currently **conflicts with `main`** in `src/components/ui/DataGrid.tsx` and `src/components/ui/DataGridRow.tsx` (main touched them in parallel). Please rebase/merge `main` and resolve those two, then it's good to go.

thomaswasle:
I rebased main and fixed conflicts. Should be good to go when Kilo Code is running again. ;-)

PR #325, opened 2026-06-16
Support Oracle-style SQL block splitting

haos666:
## Summary

Follow-up to #316 for the Oracle/DM splitter work discussed with @ymadd.

This keeps the behavior additive and dialect-gated behind the existing `oracle` SQL dialect:

- enable line-leading `/` as an Oracle-style statement terminator
- fold PL/SQL-style source units until the next slash-terminated segment instead of splitting on internal semicolons
- support Oracle q-quoting / nq-quoting so `;` and `/` inside quoted literals do not split statements
- include DM-specific block openers: `CREATE CLASS`, `CREATE CLASS BODY`, and `CREATE JAVA CLASS`
- keep other dialect presets structurally unchanged by leaving the new flags disabled outside `oracle`

This PR does not include trigger RPC forwarding (#321), batch forwarding, SQL import/export, or BLOB hooks.

## Validation

Automated checks:

- `pnpm exec vitest run tests/utils/sqlSplitter`
- `pnpm exec vitest run tests/utils/sqlSplitter/dialects.test.ts tests/utils/sqlSplitter/splitter.test.ts tests/utils/sqlSplitter/tokenizer.test.ts`
- `pnpm exec vitest run` (130 files / 2588 tests)
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test:rust` (671 Rust tests)
- `git diff --check`

Local DM/UI smoke on the DM external plugin:

- `CREATE OR REPLACE TRIGGER ... BEGIN ... END; /` plus a following `SELECT` is detected as 2 statements, not split at internal semicolons.
- q-quoted text containing both `;` and a line-leading `/` stays inside one statement.
- simple slash-separated SQL statements split as expected under the `oracle` dialect.


ymadd:
Thanks @haos666  for the follow-up — happy to take a look!

src/utils/sqlSplitter/splitter.ts
```diff
@@ -92,10 +104,181 @@ function collectSegments(sql: string, options: DialectOptions): RawSegment[] {
...
+  switch (kind) {
+    case 'FUNCTION':
+    case 'PROCEDURE':
+    case 'TRIGGER':
+    case 'LIBRARY':
+      return true;
+    case 'PACKAGE':
+    case 'TYPE':
+    case 'CLASS':
+      return next === 'BODY' || next !== undefined;
```
ymadd: **[should-fix] `CREATE TYPE` (spec) and `CREATE LIBRARY` are mis-classified as PL/SQL block openers, so `foldBlocks` swallows the following statement when no `/` is present.**

`return next === 'BODY' || next !== undefined` reduces to `next !== undefined` (the first clause is dead), so any `CREATE TYPE <name>` is an opener; `LIBRARY` returns `true` unconditionally at L201. Object-type *specs* and `CREATE LIBRARY` carry **no internal `;`** — the `;` is their terminator — so folding them has no upside and merges the next statement. Tabularis sends statements to a driver one-by-one (no `/`, no trailing `;`), so a merged `CREATE TYPE ...; SELECT ...` fails (ORA-00911) and the SELECT is lost.

The right criterion is "does it contain internal `;`": PACKAGE/CLASS specs do, TYPE spec / LIBRARY don't.

```ts
case 'FUNCTION': case 'PROCEDURE': case 'TRIGGER':
  return true;
case 'LIBRARY':
  return false;                 // one-line DDL, terminated by ;
case 'PACKAGE': case 'CLASS':
  return next !== undefined;    // specs carry internal ;
case 'TYPE':
  return next === 'BODY';       // only TYPE BODY carries internal ;
```

A blanket `next === 'BODY'` for all three breaks the existing DM `CREATE CLASS <name>` spec test, so keep PACKAGE/CLASS as `next !== undefined`. Please add regressions: `CREATE TYPE ...; SELECT ...` -> 2 statements, `CREATE LIBRARY ...; SELECT ...` -> 2.

src/utils/sqlSplitter/splitter.ts
```diff
@@ -92,10 +104,181 @@ function collectSegments(sql: string, options: DialectOptions): RawSegment[] {
...
+  if (tokens[index] === 'OR' && tokens[index + 1] === 'REPLACE') {
+    index += 2;
+  }
+  if (
+    tokens[index] === 'EDITIONABLE' ||
+    tokens[index] === 'NONEDITIONABLE'
+  ) {
+    index++;
+  }
+  if (tokens[index] === 'AND' && isJavaCompileOption(tokens[index + 1])) {
```
ymadd, line 195: Oracle grammar is `CREATE [OR REPLACE] [AND {RESOLVE|COMPILE}] [NOFORCE] JAVA {SOURCE|CLASS|RESOURCE} ...`. Only `AND RESOLVE|COMPILE` is skipped here; the optional `NOFORCE` that follows is not, so `CREATE OR REPLACE AND RESOLVE NOFORCE JAVA SOURCE ...` falls through to `default` and the Java body splits on its internal `;`. Add `if (tokens[index] === 'NOFORCE') index++;` after the AND-skip. Low frequency, but a real grammar gap.

src/utils/sqlSplitter/splitter.ts
```diff
@@ -92,10 +104,181 @@ function collectSegments(sql: string, options: DialectOptions): RawSegment[] {
...
+      terminator: endSegment.terminator,
+    });
+    index = endIndex + 1;
+  }
+
+  return output;
+}
+
+function isPlsqlBlockOpener(sql: string, segment: RawSegment): boolean {
+  const tokens = leadingSignificantTokens(sql.slice(segment.start, segment.end));
```
ymadd: `sql.slice(segment.start, segment.end)` copies the whole segment (a package body can be tens of KB) even though `leadingSignificantTokens` only needs the opener prefix. This is avoidable GC pressure on the parse path.

I would not fix this by capping the slice to a fixed byte/char count, because a segment can start with a long license/comment block and still have a valid opener later. A safer shape is to let `leadingSignificantTokens` scan the original `sql` with `start`/`end` bounds, stopping once it has collected enough tokens.

Same function, minor: the `12` cap deserves a named const/comment, and `source.slice(position)` in the word loop allocates per token. A sticky `/[A-Za-z_][A-Za-z0-9_$#]*/y` (matching the existing `DOLLAR_TAG_RE`/`GO_RE`/`SLASH_RE` convention) avoids that allocation.

src/utils/sqlSplitter/splitter.ts
```diff
@@ -92,10 +104,181 @@ function collectSegments(sql: string, options: DialectOptions): RawSegment[] {
...
+    }
+
+    if (endIndex >= segments.length) {
+      endIndex = segments.length - 1;
+    }
+
+    const endSegment = segments[endIndex];
+    output.push({
+      start: segment.start,
+      end: endSegment.end,
```
ymadd: When no `/` is found, the merged span ends at `endSegment.end`, which is *before* the segment terminator. For a block ending exactly at EOF, `END;` becomes `END` in both `Statement.text` and `Statement.range`.

The current backend sanitization (`sanitize_user_query` -> `trim_end_matches(';')`, `src-tauri/src/commands.rs:89`) masks the driver impact for now because it strips that same trailing semicolon before execution anyway. The splitter should still preserve the original statement text/range; otherwise this becomes an execution bug as soon as semicolon stripping is made statement-aware.

```ts
let eofFold = false;
if (endIndex >= segments.length) {
  endIndex = segments.length - 1;
  eofFold = true;
}
// ...
end: eofFold ? sql.length : endSegment.end,

File: src/utils/sqlSplitter/tokenizer.ts [hunk 5]
```diff
@@ -200,6 +212,68 @@ function scanDollarQuoted(source: string, position: number): Token | null {
...
+  } else {
+    return null;
+  }
+
+  const delimiterPosition = position + prefixLength;
+  if (delimiterPosition >= source.length) return null;
+  const openCodePoint = source.codePointAt(delimiterPosition);
+  if (openCodePoint === undefined) return null;
+
+  const openDelimiter = String.fromCodePoint(openCodePoint);
```
ymadd, line 240: Per Oracle, a q-quote delimiter cannot be a space, tab, or newline; these are accepted here, so `q' ...;... '` would be mis-scanned as a quoted literal. Low impact (only over-shields malformed input), but a guard rejecting whitespace delimiters would match the spec — note a single quote *is* a valid delimiter, so don't reject `'`. The UTF-16/codePoint handling and the identifier-boundary guard here look correct.

src/utils/sqlSplitter/index.ts
```diff
@@ -68,6 +69,9 @@ export interface DialectOptions {
   readonly dollarQuoting: boolean;
   readonly customDelimiter: boolean;
   readonly goDelimiter: boolean;
+  readonly slashTerminator: boolean;
```
ymadd: Nit: the option flag is `slashTerminator` but the token kind / `SegmentTerminator` value is `slashDelimiter`. The pre-existing `goDelimiter` uses one word for both the flag and the token. Aligning to `slashDelimiter` (flag + the tokenizer.ts guard) keeps the grep trail consistent.

tests/utils/sqlSplitter/splitter.test.ts
```diff
@@ -194,6 +194,225 @@ describe('splitStatements', () => {
...
+      ).toHaveLength(1);
+    });
+
+    it('keeps a PL/SQL block together at EOF without a trailing slash', () => {
+      const sql = ['CREATE OR REPLACE PROCEDURE p_demo AS', 'BEGIN', '  NULL;', 'END;'].join(
+        '\n',
+      );
+      const result = splitQueries(sql, 'oracle');
+      expect(result).toHaveLength(1);
+      expect(result[0]).toContain('NULL;');
```
ymadd, line 468: This asserts `NULL;` but not the trailing `END;`, so the EOF semicolon loss (see the foldBlocks comment) passes unnoticed. Add `expect(result[0]).toContain('END;')`. Also worth adding: `CREATE TYPE ...; SELECT ...` and `CREATE LIBRARY ...; SELECT ...` -> length 2 (opener fix), `CREATE TYPE BODY ... /` -> length 1, `WITH PROCEDURE ...`, and a CRLF `/` separator case.

ymadd (changes requested):

ymadd:
Thanks @haos666 — clean, well-tested change, and the dialect gating looks solid: the new `slashTerminator`/`plsqlBlocks`/`qQuoting` flags are all `false` for the other presets, so this should keep non-oracle splitting behavior isolated from the new Oracle path. The q-quote scanner (UTF-16/codePoint delimiter, identifier-boundary guard, `N'...'` not mis-parsed as q-quoting, `nq'...'`), the `WITH FUNCTION`-vs-CTE disambiguation, and label handling all check out.

I went through this carefully and checked the Oracle semantics against the Oracle SQL reference and the node-oracledb docs. The one correctness item I'd like resolved before merge is inline: `isCreateBlockOpener` treats `CREATE TYPE` spec / `CREATE LIBRARY` as block openers, so they swallow the next statement when no `/` follows. Heads-up that the fix revises our #316 opener list — `CREATE TYPE`/`LIBRARY` were included there on SQL*Plus `/` semantics, but Tabularis sends statements to a driver one-by-one, where `/` is not part of the statement and SQL statement delimiter semicolons are not sent. For this splitter path, "does the construct contain internal `;`?" is the safer test. Happy to discuss.

**Out of scope of this PR's files, but they block PL/SQL execution end-to-end through the UI (NOT regressions from this PR — flagging to track separately, probably as their own issues):**

1. `sanitize_user_query` (`src-tauri/src/commands.rs:89`) does `trim_end_matches(';')` on every statement before it reaches the driver (`execute_query` / `execute_query_batch`). That strips the structural `;` in `END;`, so `BEGIN ... END;` arrives as `BEGIN ... END` and can hit PLS-00103. This also masks the EOF `;` preservation issue today, and means even a correctly-split trigger/procedure won't execute from the editor until semicolon stripping is made statement/dialect-aware.
2. `runQuery` / `runMultipleQueries` run `extractQueryParams` on each statement (`src/pages/Editor.tsx`; regex in `src/utils/queryParameters.ts:21`). It matches `:OLD` / `:NEW`, so a trigger body (exactly the fixture added here) is read as having bind parameters — either a missing-params modal appears, or the pseudorecords can be interpolated and corrupt the DDL.

The MCP path (`src-tauri/src/mcp/mod.rs:1090`) calls the driver directly without `sanitize_user_query`, so it behaves differently from the editor path. I couldn't test against a live Oracle/DM instance, so these two out-of-scope notes are from reading the code plus the Oracle/node-oracledb docs — would be good to confirm on DM8.

haos666:
Thanks for the detailed review. I pushed an update addressing the requested splitter fixes:

- `CREATE TYPE` specs and `CREATE LIBRARY` now split as normal SQL statements; only `CREATE TYPE BODY` is folded as a block.
- `CREATE OR REPLACE AND RESOLVE NOFORCE JAVA SOURCE` is recognized as a Java block opener.
- `isPlsqlBlockOpener` now scans bounded source text directly instead of slicing per segment, with a named leading-token cap.
- EOF PL/SQL blocks preserve the final `END;`.
- q-quote whitespace delimiters are rejected, while single-quote delimiters remain supported.
- Added regression coverage for TYPE spec, LIBRARY, TYPE BODY, NOFORCE Java, CRLF slash delimiters, WITH PROCEDURE, EOF `END;`, and q-quote delimiter behavior.

Validation run locally:

- `pnpm exec vitest run`
- `pnpm run typecheck`
- `pnpm run lint`
- `pnpm run build`
- `pnpm run test:rust`

I also verified the updated dev app UI against a local DM connection: the query chooser now splits the reviewed cases as expected, including TYPE/LIBRARY as two statements and trigger/Java blocks plus following SELECT as two statements.

ymadd (approved):
LGTM. This addresses all my earlier review points — rename to `slashDelimiter`, `leadingSignificantTokens` substring removal + sticky `WORD_RE`, EOF fold preserving the trailing `;`, `TYPE`/`PACKAGE`/`CLASS` split, `NOFORCE` + `JAVA RESOURCE`, q-quote whitespace-delimiter guard, and CRLF separator coverage. @debba good to merge from my side.

One non-blocking edge case I'll note for the record (no action needed here): a type spec whose name tokenizes to `BODY` — `CREATE TYPE "BODY" AS OBJECT (...)`, or unquoted `CREATE TYPE body ...` since `BODY` isn't a SQL reserved word — is mis-folded as a `TYPE BODY` opener, so a following statement gets merged into it. The trigger is pathological, so it's not worth holding this PR. I'll fold the fix (a lookahead so a `BODY`-named spec doesn't satisfy the body pattern, plus a regression test) into a later splitter change.

debba:
That's great.
I will merge it, and I will expect another PR from you, thanks a lot @ymadd

PR #337, opened 2026-06-19
feat(mysql): support cleartext auth plugin for Warpgate bastions (#336)

### LINKED ISSUE #336: [Feat]: MySQL: expose option to enable mysql_clear_password (cleartext) auth plugin
### Prerequisites

### 🚀 Feature Proposal

Summary
Allow enabling the MySQL cleartext authentication plugin per connection. The underlying mysql_async driver supports this via enable_cleartext_plugin, but it isn't exposed in the connection settings.
Problem
Connecting to a MySQL account whose server-side auth requires a cleartext password (PAM, LDAP, or cloud IAM auth) fails with:

encountered unexpected or invalid data: mysql_cleartext_plugin disabled

The driver disables the cleartext plugin by default, and there's currently no UI option to turn it on, so these accounts can't be used at all.
Proposed solution
Add a per-connection toggle in the MySQL connection settings, e.g. "Enable cleartext password plugin", mapping to the driver's enable_cleartext_plugin option. It should only be allowed together with an SSL/TLS mode, since cleartext must travel over an encrypted connection.

### Motivation

Managed and enterprise MySQL deployments increasingly rely on authentication methods that require the client to send the password in cleartext over a TLS connection — notably PAM, LDAP, and cloud IAM auth (Azure Database for MySQL, AWS RDS IAM, etc.). These are mainstream, security-conscious setups, not edge cases.

A very common real-world trigger is connecting through Warpgate, a popular open-source bastion/proxy. Its MySQL protocol server only accepts TLS connections and authenticates clients exclusively via the mysql_clear_password plugin — this is mandatory, not configurable. Clients that can't enable cleartext auth (like Tabularis today) simply cannot connect to any database behind Warpgate, while clients that expose the option (DBeaver, mysql CLI with --enable-cleartext-plugin) work fine. Supporting this would let Tabularis work in bastion-protected environments, which are standard in security-conscious infrastructures.

### Example

_No response_

pokertour:
Add an opt-in "cleartext password plugin" toggle for MySQL/MariaDB so connections can authenticate through bastions like Warpgate, which require the mysql_clear_password auth plugin. The option is gated on a TLS mode being enabled — cleartext credentials are refused over an unencrypted link.

Bastions like Warpgate proxy MySQL but do not implement the prepared-statement protocol (COM_STMT_PREPARE), so any sqlx::query() — which always prepares — failed with server error 1047 ("Not implemented"). When the cleartext plugin is enabled, the whole MySQL driver now routes statements through the text protocol (COM_QUERY via sqlx::raw_sql): introspection, SHOW CREATE, query execution/streaming, view/trigger DDL, inserts/updates/ deletes (values inlined as escaped literals), EXPLAIN, and export.

Also fix a pool cache-key collision: the ad-hoc key (used for unsaved connections, e.g. the New Connection modal) omitted the username. Warpgate multiplexes many targets behind one host:port and selects the backend by username, so two different targets shared a pool and served each other's databases. The key now includes the username and the cleartext flag.

Tests: pure literal/escaping helpers and pool-key regressions.

Closes #336


src-tauri/src/pool_manager.rs
```diff
@@ -169,6 +177,20 @@ pub(crate) fn build_mysql_options(
         options = options.ssl_client_key(key);
     }
 
+    // Optionally enable the mysql_clear_password (cleartext) auth plugin, used by
+    // bastions like Warpgate. Cleartext credentials must never be sent over an
+    // unencrypted link, so refuse to enable it when TLS is disabled.
+    if params.enable_cleartext_plugin.unwrap_or(false) {
+        if matches!(ssl_mode, MySqlSslMode::Disabled) {
```
NewtTheWolf: **blocker** — `Preferred` only *tries* TLS and falls back to plaintext, so cleartext creds can still cross an unencrypted link — the exact thing this guards against. Gate on `Required`/`VerifyCa`/`VerifyIdentity` only (and drop "Preferred" from the error text).

src-tauri/src/drivers/mysql/helpers.rs
```diff
@@ -5,6 +5,67 @@ pub(super) fn escape_identifier(name: &str) -> String {
...
 }
 
+/// Renders a `&str` as a quoted MySQL string literal for the text protocol.
+///
+/// Used when a query has to bypass the prepared-statement protocol (e.g.
+/// behind a Warpgate-style bastion that rejects `COM_STMT_PREPARE`): the
+/// value can no longer travel as a bind parameter, so it is inlined as an
+/// escaped literal instead. Mirrors `mysql_real_escape_string` for the
+/// default `sql_mode` (backslash escapes enabled).
+pub(super) fn mysql_string_literal(s: &str) -> String {
```
NewtTheWolf: **blocker** — hardcoded backslash-escaping is unsound under `NO_BACKSLASH_ESCAPES` (ANSI mode / some bastion targets): `'o\'brien'` parses as `o'brien` by default but breaks there. User cell values flow through here on the text-protocol UPDATE/DELETE path, so this is an injection vector. Check the server's `@@sql_mode` and double the quote (`''`) when backslash escaping is off.

src-tauri/src/drivers/mysql/helpers.rs
```diff
@@ -5,6 +5,67 @@ pub(super) fn escape_identifier(name: &str) -> String {
...
+}
+
+/// Substitutes each `?` placeholder in `sql` with the next quoted string
+/// literal from `binds`, in order. Used to turn a parameterised
+/// introspection query into a text-protocol statement. Placeholders past
+/// the end of `binds` (and `?` chars when `binds` is empty) are left as-is.
+///
+/// Note: this is only safe for the driver's own queries, whose `?` chars are
+/// exclusively bind placeholders (never literal question marks in strings).
+pub(super) fn inline_str_placeholders(sql: &str, binds: &[&str]) -> String {
```
NewtTheWolf: **nit** — fine for the driver's own queries, but it assumes every `?` is a placeholder; it would break if ever reused for user SQL containing a `?` inside a string literal. Worth an assert or a name that signals "internal queries only".

NewtTheWolf (changes requested):
Thanks — nice work, and good catch on the pool-key/username multiplexing bug. 🙏 Text-protocol routing is clean and the pool fix is verified. Two security points before merge, plus one nit. Tests green on the branch (mysql 26, pool 27).

src-tauri/src/pool_manager.rs
pokertour: fixed in 01a67ec. build_mysql_options now only enables the cleartext plugin when the SSL mode is enforced (Required / VerifyCa / VerifyIdentity); Preferred is rejected alongside Disabled since it can silently fall back to plaintext. Dropped "Preferred" from the error text too. The UI is aligned: the checkbox in NewConnectionModal is now gated on the same enforced modes (and the hint string updated across all 8 locales). Added cleartext_plugin_rejected_with_preferred_tls + cleartext_plugin_allowed_with_enforced_tls.

src-tauri/src/drivers/mysql/helpers.rs
pokertour: Fixed in 01a67ec. mysql_string_literal is now sql_mode-aware: under NO_BACKSLASH_ESCAPES the backslash is treated as literal and the single quote is escaped by doubling ('') instead of \', so o'brien / trailing backslashes no longer break out of the literal. The server setting is read once per operation via SELECT @@sql_mode and threaded through a small TextProto { enabled, no_backslash_escapes } that replaces the old text: bool across the driver (pk_literal and inline_str_placeholders included). The verbatim query path (exec_on_mysql_conn) uses TextProto::protocol_only, so it skips the extra roundtrip since it never inlines literals. Added mysql_string_literal_no_backslash_escapes_mode and an injection-payload case under both  modes.

src-tauri/src/drivers/mysql/helpers.rs
pokertour: Done in 01a67ec — added a # Safety section to inline_str_placeholders spelling out that it treats every ? as a bind placeholder and is therefore only sound for the driver's own hand-written introspection queries, never arbitrary user SQL.

debba:
Hi @pokertour if you can solve conflicts, will merge it in upstream.

PR #342, opened 2026-06-20
feat(views): add PostgreSQL materialized views support

Davydhh:
Part of #16

Adds PostgreSQL **materialized views** to the schema tree.

## What you can do
- Browse materialized views in a dedicated **Materialized Views** group per schema
- Expand to see **columns** and **indexes**
- **Show definition** (read-only `CREATE MATERIALIZED VIEW …`)
- **Refresh** via `REFRESH MATERIALIZED VIEW`
- Open data / count rows like any other relation


debba:
Nice work, the default trait impls keep the other drivers clean. A few things I'd want sorted before merging:

**Translations** — only `en.json` got the new keys. `sidebar.materializedViews`, `sidebar.refreshMaterializedView`, `sidebar.showDefinition`, `views.refreshSuccess/refreshError/failGetDefinition` are missing in it/es/de/fr/ja/ru/zh, so those locales will show English. We keep all 8 in sync.

**The `materialized_views` capability flag isn't actually used.** It's added to both the Rust `DriverCapabilities` and `types/plugins.ts`, and the doc comment says it gates the tree group in the UI, but nothing reads it. The group renders off `materializedViews.length > 0` in `SidebarSchemaItem`, and `DatabaseProvider` calls `get_materialized_views` for every driver in all three load paths. So MySQL/SQLite/plugins fire an extra invoke on every schema load just to get back `[]`. Either gate the invoke on the capability (which is the whole point of the flag) or drop the flag and fix the comment.

**Refresh has no UI feedback.** `REFRESH MATERIALIZED VIEW` takes an ACCESS EXCLUSIVE lock and can run for a while on big matviews, but the frontend shows nothing until the success/error alert. A disabled state or spinner while the call is in flight would help. Not asking for CONCURRENTLY here — that needs a unique index, separate discussion.

**Index list styling** — in `SidebarViewItem` the index sub-tree uses literal colors (`text-green-400`, `text-blue-400`, `border-strong`) instead of the theme tokens the rest of the sidebar uses. The grouping + rendering also looks close to what we already do for table indexes; worth reusing instead of a second copy.

Smaller stuff:
- `get_materialized_view_definition` rebuilds the DDL from `pg_get_viewdef`, so it loses storage params / tablespace / `WITH [NO] DATA`. Fine for a read-only view, just noting it's not the exact statement.
- `materializedViewsOpen` defaults to `true` while routines/triggers default to `false`.
- `get_materialized_view_columns` passes `&query` (a `&&str`) where `get_materialized_views` passes the literal — the `&` isn't needed.
- No Rust tests for the four new Postgres functions, though that matches the existing driver code.

Checked and these are fine: double-click goes through `handleOpenView` so `SELECT * FROM …` works, and the `relkind IN ('r','m')` change in `get_indexes` is what lets matview indexes load.

Davydhh:
Thanks for the review — agree with all the points. All addressed, except the two you flagged as fine-as-is: the definition isn't byte-exact (added a comment noting it' reconstructed from pg_get_viewdef), and no Rust tests for the four wrappers (matches the existing driver code).

debba:
@Davydhh This looks great overall, really nice work.

One thing though: as far as I know materialized views in Postgres are read-only — you can't insert, update or delete rows directly, the only write operation is `REFRESH`. So we should disable data editing (cell edits, adding and deleting rows) for materialized views in the grid.

Once that's handled and the conflicts are resolved, I'll go ahead and merge. Thanks!

Davydhh:
@debba Good catch, thanks! You're right that materialized views only accept REFRESH, so I've made them read-only in the grid. I've also merged the latest `main` to clear the conflicts.

debba:
Thanks @Davydhh, everything's addressed and CI is green — merging this right away.

Since you've clearly gotten comfortable with the Postgres side of the codebase: I'm currently working on multi-database connection support (browsing multiple databases, each with its own schemas, from a single connection) over in #402. If you feel like helping out there too, you'd be very welcome — your work here on materialized views fits right into that direction.

PR #352, opened 2026-06-22
feat: allow passing a startup script per connection

### LINKED ISSUE #350: [Feat]: Allow passing a startup script
### Prerequisites

### 🚀 Feature Proposal

I would like to be able to define some SQL on a connection to be run on startup, similar to the "startup script" option that DataGrip offers:

Image

### Motivation

Sometimes you just need to `SET` something for each and every startup, like when you want to bypass the RLS in development. If you can't, I think there is basically no way to "just click on a table name and display the rows" - rendering much of the UI useless with RLS enabled.

### Example

`SELECT set_config('app.bypass_rls', 'on', false);` or similar `set_config` overrides per connection.

boredland:
Closes #350.

Adds an optional **startup script** to a connection — SQL run on every new pooled connection, so session settings like `SELECT set_config('app.bypass_rls', 'on', false);` (or any `SET`) apply to every query regardless of which pooled connection serves it. This is the DataGrip-style "startup script" requested in the issue, and makes RLS-bypass-in-dev practical.

### Changes
- `ConnectionParams.startup_script`, persisted with the connection (non-secret).
- Executed per physical connection so the setting sticks across the whole pool: MySQL/SQLite via sqlx `after_connect`, Postgres via deadpool `post_create`. Blank/whitespace scripts are skipped; multiple statements are separated by `;`.
- New **Advanced** tab in the connection editor with a startup-script field; i18n added for all 8 locales.

### Tests
SQLite pool tests (no DB server needed) cover: script runs on connect, blank script is skipped, invalid script surfaces an error.


boredland:
Good catch — fixed in 30b5548.

`build_connection_key` now folds a SHA-256 of the trimmed `startup_script` into the pool cache key (only when a script is set, so script-free connections keep their existing keys). Editing the script therefore yields a new key → a fresh pool whose connections run the updated script, instead of reusing the stale cached pool.

Added two unit tests: the key changes when the script changes, and a blank/whitespace script is treated as absent (no key fragmentation).

debba:
Thanks for this PR.

Couple of things before this goes in:

1. The field takes arbitrary SQL, so what's the story for scripts that error out? From the hooks it looks like a failed statement bubbles straight out of `after_connect`/`post_create` and the connection never opens, so a single typo in the script locks you out of the data source entirely. Is blocking the behaviour we actually want here? I'd lean towards letting the connection through and surfacing the script failure as a warning instead. If we do want to keep it hard-blocking (which is defensible for stuff like RLS bypass, where you don't want a half-configured session), then the error needs to clearly say it came from the startup script, otherwise it just reads like a bad host or wrong credentials.

2. Can we swap the plain textarea for `SqlEditorWrapper`? That's what the editor and the other SQL fields already use, so we'd get syntax highlighting, theming and the font settings for free and it stays consistent with the rest of the app.

boredland:
Both addressed in 2186106.

**1. Error story.** Kept it hard-blocking — for the RLS-bypass case a half-configured session silently serving queries is worse than a clear failure — but made the failure fast and clearly attributed, and distinct from a connectivity error.

Worth flagging what I found while doing this: for MySQL/SQLite the old behaviour wasn't actually "bubbles straight out". sqlx **swallows** `after_connect` errors (logs, hard-closes the connection, retries until `acquire_timeout`) and then returns `PoolTimedOut` — so a script typo read as "pool timed out", exactly the misleading case you described. Fixes:

- **MySQL/SQLite** now preflight the script on a throwaway connection at pool-creation time. A failed `execute` → `Startup script failed: …`; a failed `connect` is returned verbatim, so a bad host/credentials still reads as such rather than as a script error. The per-connection `after_connect` hook is unchanged — the script still runs on every pooled connection.
- **Postgres** (deadpool doesn't swallow/retry) just gets the attribution: `post_create` now reports `` `post_create` hook failed: Startup script failed: <cause chain> ``, surfaced on `ping`/first query.

One residual edge: a script that passes preflight but fails *non-deterministically* on a later pooled connection still hits sqlx's retry → `PoolTimedOut`. The realistic deterministic failures (typos, bad permissions, unknown settings) are caught and attributed.

**2. Editor.** Swapped the textarea for `SqlEditorWrapper`, same as the editor and the other SQL fields — syntax highlighting, theming and font settings for free. Kept the RLS example as the Monaco `placeholder`, so no i18n changes.

The SQLite integration test now asserts the error is attributed (`Startup script failed`) and returns immediately via the preflight instead of timing out.

src-tauri/src/pool_manager.rs
```diff
@@ -652,14 +737,21 @@ pub async fn get_sqlite_pool_with_id(
...
-        .max_connections(5) // SQLite has lower concurrency needs
-        .connect_with(options)
-        .await
-        .map_err(|e| {
-            log::error!("Failed to create SQLite connection pool: {}", e);
-            e.to_string()
-        })?;
+    let mut pool_options = sqlx::sqlite::SqlitePoolOptions::new().max_connections(5); // SQLite has lower concurrency needs
+    if let Some(script) = startup_script(params) {
+        run_sqlite_startup_script(&options, &script).await?;
```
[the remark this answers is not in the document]
boredland: Fixed in 947325b. Wrapped the SQLite preflight in a 30s timeout to match MySQL, so a broken (or stalled-mount/custom-VFS) script can never wedge pool creation. `cargo test --lib startup_script` still passes and returns immediately.

src-tauri/src/pool_manager.rs
```diff
@@ -525,12 +592,25 @@ async fn get_mysql_pool_for_database_with_id(
...
+    if let Some(script) = startup_script(params) {
+        tokio::time::timeout(connect_timeout, run_mysql_startup_script(&options, &script))
+            .await
+            .map_err(|_| {
+                format!(
+                    "Timed out running MySQL startup script after {} ms",
+                    connect_timeout.as_millis()
+                )
+            })??;
+        pool_options = pool_options.after_connect(move |conn, _meta| {
```
NewtTheWolf, line 652: **blocker** — the script runs **twice** on the first connection: once in the preflight throwaway connection above, then again here in `after_connect`. A side-effecting script (`INSERT`, counters, …) firing an extra time on pool creation is a real footgun. Please make the preflight validation-only (don't execute side effects) so the script runs exactly once per connection.

src/components/modals/NewConnectionModal.tsx
```diff
@@ -2038,6 +2137,27 @@ export const NewConnectionModal = ({
                 </button>
               ))}
             </div>
+              {tabFade.left && (
+                <button
+                  type="button"
+                  aria-label="Scroll tabs left"
```
NewtTheWolf: **blocker** — hardcoded `aria-label="Scroll tabs left"` (and `right` on L2153) is a user-facing string; it needs to go through i18n per [`general.md`](.rules/general.md). Please use `t(...)`.

src/components/modals/NewConnectionModal.tsx
```diff
@@ -186,9 +191,52 @@ export const NewConnectionModal = ({
...
+    const { scrollLeft, scrollWidth, clientWidth } = el;
+    setTabFade({
+      left: scrollLeft > 1,
+      right: scrollLeft + clientWidth < scrollWidth - 1,
+    });
+  }, []);
+
+  // Recompute fades when the visible tab set changes and keep the active tab
+  // scrolled into view; also follow window resizes.
+  useEffect(() => {
```
NewtTheWolf: **blocker** — calling `setTabFade()` synchronously in `useEffect` violates [`react.md`](.rules/react.md) #2. This is a legit DOM-measure case, so `useLayoutEffect` is the right fit (also avoids a one-frame flash of the wrong fade state).

```suggestion
  useLayoutEffect(() => {
```
(add `useLayoutEffect` to the `react` import on line 1)

src-tauri/src/pool_manager.rs
```diff
@@ -597,14 +677,24 @@ pub async fn get_postgres_pool_with_id(
...
-        .max_size(10)
-        .build()
-        .map_err(|e| {
-            let detail = format_error_chain(&e);
-            log::error!("Failed to create PostgreSQL connection pool: {}", detail);
-            detail
-        })?;
+    let mut builder = PgPool::builder(PgPoolManager::new(cfg, tls_connector)).max_size(10);
+    if let Some(script) = startup_script(params) {
+        builder = builder.post_create(PgHook::async_fn(move |client, _metrics| {
```
NewtTheWolf, line 744: **nit** — when the script is invalid, Postgres surfaces deadpool's raw `PostCreateHook(Message(...))` debug struct to the UI, whereas MySQL/SQLite return a clean `Startup script failed: …` from the preflight. Worth giving PG a symmetric preflight, or unwrapping the `PoolError` to show just the inner message. Purely cosmetic.

NewtTheWolf (changes requested):
Thanks for this — really clean work! 🙏 Pool-key fold for cache invalidation, error attribution, and the preflight-vs-`post_create` asymmetry are all well done. Tested locally against a Postgres RLS fixture (restricted role + `app.bypass_rls` policy): 2 → 4 → 2 rows across the pool, broken script → attributed error, 5/5 Rust tests green. Three blockers, one nit below.

NewtTheWolf (approved):
All four points addressed cleanly — thank you! 🙏 The transaction-rollback preflight is a neat way to keep the script side-effect-free until the per-connection hook, and giving Postgres a symmetric preflight kills the deadpool debug leak. Verified locally: builds, 5/5 Rust tests green.

One tiny FYI (non-blocking): the preflight validates inside a transaction, so a non-transactional statement (`VACUUM`, `CREATE INDEX CONCURRENTLY`, …) would be rejected here even though it would run fine in the real `post_create`/`after_connect` hook — and MySQL implicitly commits DDL, so a DDL script would still apply once despite the rollback. Both are irrelevant for the intended `SET`/`set_config` use; just worth a one-line code comment if you want. LGTM either way.

boredland:
> One tiny FYI (non-blocking): the preflight validates inside a transaction, so a non-transactional statement (VACUUM, CREATE INDEX CONCURRENTLY, …) would be rejected here

True, but it's hard for me to come up with a reason to do that.

debba:
It looks like something non-blocking.
 I'll merge it!

PR #354, opened 2026-06-22
Feat/customizable result colors

GabrielMalava:
This PR adds two improvements to the data experience:

1. Customizable result colors by data type: color query result values
   (numbers, text, dates, booleans) instead of rendering everything in a
   single color.
2. Better in-place editing: save pending grid edits with Cmd/Ctrl+S and
   make editing single-table SELECT results reliable.

## 1. Result colors by data type

You can now colorize query result cells based on their data type, with full
control over the colors.

- New Result Colors section under Settings → Appearance → General.
- Toggle to enable/disable colorization (off by default — values render as
  before).
- A color picker per type: **Number, Text, Date/Time, Boolean**, with a live
  preview and a "Reset to theme" button per type.
- Defaults follow the active theme's semantic colors; user overrides are saved
  in the app config and applied on top of any theme.
- Colors are applied only to plain data cells — edited/inserted/deleted rows
  and NULL keep their existing styling.

## 2. Editing improvements

Save with Cmd/Ctrl+S
- New `save_grid_changes` shortcut commits the active tab's pending changes,
  like TablePlus. It's rebindable and listed under **Settings → Keyboard
  Shortcuts**.

## Editable single-table SELECT results
- Editing query results is now validated against the table's real columns
  before building the UPDATE.
- Editing an aliased/computed column shows a clear message instead of failing
  with a cryptic `1054 Unknown column` error.
- If the primary key is missing from the result set, editing is blocked with
  guidance to include it (needed for a safe `WHERE` clause).
- Unsafe-to-edit queries (JOIN, DISTINCT, aggregates, UNION, subqueries,
  views) remain read-only, as before.


debba:
Really love it! Merging now! 🎉

If you'd like to support the project, we'd really appreciate it if you could:

- ⭐ Drop a star on the repo
- 💬 [Join us on Discord](https://discord.com/invite/K2hmhfHRSt) to chat, share ideas and get support
- 📱 Follow us on social to stay up to date with new releases

Thanks again, and welcome aboard! 🙌

PR #355, opened 2026-06-23
fix: improve accessibility for screen reader users

### LINKED ISSUE #86: [Bug]: multiple accessibility problems
### Describe the bug

Tabularis has some accessibility problems for screen reader users. For the following, I used 0.9.6 on Windows 11 with the NVDA screen reader.

The biggest problem is the query editor.

1.  NVDA does not read any text entered into the field, making it impossible to review or edit a query. I had to use a different editor and then paste my query.
2. If I tried to type, I was jumped to a list of keyword suggestions. This caused NVDA to exit focus mode (the mode where the user can type instead of having letters interpreted as screen reader-specific navigation commands). This made for a jarring and frustrating experience. Of course, I quickly found that I couldn't review anything in the query field anyway.
3. I had to press control-enter to submit the query. That's fine, and I like when that command is used. The problem is that it was not stated anywhere, and there was no button to submit the query.

Other things I noticed:

* In the table of query results, every table cell is spoken as clickable. Normally, this indicates that an element that isn't normally actionable (td, span, p, and so on) will do something when clicked or when a screen reader user presses enter or space on it. Doing so on these cells, even the header cells, did nothing at all. Should it have sorted the results or something?
* When adding a connection, it would be nice if the test results were spoken using a live region, or gained focus automatically.
* In the list of all tables in the database, there is an unlabeled button after each table name that seems to open more details about it. This button needs a label, and ARIA should be used to indicate when a table's information is expanded or collapsed.

### To Reproduce

To reproduce, just use the application with NVDA on Windows. I suspect other platforms will run into similar problems.

### OS Version

Windows 11

### Tabularis Version

0.9.6

### Relevant Log Output

```shell

```

arturbent0:
- Enable Monaco editor accessibility support (accessibilitySupport: 'on') to allow screen readers to read typed text and prevent focus jumping to autocomplete suggestions
- Add aria-label and aria-keyshortcuts to Run button to announce keyboard shortcut to screen readers
- Add role, aria-sort, tabIndex and keyboard support to DataGrid column headers
- Add aria-live region to connection test result message
- Add aria-label and aria-expanded to sidebar expand buttons, convert clickable divs to semantic button elements

Fixes #86


NewtTheWolf (dismissed):
Thanks for tackling this — solid accessibility work! 🙏 The `div`→`button` conversions, the always-mounted `aria-live` region (exactly the right pattern so it actually gets announced), and the keyboard support on sort headers are all good, and the i18n is complete across all 8 locales. Typechecks clean.

Three small things — all **optional**, none blocking. Happy to merge as-is if you'd rather leave them:

1. **`aria-sort` (mentioned in the description but not in the diff)** — `src/components/ui/DataGrid.tsx` sort headers get `role`/`tabIndex`/`aria-label`, but not `aria-sort="ascending|descending|none"`, which is what actually announces the *current* sort state to a screen reader. Would round out exactly what the PR set out to do.
2. **Monaco `accessibilitySupport: 'on'`** forces screen-reader rendering for every user; the default `'auto'` already enables it when a screen reader is detected and is the recommended setting — worth considering `'auto'` instead.
3. **`aria-keyshortcuts="Control+Enter"`** is hardcoded, so on macOS (Cmd+Enter) the announced shortcut won't match.

Want to fold these in? Otherwise it's good to go from my side. 👍

arturbent0:
Thanks for the review @NewtTheWolf 

1. I intentionally used aria-label with the sort state description instead of aria-sort, since the bot flagged that aria-sort is invalid on role="button" elements. The aria-label dynamically reflects the current sort state, which communicates the same information to screen readers without the invalid ARIA. 
2. switched to auto
3. Fixed for mac users

debba:
Thanks a lot for this, Artur. Accessibility is something we really care about and these are exactly the kind of changes that make a real difference for people, so I appreciate you taking the time.

I'll merge this shortly. Please keep them coming, improvements like this are genuinely important to us.

If you feel like it, come hang out on our Discord: https://discord.com/invite/K2hmhfHRSt

PR #356, opened 2026-06-23
Add function to include headers in CSV output

Wilovy09:
## Summary

- Add `rowsToCSVWithHeaders()` utility to `clipboard.ts` that prepends column names as the first row in CSV output
- Clicking the `#` corner button now selects all rows **and** copies them to clipboard with headers included
- `Ctrl+C` / `Cmd+C` on selected rows now also includes the column header row in the CSV output

## Motivation

When pasting query results into a spreadsheet or sharing via chat, the output was missing column names — requiring the user to manually add them. Headers are now always included when copying multiple rows, making the output immediately usable without post-processing.


debba:
Hi @Wilovy09, thanks for this 🙌 I pushed a small enhancement on top of your branch to make the header behavior configurable instead of always-on.

### What I added
- A new **`csvIncludeHeaders`** setting, persisted in `config.json` (defaults to `true`, so current behavior is unchanged).
- A toggle in the **copy toolbar** (next to the CSV delimiter selector, shown only when the copy format is CSV) to choose whether copied CSV output includes the column header row — mirroring how the delimiter selector works.
- The same option in **Settings → General**, right under "CSV Delimiter".
- Threaded the flag through `MultiResultPanel → StackedResultItem / ResultEntryContent → DataGrid`, so it applies both to the "select all + copy" corner button and to `Ctrl`/`Cmd`+`C`.
- i18n keys (`csvIncludeHeaders`, `csvIncludeHeadersDesc`, `csvHeaders`) added across all 8 locales.
- Backend: `csv_include_headers: Option<bool>` added to `AppConfig` with merge handling in `save_config`.

Verified with `eslint`, `tsc` / `vite build`, and `cargo check` — all green ✅

> Note: this covers the **copy-to-clipboard** path (the toolbar in the screenshot). The CSV **file export** still always writes headers via the Rust export pipeline — happy to extend it there too in a follow-up if we want full parity.

From my side this is ready to merge — let me know if you're good with it and I'll merge it in.

Wilovy09:
@debba 
lgtm 👍

debba:
Great, I will merge it @Wilovy09

Issue #323, opened 2026-06-14
[Bug]: Kubectl is not detected


rucciva:
### Describe the bug

kubectl is not detected even though it is installed

<img width="1800" height="912" alt="Image" src="https://github.com/user-attachments/assets/c03c2c5b-d3fe-4bd4-9e6a-5afc1bd15a01" />

```
$ which kubectl
/opt/homebrew/bin/kubectl
$  kubectl version
Client Version: v1.31.1
Kustomize Version: v5.4.2
...
```

### To Reproduce

1. go to kubectl setting
2. see that no context available

### OS Version

macOS 14.5

### Tabularis Version

v0.13.1-beta

### Relevant Log Output

```shell

```

debba:
Hi @rucciva ,
thanks for this feedback!
Are you interested in open a PR?

rucciva:
Hi @debba , to be honest not sure yet, i haven't yet discover where the problem lies.

metalgrid:
I believe the message may be misleading. I reproduced this by moving my kubeconfig away from the standard location.

It currently doesn't propagate $KUBECONFIG to the managed kubectl process so if it's in a location managed by an env var it finds nothing.

@rucciva would you please help out to confirm this?

Run the following, edit out any sensitive information if any and paste it here:
```
[ -z "$KUBECONFIG" ] && echo "no custom kubeconfig" || echo "custom KUBECONFIG envvar set"
kubectl config get-contexts --no-headers | wc -l
```

I have some improvements for the kubernetes options in #319 , I believe this should be handled there as well.

rucciva:
Ah yes confirmed, my $KUBECONFIG is indeed empty, and it works after setting the value. Thanks @metalgrid

PR #365, opened 2026-06-23
Introduce kubectl and kubeconfig overrides

metalgrid:
### Overview

 Adds the ability to override the kubectl binary path and KUBECONFIG file per
 Kubernetes connection, plus on-blur validation of those paths in both the K8s
 connections modal and the new-connection modal's inline K8s flow.

 ### What's new

 Backend (Rust)
 - New optional fields `kubectl_path` / `kubeconfig_path` on `K8sConnection` and
   `K8sConnectionInput` models, persisted to `k8s_connections.json`.
 - `K8sCommandOptions` plumbing through the tunnel, discovery commands (contexts,
   namespaces, resources, resource ports), connection test, and MCP connection
   expansion — so overrides take effect for port-forward tunnels and the cascading
   discovery dropdowns.
 - New Tauri command `validate_k8s_path_cmd` that validates:
     - `kubectl`: the path exists, is a file, and is executable (resolves bare names
       like kubectl via `PATH`, honors `PATHEXT` on Windows).
     - `kubeconfig`: the path exists and is a file.
 - `kubectl_command` now preflight-validates explicit overrides before spawning kubectl,
   giving clear errors instead of cryptic spawn failures.
 - Path helpers handle `~` expansion and `PATH` lookup.

 Frontend (TS/React)
 - Advanced collapsible "kubectl settings" section in both `K8sConnectionsModal` and
   `NewConnectionModal` (inline K8s) with kubectl Path and Kubeconfig Path inputs.
 - On-blur validation with live status: validating spinner, green border + tick when
   valid, red border + error message when invalid.
 - Discovery dropdowns reset context/namespace/resource whenever the applied
   kubectl/kubeconfig paths change, and Save/Test are blocked until paths are valid
   (with a prompt to re-select context/namespace/resource when paths change).
 - `validateK8sPath` utility + `K8sCommandOptions` threaded through all K8s discovery/test
   calls.

 i18n
 - New keys (kubectlPath, kubeconfigPath, placeholders, advancedSettings,
   validation/reset messages) across all 8 locales (en, it, de, es, fr, ja, ru, zh).

 Tests
 - Rust unit tests for path validation (moved to `k8s_tunnel/tests.rs` per project Rust
   rules): empty paths valid, missing file, directory rejected,
   executable/non-executable file, Windows/Unix handling.
 - Frontend tests for `validateK8sPath` invoke contract; updated `NewConnectionModal` test
   for the added `k8sCommandOptions` arg.

Provides native tooling for resolving issues like #323.


metalgrid:
@kilocode-bot please review the new changes.

src/components/modals/NewConnectionModal.tsx
```diff
@@ -459,6 +567,178 @@ export const NewConnectionModal = ({
...
+        applyK8sAdvancedFields();
+      }
+      return true;
+    }
+
+    setPathValidation({ status: "validating", value: trimmedPath, message: null });
+    return run(kind, async (isLatest) => {
+      try {
+        await validateK8sPath(trimmedPath, kind);
+        if (isLatest) {
```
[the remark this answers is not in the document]
metalgrid: Whoopsie.

src/utils/autocomplete.ts
```diff
@@ -347,7 +357,9 @@ export const registerSqlAutocomplete = (
         label: t.name,
         kind: monaco.languages.CompletionItemKind.Class,
         detail: "Table",
-        ...buildIdentifierInsert(t.name, range),
+        ...(t.schema
+          ? { insertText: tableInsertText(t.name, driver, t.schema), range }
```
NewtTheWolf: **Should fix (unrelated to kubectl — ideally split out):** schema-qualified tables (multiDb MySQL) take the `tableInsertText(...)` branch and skip `buildIdentifierInsert`, so they lose its opening-quote handling and `filterText`. If the user types an opening `` ` `` (Monaco auto-closes) and completes, this inserts `` `db`.`table` `` over the existing backtick → doubled/stray quote, and the missing `filterText` makes Monaco filter against the backtick and drop the suggestion. Route the schema branch through quote-aware insert too.

src/components/modals/K8sConnectionsModal.tsx
```diff
@@ -249,6 +361,159 @@ export function K8sConnectionsModal({
...
+    setPathValidation({ status: "validating", value: trimmedPath, message: null });
+    return run(kind, async (isLatest) => {
+      const isCurrent = () =>
+        isLatest() && isCurrentPathValidation(kind, validationToken);
+      try {
+        await validateK8sPath(trimmedPath, kind);
+        if (!isCurrent()) return false;
+        setPathValidation({ status: "valid", value: trimmedPath, message: null });
+        if (applyOnSuccess && areAdvancedPathsReadyToApply({ kind, value: trimmedPath })) {
+          applyAdvancedPaths();
```
NewtTheWolf: **Should fix (stale-apply race):** `validateAdvancedPath`'s `run()` callback reads a render-time snapshot of `areAdvancedPathsReadyToApply`. Edit *both* kubectl and kubeconfig paths and blur them before either validation resolves → each callback sees the sibling still `validating`, so neither reaches this `applyAdvancedPaths()`. Both fields go green but `appliedKubectl/KubeconfigPath` stay stale and the context/namespace/resource dropdowns keep the previous cluster's selection until a later Save/Test. Decide from the freshly-resolved results (functional `setState`/ref) or reconcile in an effect once both are `valid`. Same defect in `NewConnectionModal.tsx`.

src/components/modals/NewConnectionModal.tsx
```diff
@@ -513,6 +639,184 @@ export const NewConnectionModal = ({
...
+    setPathValidation({ status: "validating", value: trimmedPath, message: null });
+    return run(kind, async (isLatest) => {
+      const isCurrent = () =>
+        isLatest() && isCurrentK8sPathValidation(kind, validationToken);
+      try {
+        await validateK8sPath(trimmedPath, kind);
+        if (!isCurrent()) return false;
+        setPathValidation({ status: "valid", value: trimmedPath, message: null });
+        if (applyOnSuccess && areK8sAdvancedPathsReadyToApply({ kind, value: trimmedPath })) {
+          applyK8sAdvancedFields();
```
NewtTheWolf: **Should fix (same stale-apply race as K8sConnectionsModal):** concurrent kubectl+kubeconfig blurs read a stale validation snapshot, so this `applyK8sAdvancedFields()` never fires and `k8s_context`/`namespace`/`resource` are not reset after both paths change — until a later Save/Test. Fix both modals together (see the shared-hook note in the summary).

src-tauri/src/k8s_tunnel.rs
```diff
@@ -182,130 +221,116 @@ impl K8sTunnel {
...
+        &["config", "get-contexts", "-o", "name"],
+        options,
+        "Failed to list K8s contexts",
+    )?;
 
-    if !output.status.success() {
-        let stderr = String::from_utf8_lossy(&output.stderr);
-        return Err(format!("Failed to list K8s contexts: {}", stderr.trim()));
+    let contexts = parse_lines(&String::from_utf8_lossy(&output.stdout));
+    if contexts.is_empty() {
```
NewtTheWolf: **Question:** `get_k8s_contexts` now returns `Err(no_contexts_error(...))` when kubectl exits 0 with empty output (valid kubeconfig, no contexts / in-cluster). Previously `Ok([])`, and this is asymmetric — `get_k8s_namespaces`/`resources`/`ports` still return `Ok(empty)` on the same success-but-empty case. Intentional UX choice, or should it stay `Ok(vec![])`?

src-tauri/src/k8s_tunnel.rs
```diff
@@ -392,194 +409,321 @@ pub fn get_k8s_resource_ports(
...
+    )?;
+
+    Ok(parse_resource_ports(&String::from_utf8_lossy(
+        &output.stdout,
+    )))
+}
 
-    if !output.status.success() {
-        let stderr = String::from_utf8_lossy(&output.stderr);
+fn kubectl_command(options: &K8sCommandOptions) -> Result<Command, String> {
```
NewtTheWolf: **Question:** `kubectl_command` re-`fs::metadata`s the kubectl/kubeconfig paths on every invocation — each cascading discovery call and the tunnel spawn. A path that's valid at save time but on a network/removable mount that briefly blips becomes a hard "file not found" instead of letting kubectl handle it. Consider validating at save/apply time and letting `kubectl_spawn_error` cover runtime failures.

src-tauri/src/k8s_tunnel.rs
```diff
@@ -182,130 +221,116 @@ impl K8sTunnel {
...
-                format!(
-                    "kubectl not found: {}. Please install kubectl and ensure it is in PATH.",
-                    e
-                )
-            })?;
-
-        if !output.status.success() {
-            return Err("kubectl version check failed. Please verify your kubectl installation.".to_string());
+    /// Check that kubectl is available.
+    fn verify_kubectl(options: &K8sCommandOptions) -> Result<(), String> {
```
NewtTheWolf, line 205: **Cleanup:** `run_kubectl` only returns `Ok` on success, so the `else` branch here is unreachable. Collapses to:
```rust
run_kubectl(&["version", "--client"], options, "kubectl version check failed. Please verify your kubectl installation").map(|_| ())
```

src-tauri/src/k8s_tunnel.rs
```diff
@@ -392,194 +409,321 @@ pub fn get_k8s_resource_ports(
...
+        None => value.to_string(),
+    }
+}
 
-        #[test]
-        fn test_ignores_invalid_values() {
-            let result = parse_resource_ports("abc 3306 70000 8123");
-            assert_eq!(result, vec![3306, 8123]);
-        }
+fn home_dir() -> Option<PathBuf> {
```
NewtTheWolf: **Cleanup:** hand-rolled `home_dir()` reinvents `directories::BaseDirs`, which is already a dependency and already used for home resolution in `mcp/install.rs`. Reuse `BaseDirs::new()?.home_dir()` to stay consistent with the rest of the app.

src-tauri/src/k8s_tunnel.rs
```diff
@@ -392,194 +409,321 @@ pub fn get_k8s_resource_ports(
...
-            let key = build_tunnel_key(
-                "my-cluster",
-                "default",
-                "service",
-                "my-db",
-                3306,
-            );
-            assert_eq!(key, "my-cluster:default:service/my-db:3306");
-        }
+fn find_executable_in_path(command: &str) -> Option<PathBuf> {
```
NewtTheWolf: **Cleanup:** `find_executable_in_path` + `executable_candidate_names` + `windows_executable_extensions` reimplement a `which`-style PATH/PATHEXT lookup (~70 lines) purely to pre-validate bare command names. Spawn failures are already reported clearly via `kubectl_spawn_error`, so this is speculative — drop it for bare names, or use the `which` crate rather than hand-rolling the Windows edge cases.

NewtTheWolf (commented):
**Overall:** Solid core work — the `run_kubectl`/`K8sCommandOptions` refactor is a real improvement, tests correctly moved to `k8s_tunnel/tests.rs`, serde defaults sound, and KUBECONFIG is applied uniformly. My main concern isn't any single line — it's **scope**: this 37-file PR bundles kubectl overrides + an autocomplete fix + type-split refactors + a new `useLatestAsync` hook. Splitting the non-k8s changes into their own PRs would make this far easier to review and test in isolation.

Two items I'd want addressed before merge (inline: autocomplete quote bug, and the stale-apply race in both modals); the rest are questions or non-blocking cleanup.

One cross-cutting cleanup that doesn't fit a single line: the path-validation UI (`PathValidationState`, `EmptyPathValidationState`, `PathValidationAdornment`/`Message`, the validate/apply/bump logic, plus `DiscoveryError`) is duplicated ~150 lines between `K8sConnectionsModal.tsx` and `NewConnectionModal.tsx`. Extracting a shared hook + component means the stale-apply fix below lands once instead of twice.

metalgrid:
Feature is rewritten on a clean branch from main because it got stale and missed out on using improved tooling provided by the progressed main branch. Hopefully, it's now in a good state. It's accumulated a lot of noise in the PR. I can close this one and reopen fresh if it makes things easier to reason.

PR #367, opened 2026-06-23
informix: ship the 32-bit build on the Windows slot (v0.1.3)

danielnuld:
Follow-up to the IBM Informix plugin entry.

**Problem:** the `win-x64` asset was the 64-bit build, but the **IBM Informix ODBC driver (CSDK) is 32-bit in virtually all deployments**. A 64-bit process cannot load a 32-bit ODBC driver, so one-click installs fail with ODBC `IM002` ("data source / driver not found") even though the driver is installed.

**Fix:** point the `win-x64` registry slot at the **32-bit (`win-x86`) build**. The plugin runs as a separate stdio process, so its bitness must match the **Informix driver**, not the 64-bit host — a 32-bit plugin talks to Tabularis over stdio just fine. This makes the default install work against the standard 32-bit CSDK.

Also bumps the entry to **v0.1.3** (adds linux-arm64 and darwin-arm64 assets; release also hides the console window on Windows). All asset URLs verified downloadable.


debba:
Hi @danielnuld ,

I noticed that each time you delete the previous release and replace it with a new one. The correct approach would be to keep all releases and simply update latest_version. That way, users can always downgrade to an earlier version if needed.

Would you mind updating it to keep all the releases instead of replacing them?

This process will be much simpler with the new registry we’re currently testing for the upcoming release!

danielnuld:
Hi @debba, thanks for the heads-up — that makes sense.

I've updated the entry so previous releases are kept instead of being replaced: `releases` now lists **0.1.4 → 0.1.3 → 0.1.2**, and only `latest_version` points to the newest (0.1.4). Users can downgrade if needed.

Note: I pointed the `win-x64` slot at the **32-bit (`win-x86`)** build on every version, not just 0.1.4 — otherwise a downgrade would reintroduce the 64-bit/CSDK `IM002` problem this PR fixes. All asset URLs verified downloadable (200).

I left 0.1.0/0.1.1 out since they only ever had Windows assets and were never in the registry, but I can add them if you prefer the full history.

Issue #195, opened 2026-05-14
UI design system & visual identity — call for contributors


debba:
I'm a backend dev. I built most of the current UI by eye, without any design background. It works, but it looks like what it is: a frontend assembled one modal at a time, with colors and spacing picked on the spot.

Some examples of what I mean:

- the logo is a placeholder I never replaced
- the color palette grew organically. The CSS variables in `src/index.css` (`bg-base`, `bg-elevated`, `text-primary`, etc.) are an attempt to centralize it, but the values themselves were chosen by trial and error
- spacing is inconsistent across pages. You'll find `p-3`, `p-4`, `p-5` on similar-looking containers depending on which file I was in that day
- button, input and modal styles drift between the Connection modal, the Settings page, the plugin modals and the query editor toolbar
- icons (lucide) get a different size or stroke depending on the component
- light/dark themes are kept in sync manually, so they fall out of sync

What I'd like to do, eventually:

- a real logo, in the sizes the app actually needs (titlebar, about, favicon, GitHub social preview)
- a documented palette with semantic roles (surfaces, text levels, accent, success/warning/danger/info), with light and dark generated from the same source instead of maintained twice by hand
- one type scale, one spacing scale, used everywhere
- shared primitives for button/input/select/tab/modal/tooltip. `SettingControls.tsx` already does this for the Settings page — the same pattern should cover the rest of the app
- a short doc somewhere that says \"here's the palette, here's the spacing scale, here's how to build a modal,\" so the next contributor doesn't have to reverse-engineer my taste

To be clear: this isn't a UX rework. The sidebar, the tabs, the settings layout, the query editor flow — they're fine. This is purely about visual consistency and having a written-down system instead of vibes.

If you have a design background and want to weigh in (palette proposals, a logo concept, a Figma file, or just opinions on the current state), please comment. I'd rather do this once with someone who knows what they're doing than keep iterating on instinct.

For devs: once direction is decided the migration can be split into smaller follow-up issues, primitive by primitive.

wajrock:
Hi I'd like to work on this, are you still looking for someone ?

debba:
There is a girl who is working on it, but having another one would be wonderful! 
So you are more than welcome!

If you want you can reach us on Discord: https://discord.com/invite/K2hmhfHRSt

wajrock:
Done! Is it good for you ?

PR #369, opened 2026-06-24
feat(editor): window controls and detachable results panel

debba:
## What this does

The query results panel had a single chevron to collapse it and not much else. This PR turns the panel header into the familiar set of window controls and adds the ability to pop results out into their own window.

Manual drag-to-resize of the panel is untouched.

### Window controls (right side of the results bar)

- **Minimize** and **Close** collapse the panel — no data is lost, the existing "Show Results" button brings it back.
- **Maximize** hides the editor so results take the full height; the same button restores the split.
- **Detach** opens the results in a separate window.

### Detachable results

Detach pops the active tab's results into their own OS window. While detached, the in-app panel shows a "results detached" placeholder with a **Re-attach** button.

Each window is bound to the tab it was opened from, so:

- it keeps showing that tab's results even when you switch tabs in the main window;
- you can detach several tabs into several windows at the same time;
- closing a tab closes its detached window.

The main window stays the single source of truth. It pushes result state to each detached window over Tauri events keyed by tab id, and the windows send user actions back (pagination, page change, entry select / close / rename, load count) to be applied to the right tab. No query or DB logic is duplicated in the detached window — it renders the existing `MultiResultPanel` / `ResultEntryContent` and forwards everything else.

## How it's built

- `src-tauri/src/results_window.rs` — `open_results_window` / `close_results_window` commands; one window per tab (label `results-window-{tabId}`), remembered bounds, emits `results-window:closed` on close.
- `src/utils/resultsWindowSync.ts` — the event protocol (sync payload, action envelope) plus pure helpers (`buildSyncPayload`, `applyAction`, `singleResultToEntry`), with unit tests.
- `src/pages/ResultsWindowPage.tsx` — the detached window, reuses the existing result components.
- `src/pages/Editor.tsx` — the header controls and the sync/action wiring.
- New tooltip and placeholder strings added across all 8 locales.

## Scope / not in this PR

In-place row editing (insert / delete / pending changes) is intentionally not available from inside the detached window for now — that machinery is tightly coupled to the main editor. In-grid scroll, sort, cell copy and pagination all work.

## Testing

- `npx tsc --noEmit` — clean
- `npm run build` — clean
- `npm test` (`resultsWindowSync` plus the full suite) — green
- `cargo check` and `cargo test results_window` — green

## Context

Framed as part of the UI consistency effort in #195. This isn't a design-system change in itself, but it brings the results panel in line with conventional window affordances and adds a feature that was awkward to do before. Relates to #195.


NewtTheWolf (dismissed):
Reviewed locally — really nice feature, and cleanly built: the action handlers are properly tab-keyed, the main window stays the source of truth, and no query/DB logic is duplicated. Unit tests pass (11/11), tsc clean, i18n complete across all 8 locales, and the new Tauri commands are registered. 🙏

I can confirm all three of @kilo-code-bot's warnings are real — they're shared main-window UI state leaking into actions a *detached* window triggers, so no data loss, but worth tidying:

1. [`Editor.tsx:698`](https://github.com/TabularisDB/tabularis/blob/e9196af214fceaacb7a780f6ec80db2abf747cff/src/pages/Editor.tsx#L698) — `runQuery` calls `setIsResultsCollapsed(false)` (and would pop `setQueryParamsModal` for parameterised queries), so a re-run/page-change from a detached window **re-expands the main panel** — which slightly defeats the point of detaching. This is the one I'd fix before merge.
2. [`Editor.tsx:1095`](https://github.com/TabularisDB/tabularis/blob/e9196af214fceaacb7a780f6ec80db2abf747cff/src/pages/Editor.tsx#L1095) — `loadCount` flips the global `setIsCountLoading`, so a detached count-load shows a spinner on the *active* tab. Cosmetic.
3. [`Editor.tsx:1292`](https://github.com/TabularisDB/tabularis/blob/e9196af214fceaacb7a780f6ec80db2abf747cff/src/pages/Editor.tsx#L1292) — `applyAction` has no `detachedTabIds.has(tabId)` guard; low risk since events only come from app-owned windows, but cheap defense-in-depth.

#2 and #3 are polish and fine as follow-ups. Approving — just flagging #1 as worth a look first.

debba:
Thanks for the thorough review! Fixed all three — they boiled down to detached-window actions leaking into main-window-only UI state, so I gated each on whether the target tab is actually detached.

To do that without re-creating the callbacks or reading stale closures, I added a `detachedTabIdsRef` kept in sync next to `tabsRef`/`activeTabIdRef`:

1. **`runQuery`** — `setIsResultsCollapsed(false)` and the `setQueryParamsModal` pop now only fire when the target tab is *not* detached, so a re-run/page-change from a detached window no longer re-expands the main panel (and won't hijack the active tab with a params modal).
2. **`loadCount`** — `setIsCountLoading` is skipped for detached counts, so the spinner no longer flashes on the active main-window tab.
3. **`applyAction`** — the action listener now early-returns unless `detachedTabIdsRef.current.has(tabId)`, as defense in depth against events for a reattached/unknown tab.

Build, `tsc`, and ESLint are clean, and the full test suite passes (2635/2635).

NewtTheWolf:
Re-reviewed locally after merging `main` in. **No blockers — mergeable as-is.** A few minor nits I noticed; happy to leave them as follow-ups or fix in this PR, your call:

1. **`react.md` #2 — sync `setState` in effect** (`Editor.tsx`, orphan-cleanup effect): `setDetachedTabIds(...)` is called synchronously inside the `useEffect` body. Could let the `results-window:closed` listener own the pruning so the effect stays side-effect-only.
2. **Detach init race** (`Editor.tsx`): the `ready`/`action`/`closed` listeners are gated behind `detachedTabIds.size > 0` and `listen()` registers async, so a freshly opened window's `ready` handshake can be missed (window stuck on "Loading…"). Registering them unconditionally (the action listener already self-guards via `detachedTabIdsRef`) removes the race.
3. **Window bounds are global** (`results_window.rs`): `last_bounds: Mutex<Option<WindowBounds>>` is shared across all detached windows, so with multi-detach each new window opens on top of the last one. A `HashMap<tab_id, _>` would remember per tab.
4. **outer vs inner size drift** (`results_window.rs`): bounds are saved from `outer_size()` but restored via `.inner_size(...)`, so the window grows by the decoration height on each detach→close→reopen cycle.
5. *(tiny)* the two `results_window` tests only exercise `Mutex<Option<_>>` get/set; `window_label()` is the actual pure helper and is untested.

Want me to push fixes for any of these, or merge and follow up separately?

debba:
Thanks for the thorough re-review! I've addressed all five points in two commits (`a95471e` and `e4f5c6c`):

1. **Sync `setState` in effect** — the orphan-cleanup effect no longer prunes `detachedTabIds` itself. Closing the window emits `results-window:closed`, and that listener now owns the pruning, so the effect stays side-effect-only.
2. **Detach init race** — removed the `detachedTabIds.size > 0` gate; the `ready`/`action`/`closed` listeners are now registered unconditionally at mount, so a freshly opened window's `ready` handshake can't be missed. Each handler self-guards (action via `detachedTabIdsRef`, ready via the `tabsRef` lookup, closed via `prev.has`), and `detachedTabIds` was dropped from the deps to avoid needless re-registration.
3. **Global window bounds** — `last_bounds: Mutex<Option<WindowBounds>>` is now `bounds: Mutex<HashMap<String, WindowBounds>>`, keyed by `tab_id`, so each detached window reopens at its own last position instead of stacking.
4. **outer vs inner size drift** — bounds are now saved from `inner_size()` to match the `.inner_size(...)` restore, so the window no longer grows by the decoration height on each detach→close→reopen cycle.
5. **Tests** — added coverage for the `window_label()` pure helper and for per-tab bounds storage.

PR #372, opened 2026-06-24
feat(views): add SQL beautify button to view editor

danielnuld:
## What

Adds a **Beautify** button to the view editor modal (create & edit) that pretty-prints the view definition SQL.

## Why

When editing a view, the definition returned by the database is often a single dense line (e.g. MySQL stores `SELECT c.id,c.name,...` with no whitespace). This makes it hard to read and edit. A one-click formatter improves the editing experience.

## How

- Adds [`sql-formatter`](https://github.com/sql-formatter-org/sql-formatter) (`^15.8.2`) as a dependency.
- New util `src/utils/sqlFormat.ts` exposing `formatSql(sql, dialect)`, which maps the active driver's dialect to the matching `sql-formatter` language:
  - `postgres → postgresql`, `mysql → mysql`, `mssql → transactsql`, `sqlite → sqlite`, `oracle → plsql`, everything else → `sql`.
  - Keywords are upper-cased; on empty input or a parse error the original text is returned unchanged so the user's SQL is never lost.
- `ViewEditorModal` gets a `Beautify` button next to the *View Definition* label, wired to the active connection's `sql_dialect`.
- Adds the `views.formatSql` i18n key to all 8 locales (en, es, it, fr, de, ja, ru, zh).

## Testing

- `pnpm typecheck` and `pnpm build` pass; ESLint clean on changed files.
- Manually verified against a MySQL 8 test database: opened a view stored as a one-line definition, clicked **Beautify**, and the `SELECT … JOIN … WHERE … GROUP BY … HAVING … ORDER BY` reformatted with proper indentation and keyword casing.

sql pretty


danielnuld:
Thanks for the review. Both warnings addressed in bcd169b:

- **`ViewEditorModal.tsx` (stale debounced state):** `handleFormat` now reads the editor's live value through an `onMount` editor ref instead of the debounced `definition` state, so edits typed within the 300ms window are no longer discarded.
- **`sqlFormat.ts` (missing tests):** added `tests/utils/sqlFormat.test.ts` (13 cases) covering each dialect mapping, keyword casing, empty/whitespace passthrough, and the parse-error fallback. `toFormatterLanguage` is now exported so the mapping is asserted directly.

debba:
@danielnuld I really love it!

I'd actually extend this to the Query Editor as well. If you're interested, feel free to open a separate PR for that.

About It I was also thinking it might make sense to have a single shared component used across the different sections, containing the query textarea, the AI controls (which aren't available in the View Editor yet), and the Beautify action.

In the meantime, I'll merge this one. Thanks!

PR #373, opened 2026-06-25
fix: preserve composite indexes in generated SQL

### LINKED ISSUE #349: Generated SQL splits composite indexes into multiple single-column indexes
## Summary

When using the database tool to inspect a table, composite indexes are displayed correctly in the UI. However, when right-clicking the index and selecting "Generate SQL", the generated statement is split into multiple single-column `CREATE INDEX` statements instead of one composite index definition.

## Environment

- Database: MySQL 8.4.10

## Example

Assume a table has this composite index:

```sql
CREATE INDEX `idx_orders_lookup`
ON `orders` (
  `customer_id`,
  `status`,
  `created_at`,
  `id`
);
```

The index appears correctly in the database/index view as one multi-column index.

But "Generate SQL" produces something like:

```sql
CREATE INDEX `idx_orders_lookup` ON `orders` (`customer_id`);
CREATE INDEX `idx_orders_lookup` ON `orders` (`status`);
CREATE INDEX `idx_orders_lookup` ON `orders` (`created_at`);
CREATE INDEX `idx_orders_lookup` ON `orders` (`id`);
```

## Expected behavior

The generated SQL should preserve the composite index as a single `CREATE INDEX` statement:

```sql
CREATE INDEX `idx_orders_lookup`
ON `orders` (
  `customer_id`,
  `status`,
  `created_at`,
  `id`
);
```

## Actual behavior

The generated SQL creates one `CREATE INDEX` statement per indexed column.

This changes the meaning of the index. Four single-column indexes are not equivalent to one composite index, especially for queries that rely on the left-prefix order of the composite index.

It can also generate invalid SQL on databases that require index names to be unique per table, because each generated statement reuses the same index name.

## Why this matters

A composite index such as:

```sql
(customer_id, status, created_at, id)
```

can support queries like:

```sql
WHERE customer_id = ?
  AND status = ?
  AND created_at >= ?
ORDER BY id
```

Splitting it into separate single-column indexes changes query planning and can cause significantly worse performance.

## Suggested fix

When generating SQL for an index, group all columns that belong to the same index definition and emit one `CREATE INDEX` statement with the columns in their original index order.

snvtac:
This PR preserves composite indexes when generating table SQL from index metadata.

### Changes
- Added `seq_in_index` to the frontend `Index` metadata type.
- Grouped index metadata rows by index name before rendering `CREATE INDEX` statements.
- Ordered grouped columns by `seq_in_index`, falling back to input order when sequence data is unavailable.
- Added regression coverage for composite non-unique indexes, composite unique indexes, sequence ordering, and table-level generated SQL output.

### Validation
- `pnpm exec vitest run tests/utils/sqlGenerator.test.ts` — 64 passed
- `pnpm typecheck`
- `pnpm lint` — passes with one existing `SettingsProvider.tsx` hook dependency warning
- `pnpm build` — passes with an existing Vite dynamic/static import chunking warning

Closes #349


michabbb:
thanks a lot ! 🙏 👍

PR #377, opened 2026-06-25
fix(postgres): support routine introspection on PostgreSQL < 11

### LINKED ISSUE #375: [Bug]: Schema introspection fails on PostgreSQL < 11 (column "prokind" does not exist)
## Description

Connecting to a PostgreSQL 9.6 server fails because the schema/routine
introspection query references `pg_proc.prokind`. That column was introduced
in **PostgreSQL 11**; on 9.x and 10 it does not exist, so the query errors out
with SQLSTATE `42703`.

The same incompatibility historically hit phpPgAdmin, jOOQ, EF Core Power Tools,
etc. when PG 11 changed `pg_proc`: the boolean columns `proisagg` and
`proiswindow` were collapsed into the single `prokind` char column.

## Environment

- Tabularis version: 0.13.3
- OS: macOS 26.5.1 
- PostgreSQL server: 9.6

## Actual behavior

```
ERROR: column "prokind" does not exist
SQLSTATE: 42703
HINT: Perhaps you meant to reference the column "pg_proc.probin".
file: parse_relation.c, routine: errorMissingColumn
```

## Expected behavior

The connection opens and the schema tree loads on PostgreSQL 9.6 and 10,
the same way it does on 11+.

## Root cause & suggested fix

Make the routine-introspection query version-aware. Branch on the server
version (e.g. `SHOW server_version_num` / `current_setting('server_version_num')::int`):

- **PG >= 110000** → keep using `prokind`.
- **PG < 110000** → derive an equivalent value from the old columns, e.g.:

```sql
  CASE
    WHEN p.proisagg    THEN 'a'   -- aggregate
    WHEN p.proiswindow THEN 'w'   -- window
    ELSE 'f'                      -- normal function
  END AS prokind
```

  (Procedures — `prokind = 'p'` — don't exist before PG 11, so `f`/`a`/`w`
  fully cover the pre-11 cases.)

Same treatment is needed for any other catalog query that assumes `prokind`.

### To Reproduce

1. Add a connection to a PostgreSQL 9.6 server.
2. Connect (the failure happens while building the object tree / listing
   stored routines).

### OS Version

macos 26.5.1 

### Tabularis Version

v0.13.3

### Relevant Log Output

```shell

```

earmellin:
pg_proc.prokind was added in PG 11, replacing proisagg/proiswindow.
Referencing it fails with SQLSTATE 42703 on 9.x and 10. get_routines now
picks a version-appropriate query based on server_version_num.

Tested against PostgreSQL 9.6.

Fixes #375


debba:
Hey @earmellin, thanks a lot for this — really nice catch on the prokind change, and even better that you tested it against 9.6.

If you feel like sending more PRs our way down the line, we'd genuinely love that. And if you want to hang out with the rest of us, come join our Discord: https://discord.com/invite/K2hmhfHRSt

Thanks again!

PR #378, opened 2026-06-25
fix(postgres): honor ssl_mode in build_connection_url

darkrideroffate:
PostgreSQL's `build_connection_url` never set `sslmode`, so the sqlx path (`test_connection` → `build_connection_url` → `AnyConnection::connect`) defaulted to `prefer` and attempted TLS even when the user picked **Disable**. The pool path already honors `ssl_mode`, so Load Databases worked while opening the connection failed against servers that reject the negotiated TLS (e.g. CloudNativePG forcing TLS 1.3): `bad protocol version`.

Now maps `ssl_mode` onto the `sslmode` URL param, mirroring the MySQL driver, plus tests for the `disable` and default cases.


src-tauri/src/drivers/postgres/mod.rs
```diff
@@ -1412,13 +1412,23 @@ impl DatabaseDriver for PostgresDriver {
         use urlencoding::encode;
         let user = encode(params.username.as_deref().unwrap_or_default());
         let pass = encode(params.password.as_deref().unwrap_or_default());
+        let sslmode = match params.ssl_mode.as_deref() {
```
NewtTheWolf, line 1753: **Non-blocking follow-up:** `verify-ca`/`verify-full` still won't actually work on this path — and now they *fail* instead of silently skipping verification. The default `test_connection` (driver_trait.rs:250) connects purely from this URL, which carries no `sslrootcert`, so `params.ssl_ca` (the custom CA that RDS / CloudNativePG users supply) is never propagated. Before this PR those modes silently ran as `prefer` (TLS, no verification); now they emit `sslmode=verify-full` against the system trust store only and will fail for any server needing a custom CA.

Arguably a safer fail-closed, but it means the verify-* modes stay broken on the sqlx test path, just differently. Options: append `&sslrootcert=<ssl_ca>` when a CA is configured, or leave it and note the limitation. The tests here cover only `disable`/default, so this path isn't exercised either way.

(Minor, and correct: the `_ => "prefer"` fallback matches `build_postgres_configurations`' default — nice consistency.)

NewtTheWolf (commented):
Good, well-targeted fix. The root cause is correctly identified: `build_connection_url` never emitted `sslmode`, so the default `test_connection` path (`driver_trait.rs:250` → `AnyConnection::connect`) fell back to sqlx's `prefer` and negotiated TLS even under **Disable** — diverging from the pool path, which already honored `ssl_mode`. Mapping the value onto the URL param fixes that, mirrors the MySQL driver, and the `disable` + default tests lock in the behavior. 👍

One non-blocking gap left inline for `verify-ca`/`verify-full`. Nothing blocking — it strictly improves the reported case.

NewtTheWolf (dismissed):
Confirmed the reported bug is fixed: `build_connection_url` now appends `?sslmode={}` mapping ssl_mode → disable/allow/prefer/require/verify-ca/verify-full (default prefer), so the sqlx `test_connection` path no longer diverges from the pool path under Disable. Tests lock in the disabled + default-prefer cases.

Follow-up (non-blocking, pre-existing): verify-ca/verify-full emit the mode but `params.ssl_ca` is still not propagated as `sslrootcert` on this URL path, so those modes rely on a system-trusted CA. Worth a follow-up issue. LGTM.

NewtTheWolf (approved):
Re-approving after merging current main into the branch to resolve the conflict with #450 (both added a new test module at the same spot; kept both — build_connection_url_tests + pg_vector_binding_tests). Ran the postgres suite locally: 101 passed. Fix itself unchanged. sslrootcert for verify-ca/verify-full tracked as a non-blocking follow-up.

PR #380, opened 2026-06-25
Enh/nested groups slash create

p4pupro:
## What

The only way to create a subgroup was through the right-click context menu on a group header. The menu was easy to miss, and the new-group toolbar had no path semantics — you could only create top-level groups and you had to drill down to nest them. This PR fixes both discoverability gaps.

1. **Visible `+` button on every group header.** Hovering a group now reveals an amber plus button (before the `MoreVertical` menu trigger). Clicking it focuses an inline input rendered between the header and the group's children, with `Enter` to confirm and `Escape` to cancel. Blur also commits.

2. **`/` as a path separator in both the toolbar "New group" input and the inline "New subfolder" input.** Entering `TEST/flexways`:
   - reuses an existing `TEST` group case-insensitively if found;
   - creates the missing segments under it (here `flexways`);
   - returns the deepest group so the UI stays in sync.

   `TEST/flexways/qa` creates all three levels in one call. Sibling lookups are scoped to the current parent only, so `/a/b` and `/x/b` resolve to different `b` groups.

## Backend

- New Tauri command `create_group_path(path, parent_id?)` that walks the path segment-by-segment, calling the existing `create_connection_group` for any segment that does not already exist under the resolved parent. It reuses the same per-parent `sort_order` logic, so siblings land at the end of their own chain.
- Helpers:
  - `parse_group_path` — splits on `/`, trims whitespace, drops empty segments, rejects an empty result.
  - `find_child_group` — case-insensitive, parent-scoped lookup.
- 6 new unit tests in `group_tree_tests` covering the parser (split on `/`, trim + drop empty, reject empty, keep single segment) and the lookup (case-insensitive, parent-scoped).

## Frontend

- `GroupHeader` gains an optional `onCreateSubgroup` prop and renders the new `Plus` button before the `MoreVertical` menu trigger.
- `DatabaseContext` + `DatabaseProvider` expose `createGroupPath` which invokes the new command and re-fetches `get_connection_groups` so the UI reflects reused + newly created segments in one render.
- `Connections.tsx` wires the inline subgroup input and replaces the toolbar "New group" handler with `createGroupPath` so the same path semantics apply at the top level.

## i18n

- New key `groups.subgroupNamePrompt` in `en, de, fr, ja, ru, zh`. The `es` and `it` locales were already missing the `groups` namespace after #1, so they keep falling back to the `defaultValue`.

## Testing

- `cargo test --lib`: **704/704** pass (the 6 new tests included).
- `pnpm tsc --noEmit`: clean.
- `pnpm vitest run`: 2616/2616.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -3379,23 +3379,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "native-tls"
-version = "0.2.18"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "465500e14ea162429d264d44189adc38b199b62b1c21eea9f69e4b73cb03bbf2"
-dependencies = [
- "libc",
- "log",
- "openssl",
- "openssl-probe",
- "openssl-sys",
- "schannel",
- "security-framework 3.7.0",
- "security-framework-sys",
- "tempfile",
-]
-
-[[package]]
 name = "ndk"
 version = "0.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5895,10 +5878,10 @@ dependencies = [
  "indexmap 2.13.0",
  "log",
  "memchr",
- "native-tls",
  "once_cell",
  "percent-encoding",
  "rust_decimal",
+ "rustls",
  "serde",
  "serde_json",
  "sha2",
@@ -5909,6 +5892,7 @@ dependencies = [
  "tracing",
  "url",
  "uuid",
+ "webpki-roots 0.26.11",
 ]
 
 [[package]]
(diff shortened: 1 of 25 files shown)


debba:
Hi @p4pupro ,
I see you closed the PR.
Was it done by mistake? We can restore it.

p4pupro:
Hi @debba,

Sorry I miss click, but my approach is test in my own fork
 and if this is working fine without issues, then I will sent de PR.

Thank you!

debba:
Ofc no problem, I asked you because we changed some policies in PRs and I also wanted to be sure that was not automatically closed.
Btw always open for new PRs :)

p4pupro:
OK, thanks. I’ll soon be submitting not one but two PRs: one relating to logging in with AWS RDS IAM, 
and another to enable the creation of a directory hierarchy within collection groups – currently, only a single level is permitted.

PR #394, opened 2026-06-29
fix(postgres): bind uuid-shaped PK as text for varchar columns

### LINKED ISSUE #392: [Bug]: Update failed: error serializing parameter 0
### Describe the bug

Trying to delete a record I am getting "Update failed: error serializing parameter 0" (pgsql)

Similar error when trying to edit a field of the table I get ; Update failed: error serializing parameter 1

PostgreSQL update context:
- table: "public"."objects"
- column: "status" (character varying)
- new value JSON type: string
- primary key: "guid" JSON type string
- SQL: UPDATE "public"."objects" SET "status" = $1 WHERE "guid" = $2

### To Reproduce

Just list table, delete a row then hit Submit

Pgsql table with varchar(36) primary key and jsonb column.

### OS Version

Win 11

### Tabularis Version

v0.13.3

### Relevant Log Output

```shell

```

NewtTheWolf:
Closes #392

## Problem
Editing or deleting a row in a PostgreSQL table whose primary key is `varchar`/`text` but holds a uuid-shaped value (e.g. `guid varchar(36)`) failed with **"error serializing parameter N"**.

## Cause
`build_pk_predicate` bound any uuid-shaped PK string as the native `Uuid` type, ignoring the column's declared type. tokio-postgres then rejected the `Uuid` param against the `varchar` column.

## Fix
- Thread the PK column's real type into `build_pk_predicate`; use the `Uuid` binding only for actual `uuid` columns. `varchar`/`text` PKs bind as text.
- Same gating applied to the bigint-string cast (avoids `varchar = bigint`).
- All four PK-filter call sites covered: `update_record`, `delete_record`, and both blob-fetch functions.
- Falls back to the previous shape heuristic when column metadata is unavailable.

## Tests
Two new unit tests assert the bound param type per column type (`uuid` vs `varchar`). `cargo test` 228 passed, clippy clean.


debba:
@NewtTheWolf merged `main` in to pick up the composite-PK rework from #324, which conflicted with this branch in the Postgres driver.

The two changes had to be combined rather than picked one over the other:

- `build_pk_map_predicate` now takes a `pk_types: &HashMap<String, String>` and threads each column's declared type into `build_pk_predicate`. That way the varchar/uuid fix from this PR applies to *every* member of a composite PK, not just a single-column one.
- Added a `get_pk_column_types` helper in `mod.rs` that resolves the type of each PK column, wired into the four call sites (update/delete record, blob save/fetch).
- Tests: kept both existing suites and added `composite_pk_threads_per_column_type_for_uuid_varchar`, which covers a composite key with one `uuid` column and one `varchar` column holding a uuid-shaped string — the uuid column binds as native `Uuid`, the varchar one as text, in the same predicate.

Full `drivers::postgres` suite is green (234 passed). Let me know if you'd prefer the type lookup batched into a single query instead of one per PK column.

PR #402, opened 2026-06-30
feat(postgres): browse multiple databases on one connection

### LINKED ISSUE #340: [Feat]: Browse/access all databases on same connection
### Prerequisites

### 🚀 Feature Proposal

Probably more noticeable for pgsql as the concept of database is different from mysql, but it would be quite useful imo to allow to leave the "database name" connection setting empty and that would indicate that the connection should be able to have access to all databases connecting user is allowed to.

Same/similar for mysql in terms of not having to explicitly have to select which dbs are to be available in the connection.

### Motivation

Make the product more standard

### Example

_No response_

debba:
Closes #340.

## What this does

You can now point a single PostgreSQL connection at several databases and browse them all from the sidebar, the same way MySQL and MariaDB connections already work. Pick the databases you want in the connection dialog's "Databases" tab and each one shows up as a top level node you can expand into its schemas and tables.

## Why Postgres needed special handling

MySQL lets one connection see every database, so the old multi database code treated a database name as if it were a schema and qualified table names with it. Postgres does not allow that: a connection is bound to one database and you cannot query across databases in a single session. So this PR keeps a separate connection pool per database and presents a real database to schema to table tree instead of the flat database to table layout MySQL uses.

The pool keying already included the database name, so most of the plumbing was about telling the backend which database a given read or query should run against.

## How it works

* Metadata reads (`get_tables`, `get_views`, `get_schemas`, `get_columns`, `get_indexes`, `get_foreign_keys`, the routine and trigger helpers) and `execute_query` / `execute_query_batch` now take an optional `database` argument. When set, the params get pointed at that database and the matching pool is used. This mirrors the convention the record mutation commands (`delete_record`, `update_record`, `insert_record`) already follow, so it is purely additive.
* Editor tabs carry the target database alongside the schema, so previews, queries and row edits route to the correct pool while still qualifying tables with the Postgres schema.
* `isMultiDatabaseCapable` no longer excludes schema based drivers; a new `isSchemaBasedMultiDb` helper distinguishes the hierarchical Postgres layout from the flat MySQL one.
* Empty database selections fall back to the `postgres` maintenance database, since Postgres cannot connect without a target. This is what makes "Load Databases" work before you have picked anything.

## UX change worth a look

Now that PostgreSQL counts as multi database capable, the connection dialog hides the single "database" field on the General tab and shows the "Databases" tab with a checklist instead, exactly like MySQL. Existing single database Postgres connections keep working, but the edit dialog looks different than before. Happy to keep the single field visible for Postgres if reviewers would rather not change that flow.

## Known limitations / follow ups

These are not regressions, just things left out of this pass to keep the change focused on browsing and querying (updated after the two routing-fix commits on this branch):

* **Object-creation DDL from the nested schema view is not database aware.** _Create Index_, _Create Foreign Key_ and _Create Table / View / Trigger_ launched from a nested Postgres schema node still route via the connection's *global* active schema and carry no `database`, so they can hit the primary database (and the wrong schema). Add/edit column, drop index/foreign key, and the view/routine/trigger context-menu actions (show data, view definition, edit, drop) **are** now routed to the node's database.
* **AI Query Generation is not database aware.** `AiQueryModal` builds its schema context from the connection's primary database and the generated-query flows (including Visual-Explain deep links from the AI activity panel) carry no `database`, so AI features target the primary database on multi-database Postgres connections.
* **Clipboard Import is not database aware.** The import creates its table on the connection's primary database regardless of which database node it was launched from.
* **SQL autocomplete is not database aware in multi-database mode.** `registerSqlAutocomplete` / `getTableColumns` take a `schema` but no `database`, so `get_columns` runs against the connection's primary pool — completions for tables in a non-primary database are missing or wrong. It still assumes the flat MySQL shape.
* **Per-database dumps cover one schema at a time.** `dump_database` routes to the selected database's pool and both the dump dialog and the ER diagram window now expose a schema dropdown (defaulting to the active schema, then `public`) to pick which schema to dump/diagram. Dumping *all* schemas of a database in one file is still not supported.
* **Sidebar highlight and auto-expand reuse `activeSchema`** for both the database node and the schema node, so the highlight can be slightly off. Navigation works fine.
* **No cap or idle eviction on the per-database pools yet.** Each selected database keeps its own pool (max 10 connections) for the lifetime of the connection. Disconnect (and health-check failure) now closes *all* of a connection's per-database pools, not just the primary one.

## Testing

* New unit tests for the `postgres_dbname` fallback helper and for the `isMultiDatabaseCapable` / `isSchemaBasedMultiDb` changes.
* Full suites pass locally: 753 Rust tests, 2705 frontend tests, plus `tsc` and ESLint clean.
* Still needs a manual pass against a real Postgres server with two or more databases selected.

---

### Sidebar — Tree & Object Types
- [x] Databases listed (multi-db, schema-based routing)
- [x] Schemas listed (system schemas excluded)
- [x] Tables / Views / Routines (Functions + Procedures split) / Triggers shown
- [x] Per-table: Columns, Keys (PK/Unique), Foreign Keys, Indexes expand correctly

### Sidebar — Connection / Database actions
- [ ] Import Database
- [ ] Dump Database
- [ ] View Schema Diagram (ER)
- [ ] New Console (opens editor on the correct database/schema)
- [ ] Refresh
- [ ] Filter/search: databases, schemas (multi-select), tables, triggers

### Sidebar — Schema-level actions
- [ ] Refresh tables
- [ ] Create Table (schema-aware)
- [ ] Create View
- [ ] Create Trigger (gated by `triggers` capability)

### Sidebar — Table context menu
- [ ] Show Data
- [ ] New Console
- [ ] Count Rows
- [ ] View Schema (structure)
- [ ] View ER Diagram
- [ ] Generate SQL (DDL)
- [ ] Clipboard Import
- [ ] Copy Name
- [ ] Add Column
- [ ] Delete Table

### Sidebar — Other object context menus
- [ ] View: Show Data / Count Rows / Edit View / Copy Name / Drop View
- [ ] Column: Modify / Copy Name / Delete (disabled on views)
- [ ] Index: Copy Name / Delete; folder → Add Index
- [ ] Foreign Key: folder → Add FK (cross-schema `ref_schema` qualification)
- [ ] Routine: View Definition (pg_get_functiondef) / Copy Name
- [ ] Trigger: View Definition (pg_get_triggerdef) / Edit / Copy Name / Drop

### Sidebar — DDL modals (schema-aware)
- [ ] CreateTable / ModifyColumn / CreateIndex / CreateForeignKey
- [ ] ViewEditor / TriggerEditor
- [ ] Dump / Import / Schema Inspector

### Query Editor — Execution
- [ ] Run query / Run selection (Ctrl+Enter)
- [ ] Run multiple statements (batch, shared session state)
- [ ] Multi-statement split dropdown
- [ ] Stop / cancel running query
- [ ] Parametric queries (`:param`) with input modal

### Query Editor — Editing experience
- [ ] SQL syntax highlighting (Monaco)
- [ ] Schema-aware autocomplete (tables/columns/keywords)
- [ ] Line numbers / word wrap / font settings
- [ ] Undo/redo, clipboard paste

### Query Editor — AI & Analysis
- [ ] AI Query Generation
- [ ] AI Explain
- [ ] Visual EXPLAIN / EXPLAIN ANALYZE (costs, rows, loops, buffer stats, JSON plan)

### Query Editor — Tabs & persistence
- [ ] New / Close / Switch tab, rename, close others/left/right
- [ ] Save Query
- [ ] Query History (status, duration, rows, DB) + re-run

### Results Grid — Display & navigation
- [ ] Pagination + Load Count
- [ ] Sort by column header
- [ ] Table toolbar: WHERE filter (SQL + visual builder AND/OR), ORDER BY, LIMIT
- [ ] Multi-row selection (shift/ctrl click, select all)

### Results Grid — Data editing (requires PK)
- [ ] Inline cell edit
- [ ] Row Editor sidebar
- [ ] Insert / Duplicate / Delete row (soft)
- [ ] Pending changes tracking
- [ ] Batch transactional commit (Ctrl+S) + revert
- [ ] Set NULL / Default / Empty / NOW()

### Results Grid — Postgres-specific types
- [ ] JSON / JSONB (dedicated editor)
- [ ] PostGIS geometry (WKB → WKT preview)
- [ ] Arrays (int[], text[], …)
- [ ] UUID
- [ ] BYTEA (hex/base64 + binary editor)
- [ ] INET/CIDR, range types, NUMERIC arbitrary precision

### Results Grid — Navigation & export
- [ ] FK navigation (jump to referenced table with filter) + related records panel
- [ ] Copy cell/row/column (CSV/JSON/SQL INSERT)
- [ ] Export CSV/JSON to file (progress + cancel)
- [ ] Multi-result tabs
- [ ] Detach / minimize / maximize results panel

### Known gaps — confirm intentionally out of scope for this PR
- [ ] Materialized Views (list/create/drop/refresh) — NOT implemented
- [ ] Sequences (first-class management) — NOT implemented
- [ ] Custom types / Enums / Domains — NOT implemented
- [ ] Extensions (PostGIS, hstore, …) — NOT listed/managed
- [ ] Check / Unique constraints (dedicated listing) — NOT implemented
- [ ] CREATE/DROP DATABASE, CREATE/DROP SCHEMA — NOT implemented
- [ ] TRUNCATE / RENAME table — NOT implemented
- [ ] Query cancellation cancels app task only, NOT DB-side (pg_cancel_backend)


NewtTheWolf:
Reviewed locally (tests pass: frontend 140, backend pool_manager 36). The architecture is solid — per-database pool keying (`postgres:conn:<id>:<db>`), the `postgres` maintenance-db fallback is contained to connection params (never reaches SQL), and the multi-db logic is capability-driven (no `driver === "postgres"` conditionals). Nice.

The one real issue is that `database` threading is **incomplete** — three call paths still route to the connection's primary DB:

**Blockers**
- `execute_query_batch` (`src-tauri/src/commands.rs:3172`) never got the `database` param (only `schema`), so batch/multi-statement execution runs against the primary DB. `execute_query` does it right at `commands.rs:3104` — mirror that.
- Export sends the **schema** as the database — `src/pages/Editor.tsx:2704` uses `activeTab?.schema` as `targetDatabase`. For Postgres (now multi-db-capable) that routes to the wrong pool. Use the `buildTableRoutingParams(...)` helper the PR already added (it's used at 575/1959/2132, just not here).
- `count_query` has no `database` — frontend `src/pages/Editor.tsx:1141` and backend `src/commands.rs:3321` both lack it, so pagination totals run against the primary DB → wrong count or error for non-primary databases.

**Nits**
- `delete_record`/`update_record`/`insert_record` (`commands.rs:2761`, `:2792`, `:3045`) use `if let Some(db) = database` without the `.filter(|d| !d.is_empty())` guard the 13 read commands use — with the new maintenance-db fallback, a `Some("")` write lands on the `postgres` database.
- `driver === "mysql"` added at `NewConnectionModal.tsx:1187` (pipes_as_concat) violates `.rules/frontend.md` no-driver-conditionals.
- The MySQL/SQLite `pk_col` → composite `pk_map` rework + added `escape_identifier` is a behavioural refactor riding along (`.rules/rust.md` #6). Good change on its own (closes an identifier-injection vector, unit-tested) but ideally its own PR.

Repro for the three blockers, on a connection with multiple Postgres DBs selected: run a multi-statement query in a non-primary DB (B1), paginate a large table in a non-primary DB (B3 count), export a non-primary-DB table to CSV (B2) — each hits the primary pool instead.

debba:
## Results Grid — multi-database routing review + fixes (86640146)

Verified every Results Grid checklist item at code level for Postgres multi-db routing. Most paths were already correct (pagination, sort, toolbar WHERE/ORDER BY/LIMIT, pending changes, FK navigation + related records panel, and the new typed bindings in `binding.rs`). Five routing gaps were found and fixed in 86640146:

### Fixes

1. **Commit (Ctrl+S) regression on plain single-database Postgres — critical** (`src/pages/Editor.tsx`)
   `isMultiDatabaseCapable` now includes Postgres, so the flat-driver fallback `{ database: activeTab.schema }` fired on ordinary Postgres connections and sent the **PostgreSQL schema name** (e.g. `public`) as the database — routing every update/insert/delete commit to a pool for a database literally named `public` (connection failure, or the wrong database if one with that name exists). The fallback is now gated with `!isSchemaBasedConn`.

2. **Export CSV/JSON** (`src/pages/Editor.tsx`)
   Same pattern: export passed `activeTab.schema ?? activeSchema` as `database`. On schema-based drivers it now routes via `activeTab.database` (falling back to the primary pool when absent); flat drivers (MySQL) unchanged.

3. **Load Count** (`src-tauri/src/commands.rs::count_query` + `Editor.tsx`)
   The backend command had no `database` parameter at all, so counts on a non-primary-DB table ran against the primary pool (relation-not-found, or a silently wrong count on a same-named schema/table). Added the param backend-side and the frontend now passes `tab.database`.

4. **Multi-statement scripts / multi-result tabs** (`src-tauri/src/commands.rs::execute_query_batch`)
   The frontend already sent `database`, but the command signature didn't declare it — Tauri silently dropped the argument and the whole batch ran on the primary pool. The param is now accepted and routed.

5. **BYTEA preview/download** (`commands.rs::save_blob_to_file` / `fetch_blob_as_data_url` + UI plumbing)
   Both blob commands were not db-aware and `BlobInput` used the connection-global active schema. Added `database` to both commands and plumbed the tab's schema/database through `DataGrid → RowEditorSidebar → FieldEditor → BlobInput` (the tab's schema now takes precedence over the global one).

### Verified as already correct
- Pagination, sort, toolbar filters → routed `execute_query` (`database` param)
- PK/column metadata → `buildTableRoutingParams` (`get_columns`/`get_foreign_keys`)
- FK navigation + related records → `ref_schema` + `tab.database` carried through
- Typed bindings (JSON/JSONB, UUID, NUMERIC, temporal, BYTEA, arrays-as-literals, WKT) — consistent Rust-type/wire-type pairs via `prepare_typed`

### Known non-fixes (pre-existing, out of scope)
- Ctrl+S commit is not transactional (independent `Promise.all` invokes — pre-existing design)
- Editing INET/CIDR/range columns still relies on a text→type assignment cast Postgres doesn't have; it failed before this PR too (client-side serialize error)
- `NewRowModal` is dead code (never opened) — left untouched

### Checks
`pnpm typecheck` clean · `vitest run` 2730 passed · `pnpm lint` 0 errors · `cargo check` clean · `cargo test` 763 passed (4 pre-existing `askpass` failures are environmental: `SUN_LEN` socket-path limit in the sandbox)

debba:
## Second round of multi-database routing fixes (02d2e77)

Follow-up to the results-grid fixes in 86640146 — this commit addresses the remaining routing gaps found during verification of the sidebar, tooling and lifecycle paths:

### Fixed

1. **Dump Database unusable on Postgres (blocker)** — `DumpDatabaseModal` read tables from `databaseDataMap[db].tables`, which is always empty for schema-based drivers (the provider fills `schemas`/`schemaDataMap` instead), so validation rejected every dump with "no tables". The dialog now fetches the table list via `get_tables` for the exact schema the backend will dump (the active schema, defaulting to `public`), routed to the target database's pool. MySQL flow unchanged.
2. **Visual EXPLAIN / EXPLAIN ANALYZE** — `explain_query_plan` had no `database` param; added it (standard `DatabaseSelection::Single` override) and the editor passes `activeTab.database` through `VisualExplainModal`. AI-activity deep links still carry no database (part of the AI limitation, see PR body).
3. **Sidebar context menus** — `SidebarSchemaItem` forwarded `database` to table items only. Now views, routines, triggers and columns carry their node's schema+database, and every action routes correctly: view Show Data / Count Rows / Edit / Drop (backend `drop_view`, `create_view`, `alter_view`, `get_view_definition`, `get_view_columns` gained `database`), routine View Definition + parameters, trigger View Definition / Edit / Drop (backend `create_trigger`/`drop_trigger` gained `database`), column Modify (ModifyColumnModal now receives the node's schema/database) and column Delete.
4. **Import Database** — `import_database` gained a `database` param; the modal scopes the import to the database node it was launched from (previously it always ran on the primary pool with the global active schema).
5. **ER Diagram from a database node** — the two database-node call sites now pass `database` on schema-based drivers, and `SchemaDiagramPage` no longer falls back to treating the database name as a schema when a `database` param is present. MySQL fallback behavior preserved.
6. **Saved queries / favorites / history re-run** — these passed the stored database into `runQuery`'s `schema` slot, which on Postgres would run `SET search_path TO "<db>"` on the primary pool. A `runSavedSql` helper now routes the value into the tab's `database` (pool key) on schema-based drivers; flat drivers unchanged. Save Query also defaults to `activeTab.database` instead of the pg schema.
7. **Pool leak on disconnect** — `close_pool_with_id` removed only the primary database's key, leaking every `driver:conn:{id}:{otherdb}` pool. It now sweeps all pools sharing the connection's key prefix (covers both `disconnect_connection` and the health-check failure path). Three new unit tests in `pool_manager_tests.rs` cover the sweep, the no-connection-id case, and the empty-map case.

### PR body updated
The "Known limitations / follow ups" section now reflects this work and adds the two remaining known gaps: **AI Query Generation** and **Clipboard Import** are not database aware.

### Skipped intentionally (declared out of scope in the PR)
Create Table/View/Trigger/Index/FK from nested nodes, autocomplete, sidebar highlight, transactional commit, `pg_cancel_backend`.

### Checks
`pnpm typecheck` clean · `vitest run` 2730 passed · `pnpm lint` 0 errors · `cargo check` clean · `cargo test` 766 passed (+3 new; the 4 `askpass` failures are the sandbox `SUN_LEN` socket-path issue, pre-existing)

debba:
## Schema picker for ER diagram and Dump (3540f99)

Follow-up on reviewer feedback: on PostgreSQL both the ER diagram and the dump were silently scoped to the connection's global active schema. Both surfaces now expose a changeable schema dropdown (schema-based drivers only — MySQL and single-database drivers are unchanged, no picker shown):

- **Dump dialog** — a schema `Select` lists the target database's schemas (`get_schemas` routed to that database's pool), defaults to the active schema (then `public`), and drives both the displayed table list and the `schema` sent to `dump_database`. Switching schema re-fetches and re-selects the tables.
- **ER diagram window** — the same dropdown sits in the window header and re-scopes the diagram to another schema of the target database without reopening it (title follows the picked schema). Since the window runs without the opener's capability context, `open_er_diagram_window` gained a `schema_based` flag that all five opener call sites pass when `capabilities.schemas === true`.

PR body's dump limitation ("scoped via the global active schema") is superseded by this commit for the picker flows.

Checks: `pnpm typecheck` clean · `vitest run` 2730 passed · `pnpm lint` 0 errors · `cargo check --tests` clean

PR #404, opened 2026-07-01
feat(mysql): support AWS RDS IAM authentication

p4pupro:
## Summary

Adds AWS RDS IAM database authentication for MySQL connections. The password
field is treated as a short-lived RDS auth token (e.g. from
`aws rds generate-db-auth-token`); TLS is mandatory and the connection opts in
to MySQL's `mysql_native_password` cleartext plugin so the token can be
exchanged under TLS.

Fixes the `1045 Access denied` users currently get when wiring a generated RDS
auth token into a normal Tabularis connection.

## What changed

- New `use_iam_auth` flag on the MySQL connection modal. The password field is
  fed straight into `MySqlConnectOptions` on every connect; keychain lookups
  are skipped because the token must come from the form (15-minute expiry).
- `enable_cleartext_plugin(true)` is set on the pool builder when IAM auth is
  active. RDS requires it for token exchange under TLS.
- SSL mode is auto-escalated to `VerifyCa` when a CA bundle is configured and
  the user picked `Required` or `Preferred`, so the IAM/TLS invariant holds
  without manual setup.
- Pool key segments by `iam:{}` so an IAM-auth connection never shares a pool
  with a regular one.
- `test_connection` and `list_databases` fail fast with a clear error when IAM
  is enabled but the password is empty, and log a warning on failure so the
  logs distinguish "user typo" from "broken connection".
- `mcp/mod.rs` forwards the IAM flag into the connection-options builder.

## UI

New "Use AWS IAM Authentication (RDS)" checkbox in the MySQL connection
modal. When enabled, the password placeholder switches to the RDS-token hint
and an inline message surfaces if SSL is off.

AWS IAM auth option in the MySQL connection modal

## Tests

Twelve new unit tests in `pool_manager_tests.rs` covering:

- IAM flag changes the pool key.
- IAM + disabled SSL is rejected.
- IAM + escalation still keeps the cleartext plugin on.
- IAM combinations across all SSL modes.
- Test/duplicate/list flows skip the keychain for IAM connections.
- Empty-password IAM returns a clear error.

`cargo build --release`, `cargo test --lib` (761 passing; the four askpass
failures are pre-existing on `main`), and `pnpm tsc --noEmit` all pass.

## i18n

Added `useIamAuth`, `useIamAuthHint`, and `useIamAuthTlsRequired` across all
eight locales (en, es, de, fr, it, ja, ru, zh).


src-tauri/src/pool_manager.rs
```diff
@@ -170,29 +175,80 @@ pub(crate) fn build_mysql_options(
...
+            return Err(
+                "AWS IAM authentication requires a TLS/SSL mode to be enabled \
+                 (Preferred, Required, Verify CA, or Verify Identity). Refusing \
+                 to send the RDS auth token over an unencrypted connection."
+                    .to_string(),
+            );
+        }
+        // Saved connections get the token injected from the keychain after
+        // this builder returns, so an empty password is fine for them.
+        if password.is_empty() && params.connection_id.is_none() {
```
NewtTheWolf: **🔴 Blocking — this re-introduces the exact `Access denied` error the PR set out to fix, on a saved-connection path the new guards don't cover.**

The empty-token guard here only fires when `connection_id.is_none()`, and the comment just above justifies it with *"Saved connections get the token injected from the keychain after this builder returns."* That justification is no longer true: `find_connection_by_id` (commands.rs:418) and `duplicate_connection` (936) now deliberately **skip** the keychain for IAM connections — that's the whole point of the 15-minute-token design.

So the token is never injected for a saved IAM connection, and the guard is bypassed precisely for those. Concrete repro:

1. Save a MySQL connection with `use_iam_auth = true`, `save_in_keychain = true` → on-disk password is stripped to `None`.
2. Restart the app (or let the token expire so the pool is evicted), then browse tables / run a query.
3. `get_tables` / `execute_query` → `find_connection_by_id` (skips keychain, `password` stays `None`) → `resolve_connection_params_with_id` (sets `connection_id = Some`) → `get_mysql_pool_for_database` → **here**: `use_iam_auth = true`, `password` empty, `connection_id.is_some()` → this guard is skipped, `.password()` is skipped, `enable_cleartext_plugin(true)` is forced → connects with **no password** → server returns `Access denied (using password: NO)`.

`test_connection` and `list_databases` have their own empty-token guard, but the query/connect commands do not — so the bug only survives on exactly the path users hit most.

**Fix:** enforce the empty-token check in `build_mysql_options` regardless of `connection_id` — it's the single choke point every connect routes through — and delete the now-false comment.
```rust
if params.use_iam_auth.unwrap_or(false) && password.is_empty() {
    return Err("AWS IAM authentication is enabled but no RDS auth token was provided. \
        Paste the output of `aws rds generate-db-auth-token` into the password field. \
        Tokens expire every 15 minutes.".to_string());
}
```

src-tauri/src/pool_manager.rs
```diff
@@ -170,29 +175,80 @@ pub(crate) fn build_mysql_options(
...
+        && matches!(ssl_mode, MySqlSslMode::Required | MySqlSslMode::Preferred)
+    {
+        ssl_mode = MySqlSslMode::VerifyCa;
+    }
+
+    // AWS RDS IAM auth: `password` carries the pre-signed RDS auth token
+    // (from `aws rds generate-db-auth-token`), sent cleartext via
+    // mysql_clear_password over TLS. Refuse to send it unencrypted.
+    if params.use_iam_auth.unwrap_or(false) {
+        if matches!(ssl_mode, MySqlSslMode::Disabled) {
```
NewtTheWolf, line 213: **🔴 Security — the RDS auth token can travel in cleartext.**

This guard only rejects `MySqlSslMode::Disabled`, and its error message even advertises `Preferred` as an acceptable mode. But `Preferred` is *opportunistic* TLS: sqlx tries TLS and silently falls back to an **unencrypted** connection if the handshake doesn't complete. Meanwhile `enable_cleartext_plugin(true)` is set unconditionally for IAM (line 295), so under `mysql_clear_password` the token is sent to the server in cleartext.

Result: with `ssl_mode = preferred` and no CA (so the `has_user_ca → VerifyCa` escalation at line 191 doesn't fire, and Preferred stays Preferred), a network attacker who blocks/strips the TLS upgrade gets the pre-signed RDS token on the wire in plaintext. The line-295 comment *"Safe because the token only travels over the TLS link above"* is false in this mode, and it defeats the stated invariant (*"Refusing to send the RDS auth token over an unencrypted connection"*).

**Fix:** require an actively-enforced TLS mode for IAM — reject `Preferred` alongside `Disabled` (accept only `Required`/`VerifyCa`/`VerifyIdentity`), or force-upgrade `Preferred → Required` whenever `use_iam_auth` is on. Update the error string to match.

NewtTheWolf (commented):
Nicely structured — serde/persistence round-trip and i18n are clean, and the SSL-escalation reasoning is sound. Two blocking issues though (one re-introduces the opaque `Access denied` this PR set out to fix, on a path the guards don't cover), plus a security gap and an unreachable UI. Details inline; ranked by severity.

src/components/modals/NewConnectionModal.tsx
```diff
@@ -1564,6 +1566,45 @@ export const NewConnectionModal = ({
               </button>
             </div>
           </div>
+
+          {driver === "mysql" && (
+            <div className="pt-2 border-t border-strong/50">
+              <label className="flex items-start gap-2 cursor-pointer">
+                <input
+                  type="checkbox"
+                  checked={!!formData.use_iam_auth}
```
NewtTheWolf: **🟠 The IAM checkbox is effectively unreachable when creating a fresh MySQL connection.**

This block lives inside the SSL Certificate Files section, which only renders when `formData.ssl_mode` is truthy and not `disable`/`disabled`. A brand-new connection has `formData.ssl_mode === ""` — the SSL dropdown *displays* "Required" through a `formData.ssl_mode || "required"` fallback, but the underlying state is still empty. So the cert block, and this checkbox with it, stays hidden until the user happens to re-select an SSL mode from the dropdown. For the feature's primary audience (RDS users setting up IAM auth) it looks like the option simply isn't there.

Second failure mode: enable IAM under `Required`, then switch SSL to `Disabled` → the block unmounts while `formData.use_iam_auth` stays `true`. The user can no longer untick it, and the connection saves with `use_iam_auth = true` + SSL off — a combination `build_mysql_options` then rejects at connect time.

**Fix:** gate the IAM checkbox on `driver === "mysql"` (and the *effective* ssl mode, i.e. the same `|| "required"` fallback used for display), independent of whether an explicit `ssl_mode` string has been chosen; and clear `use_iam_auth` when SSL is set to Disabled.

src/components/modals/NewConnectionModal.tsx
```diff
@@ -1564,6 +1566,45 @@ export const NewConnectionModal = ({
...
+                      defaultValue: "Use AWS IAM Authentication (RDS)",
+                    })}
+                  </div>
+                  <div className="text-xs text-muted mt-0.5">
+                    {t("newConnection.useIamAuthHint", {
+                      defaultValue:
+                        "The password field is treated as an RDS auth token (from `aws rds generate-db-auth-token`). Requires TLS. Tokens expire every 15 minutes.",
+                    })}
+                  </div>
+                  {formData.use_iam_auth &&
```
NewtTheWolf: **🟠 Dead code — this warning can never render.**

The warning is meant to alert IAM users when SSL is off (mirroring the backend's rejection of IAM + Disabled). But its condition (`ssl_mode` is `disabled`/`disable`/empty) is fully mutually exclusive with the enclosing block's guard (`ssl_mode && ssl_mode !== "disable" && ssl_mode !== "disabled"`). In every state where this condition is true, the parent block is unmounted — so the one situation the warning exists to catch (IAM enabled, SSL disabled) is exactly the situation where it's invisible.

This is the flip side of the previous comment: once the checkbox is moved out from under the SSL-cert block, this warning will start being reachable and should be re-tested against the actual off/downgraded SSL modes.

src-tauri/src/pool_manager_tests.rs
```diff
@@ -287,6 +301,174 @@ mod tests {
...
+        let params = mysql_params_with_ca(
+            "verify_identity",
+            "/Users/dperez/.ssh/rds-combined-ca-bundle.pem",
+        );
+        let opts = build_mysql_options(&params, None).expect("must build");
+        assert!(matches!(opts.get_ssl_mode(), MySqlSslMode::VerifyIdentity));
+    }
+
+    #[test]
+    fn mysql_options_iam_auth_combined_with_escalation_keeps_cleartext_plugin() {
```
NewtTheWolf, line 519: **🟡 This regression test doesn't test the regression it's named for.**

`mysql_options_iam_auth_combined_with_escalation_keeps_cleartext_plugin` asserts only `ssl_mode == VerifyCa` — it never checks that the cleartext plugin is still enabled after escalation. That's the exact behavior the guard was added for: if a refactor let SSL escalation drop the `enable_cleartext_plugin` opt-in, this test would still pass green.

The sibling `mysql_options_iam_auth_passes_password_through_under_tls` already shows the assertion is easy:
```rust
assert!(format!("{options:?}").contains("enable_cleartext_plugin: true"));
```
Add that here so the test actually covers escalation-keeps-plugin.

src-tauri/src/commands.rs
```diff
@@ -1805,7 +1811,29 @@ pub async fn test_connection<R: Runtime>(
     let mut expanded_params = expand_ssh_connection_params(&app, &request.params).await?;
     expanded_params = expand_k8s_connection_params(&app, &expanded_params).await?;
 
-    if request.params.password.is_none() && expanded_params.password.is_none() {
+    // AWS RDS IAM auth tokens are short-lived (15 min) and must come from the
+    // password field on every test/connect, never from the keychain. Skip the
+    // keychain fallback so a stale token can't be reused.
+    let iam_auth = expanded_params.use_iam_auth.unwrap_or(false);
```
NewtTheWolf, line 2015: **🟡 The commands-layer IAM behavior has zero test coverage — and it contradicts the one branch that *is* tested.**

Everything the product actually depends on for IAM correctness is untested: the empty-token rejection here and in `list_databases`, the keychain-skip in `find_connection_by_id`/`duplicate_connection`, and the MCP `password = None` override.

Worse, there's a direct contradiction. `build_mysql_options` **allows** an empty password for a saved IAM connection, and that permissive branch is locked in by a test (`mysql_options_iam_auth_allows_empty_password_when_connection_id_set`). This command layer **rejects** the same empty-password saved-IAM case. Only the permissive side has coverage, so a regression that re-enabled the keychain fallback for IAM (removing the reason the strict guard exists) would sail through CI.

Add command-level tests for the empty-token rejection and the keychain-skip; and reconcile the two layers so they agree on the saved-connection empty-token case (see the blocking comment on `build_mysql_options`).

src-tauri/src/commands.rs
```diff
@@ -1805,7 +1811,29 @@ pub async fn test_connection<R: Runtime>(
...
+
+    // IAM auth needs an RDS auth token right now. Without this guard the
+    // builder accepts an empty password (saved connections inject later),
+    // the server replies with the opaque "Access denied (using password:
+    // YES)", and the user can't tell whether the token is missing, wrong,
+    // or expired.
+    if iam_auth
```
NewtTheWolf: **🟡 Cleanup — duplicated guard.**

This `iam_auth` compute + the double `password.as_deref().unwrap_or("").is_empty()` check + the full multi-line user-facing error string are copy-pasted verbatim into `list_databases` (line 2714). A wording tweak or logic change (e.g. later honoring a saved token) has to be made in both places and will drift.

Extract a single helper and call it from both:
```rust
fn require_iam_token(iam_auth: bool, req_pw: Option<&str>, expanded_pw: Option<&str>) -> Result<(), String> {
    if iam_auth && req_pw.unwrap_or("").is_empty() && expanded_pw.unwrap_or("").is_empty() {
        return Err("AWS IAM authentication is enabled but the password field is empty. \
            Paste the output of `aws rds generate-db-auth-token` into the password field \
            and try again. Tokens expire every 15 minutes.".to_string());
    }
    Ok(())
}
```

src-tauri/src/pool_manager.rs
p4pupro: empty-token guard now fires unconditionally in build_mysql_options, 
and the permissive test is inverted to pin the strict behaviour. 
The "Saved connections get the token injected from the keychain" comment is gone; 
that contract was stale because IAM connections skip the keychain upstream.

src-tauri/src/pool_manager.rs
p4pupro: Preferred is force-upgraded to Required under IAM
   (was opportunistic TLS + forced cleartext plugin = token over the
   wire in the clear under a stripped STARTTLS). Disabled still
   rejected; the error string no longer lists Preferred as acceptable.

src/components/modals/NewConnectionModal.tsx
p4pupro: IAM checkbox moved out of the SSL Certificate Files
   block, gated on the effective SSL mode. Switching SSL to Disabled
   now clears use_iam_auth, so the "block unmounts while flag is true"
   failure mode is also closed.

src/components/modals/NewConnectionModal.tsx
p4pupro: The TLS-required warning is now reachable
   because the new gate uses the effective SSL mode.

src-tauri/src/pool_manager_tests.rs
p4pupro: the test now actually asserts `enable_cleartext_plugin: true` after escalation.

src-tauri/src/commands.rs
p4pupro: require_iam_token helper extracted,
   called from both sites, with 5 unit tests covering the empty-token
   guard at the command layer. The command-layer / builder
   contradiction is resolved by (first commit). Full keychain-skip coverage for
   find_connection_by_id / duplicate_connection needs a Tauri
   MockRuntime harness and is best landed as a follow-up rather than
   via ad-hoc mocks.

src-tauri/src/commands.rs
p4pupro: same as (last comment).

src-tauri/Cargo.toml
```diff
@@ -28,7 +28,7 @@ serde = { version = "1.0", features = ["derive"] }
 log = "0.4"
 tauri = { version = "2.10.2", features = ["protocol-asset", "devtools"] }
 tauri-plugin-log = "2"
-sqlx = { version = "0.8.6", features = ["runtime-tokio", "sqlite", "mysql", "postgres", "tls-native-tls", "chrono", "uuid", "rust_decimal", "json"] }
+sqlx = { version = "0.8.6", features = ["runtime-tokio", "sqlite", "mysql", "postgres", "tls-rustls", "chrono", "uuid", "rust_decimal", "json"] }
```
NewtTheWolf: **Blast radius: this swaps the TLS backend for *every* MySQL connection, not just RDS.** rustls differs from native-tls in ways that can silently break existing users:

- **No OS trust store** — native-tls used the system cert store; sqlx+rustls uses bundled webpki roots. A corporate MySQL server whose CA is trusted only via an OS-installed root (no explicit `ssl_ca`) under `VerifyIdentity`/`VerifyCa` will now fail to connect.
- **TLS 1.0/1.1 dropped** (only `tls12` + 1.3 enabled) — older servers/proxies break.

The macOS-EKU motivation is real and aligning with the Postgres rustls path is reasonable long-term, but this warrants explicit non-RDS regression testing (VerifyIdentity against an OS-trusted-CA server; an older TLS server) and a release note — it's a big change to be implicit inside an IAM PR. Related: the escalation tests in `pool_manager_tests.rs:744` still describe *native-tls* CA-forwarding behavior, so that rationale is now stale.

src-tauri/src/pool_manager.rs
```diff
@@ -170,29 +175,90 @@ pub(crate) fn build_mysql_options(
...
+        "preferred" | "prefer" => MySqlSslMode::Preferred,
+        "required" | "require" => MySqlSslMode::Required,
+        "verify_ca" => MySqlSslMode::VerifyCa,
+        "verify_identity" => MySqlSslMode::VerifyIdentity,
+        _ => MySqlSslMode::Required,
+    };
+    if has_user_ca
+        && matches!(ssl_mode, MySqlSslMode::Required | MySqlSslMode::Preferred)
+    {
+        ssl_mode = MySqlSslMode::VerifyCa;
```
NewtTheWolf: The `Required|Preferred + ssl_ca → VerifyCa` escalation isn't gated on IAM (the tests confirm it's intended for all connections). So existing **non-IAM** MySQL connections with a CA set + `Required` silently gain chain validation; a user whose bundle is incomplete/mismatched loses a previously-working connection.

For IAM the only hard requirement is an *encrypted* link, which `Required` already provides — so gating this on `use_iam_auth` would keep the RDS fix while removing the regression surface for everyone else. If it's meant to be global, worth a release note.

src/components/modals/NewConnectionModal.tsx
```diff
@@ -1433,9 +1435,12 @@ export const NewConnectionModal = ({
           }
           onChange={(v) => {
             updateField("ssl_mode", v);
-            // Cleartext auth must never go over an unencrypted link.
-            if (driver === "mysql" && v === "disabled") {
+            // Cleartext auth and RDS IAM must never go over an unencrypted
+            // link: both rely on a TLS-guaranteed channel to protect the
+            // password / pre-signed token.
+            if (driver === "mysql" && (v === "disabled" || v === "disable")) {
```
NewtTheWolf: UI-state divergence: this clears `use_iam_auth` only for `disabled`, but the checkbox below is `checked={!tlsOff && !!formData.use_iam_auth}` + `disabled`, and `tlsOff` also covers `preferred`. So: enable IAM under Required, then switch SSL to Preferred → the checkbox renders unchecked while `formData.use_iam_auth` stays `true`, and the connection saves with IAM on despite the UI showing it off.

Secure (the backend force-upgrades Preferred→Required), but the UI misrepresents the persisted value. Clear `use_iam_auth` whenever the effective mode is TLS-off, not just on `disabled` — e.g. reuse the same `tlsOff` predicate here.

NewtTheWolf (commented):
Really solid, carefully-reasoned PR — the IAM feature itself is well-built: the TLS invariant is enforced end-to-end (Disabled→reject, Preferred→Required, cleartext plugin only under TLS), the keychain skip is thorough across load/duplicate/MCP, you log `password_len` rather than the token, and the pool-key isolation is right. 🙏 12 tests to boot.

No blocking bugs from me — but two **app-wide** changes are riding along with the IAM feature that I'd want verified/documented before merge (inline), plus one small UI-state bug. Marking as comment rather than request-changes.

✅ Nice touches: `ensure_rustls_crypto_provider()` added ahead of the mysql connect avoids the rustls first-handshake panic; the pool key segmenting `iam:` keeps IAM and password pools apart.

p4pupro:
Thanks for the careful read, and for keeping it as a comment rather than request-changes 🙏

Walking back through the three app-wide concerns against what's actually in commit id:  300bdc20

1. Blast radius ([tls-rustls] swap affects every MySQL connection, not just RDS)

I addressed this in two ways:

Trust store. Picked tls-rustls-ring-native-roots (not [tls-rustls] plain), so the macOS Keychain / Windows Schannel / Linux openssl-probe roots are still loaded. Corporate MySQL servers whose CA is trusted only via an OS-installed root keep working under [VerifyIdentity]/[VerifyCa] without an explicit [ssl_ca]. The only loss is webpki-roots-only users, and webpki-roots was never on the path before (we were on native-tls) — so this is a strict superset, not a behavioural change for non-RDS users.
Protocol version. rustls is built with tls12 + 1.3, same as the native-tls path. The only sessions that break are TLS 1.0/1.1, which have been formally deprecated by IETF (RFC 8996) and removed from every maintained TLS stack; native-tls on macOS already stopped negotiating them. No regression for any server that worked before.
What I did not do: add a non-RDS VerifyIdentity regression test against an OS-trusted CA. The honest answer is I don't have a non-RDS MySQL server in my dev loop to test against, and the point of [ring-native-roots] is precisely to delegate trust to whatever the OS gives us. Happy to add one if there's a public test endpoint you'd suggest, but the more useful test is "no regression in mysql_options_* for non-IAM, no-CA paths" — those are now covered by [mysql_options_does_not_escalate_when_ca_absent], ..._when_ca_is_blank, ..._for_non_iam_with_ca, and ..._verify_identity_preserved_for_non_iam.

2. [Required|Preferred + ssl_ca → VerifyCa] is not gated on IAM

You're right that the escalation isn't IAM-gated, and I deliberately left it that way. The macOS EKU issue is not specific to IAM — it's macOS-native-tls failing on user-supplied CA bundles in general. A non-IAM user pointing at the same bundle would hit the same "One or more parameters passed to a function were not valid" opaque handshake. Gating the escalation on IAM would leave every non-IAM user with a CA on macOS broken after the rustls switch, which is a worse failure mode than "user with a bad bundle now sees a clear chain error instead of an opaque handshake".

The "supplying a CA file is itself an explicit opt-in to stricter validation" line in the test module comment is the rationale. Release note is on me — I left a long-form explanation in the commit body and a multi-line comment in [Cargo.toml], but the user-facing changelog is auto-generated by release-please from the commit message, which currently says [fix(mysql): switch sqlx TLS backend from native-tls to rustls] and the body. Let me know if you want a BREAKING trailer or a more explicit one-liner.

3. UI-state divergence: use_iam_auth only cleared on [disabled]

This one was already fixed in 2115fd0f ("fix(ui): clear use_iam_auth whenever the effective SSL mode is TLS-off") using the exact tlsOff predicate you suggested. The handler now does:

with tlsOff = !["required", "verify_ca", "verify_identity"].includes(effectiveSslMode) and [effectiveSslMode = formData.ssl_mode || "required"]. So switching to preferred (or empty + a forced change, or anything that ends up TLS-off) now also clears both flags, and the checkbox reflects the persisted value. The standalone 2115fd0f commit has the details and is already on the branch.

PR #405, opened 2026-07-01
feat(groups): support nested connection groups with cascade delete

p4pupro:
## Summary

Connection groups can now nest to N levels. New sub-folders are created from
a visible `+` button on each group header or by typing a `/`-separated path
in the toolbar ("New group" → `clients/eu/finance` creates the missing
segments). Deleting a group removes its entire subtree of sub-folders and
connections, and the export/import round-trip preserves the hierarchy.

## What changed

- `ConnectionGroup` already carried `parent_id`; this wires it up end-to-end
  in both the renderer and the `find_connection_by_id` /
  `get_connections_with_groups` flows so sub-folders render in list and grid
  view with a per-level indent.
- New Tauri command `create_group_path(path, parent_id?)` that walks the path
  segment by segment, reusing existing groups case-insensitively and
  creating the missing ones under the resolved parent. Sibling lookups are
  scoped to the current parent so `/a/b` and `/x/b` are distinct.
- A pure `collect_group_subtree(groups, root_id)` helper computes the set of
  group IDs to delete (root + all transitive descendants). `delete_connection_group`
  now drops every group in that set plus every connection whose `group_id`
  is in the set. The frontend re-fetches via `get_connections_with_groups`
  after the call so the optimistic state is never responsible for the cascade.
- Export/import now writes and reads `parent_id` per group, so a connection
  bundle exported with nested folders re-imports with the same tree.
- The existing `ensure_no_cycle` check still runs on create/update, so the
  hierarchy can never become a cycle even if a user tries to drag a folder
  into its own descendant.

## UI

- Visible `+` button on every group header (replaces the old right-click-only
  path). Hover reveals an amber plus that focuses an inline input rendered
  between the header and its children; Enter confirms, Escape cancels.
- Toolbar "New group" input accepts `/`-separated paths: typing
  `clients/eu/finance` creates the missing segments and returns the deepest
  group so the caller can keep typing.
- Indent in both the list and grid view: sub-folders and their connections
  are pushed right one level per ancestor.
- Group delete is now destructive: a single click on the existing trash
  action removes the whole subtree (groups + connections) with a confirm
  prompt that names how many children will go.

Nested groups with the + button and indent

Image showing the related UI state

Cascade delete confirmation

## Tests

- `collect_group_subtree` is covered for: leaf only, full descendant chain,
  subgroup not including siblings, and unknown id (returns singleton).
- `cascade_delete_removes_parent_descendants_and_connections` and
  `cascade_delete_subgroup_leaves_parent_and_other_subgroups_alone` exercise
  the Tauri command end-to-end against an in-memory store.
- `group_tree_tests.rs` covers the cycle prevention and the path-walking
  helpers (`parse_group_path`, `find_child_group`).
- `export_import_tests.rs` round-trips a three-level tree and asserts
  `parent_id` survives the JSON dance.

`cargo build --release`, `cargo test --lib` (767 passing; the four askpass
failures are pre-existing on `main`), and `pnpm tsc --noEmit` all pass.

## i18n

Added `subgroupNamePrompt` under `groups` and translated the new `useIamAuth`
strings in all eight locales (en, es, de, fr, it, ja, ru, zh). `en` and `it`
didn't have a `groups` namespace before, so it was created.


debba:
Pushed a few corrections on top of your branch, mostly tied to the SQL-client import from #393:

- `convert.rs` wasn't setting `parent_id` on groups created during import, so the build broke once nested groups landed — fixed.
- The import preview's group dropdown listed groups flat; it now renders the hierarchy, and "New group…" can be created inside an existing group (added a parent selector).
- The connection context menu's move-to-group targets are now folded into a hover submenu instead of one long flat list.
- Pulled the group-tree flattening into a small shared helper (`groupTree.ts`) so the context menu, the Select dropdown and the import modal all order groups the same way.
- Filled in the missing Italian and Spanish `groups` translations.

Everything typechecks and the Rust + frontend tests pass.

debba:
Addressed the drag re-parenting warning in `da72707f`.

The `sourceDepth > 0` guard was bailing out for any group already nested one level deep, so subgroups couldn't be moved under a different parent — and it wrongly showed the "cannot move into its own subfolder" error. Since deep nesting is supported everywhere else (recursive render, `/`-path subfolders, import), I dropped that check and kept only the real cycle guard (`isAncestor`), which already blocks moving a group into one of its own descendants.

p4pupro:
Thanks @debba for picking that up — agreed, sourceDepth > 0 was a leftover
from the "only top-level" model and doesn't belong in an N-level
implementation. 
The isAncestor guard is the only one that matters.

PR #415, opened 2026-07-02
fix(mysql): display all result sets from stored procedures returning multiple result sets

### LINKED ISSUE #414: [Bug]: MySQL stored procedures returning multiple result sets only display the first result set
### Describe the bug

**Describe the bug**
When executing a stored procedure that returns multiple result sets, tabularis only displays the first result set. The remaining result sets are not shown.

The same stored procedure returns all result sets correctly when executed in MySQL Workbench.

**Screenshots**
Providing screenshots for comparing the output from tabularis and MySQL Workbench.
[MySQL Workbench](https://github.com/user-attachments/assets/ccdb6166-bc78-4985-bda3-6e80902c9cd7)
[tabularis](https://github.com/user-attachments/assets/bb322a10-e85f-4364-bd89-315f1d4764fa)

**Application:**

- App client: MySQL
- App version: 0.13.4
- Installation source: tabularis_0.13.4_x64_en-US.msi

**Environment:**

- OS name: Windows 11 Pro
- OS version: 25H2
- DB name: MySQL
- DB version: 8.0.46-0ubuntu0.24.04.3

### To Reproduce

1. Connect to a MySQL database using tabularis.
2. Execute a stored procedure that returns multiple result sets (multiple SELECT statements).
3. Observe that only the first result set is displayed.
4. Execute the same stored procedure in MySQL Workbench.
5. Observe that all result sets are returned correctly.

### OS Version

Windows 11 Pro, 25H2

### Tabularis Version

0.13.4

### Relevant Log Output

```shell
The stored procedure intentionally returns multiple result sets using multiple SELECT statements.

Example:


CREATE PROCEDURE sp_test()
BEGIN
    SELECT 'First result' AS message;
    SELECT 'Second result' AS message;
    SELECT 'Third result' AS message;
END;


Expected output:

Result Set 1: `First result`
Result Set 2: `Second result`
Result Set 3: `Third result`

Actual behavior in tabularis:

- Only Result Set 1 is displayed.

Actual behavior in MySQL Workbench:

- All three result sets are displayed correctly.

I'm not sure whether this is a missing feature or a bug, but it prevents working with stored procedures that intentionally return multiple result sets.
```

debba:
Fixes #414

## Problem

When executing a MySQL statement that returns multiple result sets — typically a `CALL` to a stored procedure whose body contains several `SELECT`s — Tabularis only displayed the first result set. MySQL Workbench shows all of them.

```sql
DELIMITER //
CREATE PROCEDURE sp_test()
BEGIN
    SELECT 'First result' AS message;
    SELECT 'Second result' AS message;
    SELECT 'Third result' AS message;
END//
DELIMITER ;

CALL sp_test();  -- Tabularis showed only "First result"
```

## Root cause

`exec_on_mysql_conn` consumed the wire stream with sqlx's `fetch()`, which yields a flat stream of rows with no result-set boundaries: column metadata was taken from the first row and everything was folded into a single `QueryResult { columns, rows }`. The data model itself had no way to represent more than one result set per statement. This is the same thing MySQL Workbench handles by looping `mysql_next_result()` after enabling `CLIENT_MULTI_RESULTS`.

## Fix

**Backend**
- `exec_on_mysql_conn` now uses sqlx's `fetch_many()`, whose event stream interleaves rows with one terminator per result set.
- New pure helper `drivers/mysql/multi_result.rs` (`ResultSetCollector`) folds that stream back into discrete sets; the per-page row cap now applies to each result set independently.
- `QueryResult` gains an optional `additional_results: Option<Vec<QueryResult>>`. The first result set stays in `columns` / `rows`, so every consumer unaware of multi-result statements keeps working unchanged; the field is skipped in JSON when absent, leaving Postgres/SQLite/plugin drivers untouched.

**Frontend**
- `QueryResult` TS type mirrors the new field.
- When `execute_query` returns additional result sets, `runQuery` renders them through the existing multi-result tab UI (the one used for multi-statement runs): one tab per result set, labelled "Result 1..N" via the new `editor.multiResult.resultSetPrefix` i18n key (added to all 8 locales).
- Row-editing metadata (`activeTable` / `pkColumns`) is intentionally skipped — procedure output is not row-editable.

## Note on `DELIMITER`

Creating the procedure still requires the usual client-side `DELIMITER` dance (see the snippet above): `DELIMITER` is not server SQL, it only tells the client's statement splitter to stop splitting on the `;` inside the procedure body. Tabularis' SQL splitter already supports `DELIMITER //` for the MySQL dialect, exactly like the `mysql` CLI and Workbench — so when testing this PR, remember to wrap the `CREATE PROCEDURE` in `DELIMITER //` … `END//`, otherwise the procedure itself won't be created (that part is expected behaviour, not part of this bug).

## Testing

- **End-to-end**: ran the actual driver path (`drivers::mysql::execute_query`) against a live MariaDB with the reproducer procedure from the issue — the call now returns the first set as primary plus 2 `additional_results`; a plain `SELECT` is byte-for-byte identical to before (no extras, pagination unchanged).
- `cargo test` (mysql driver suite): 30 passed, including 7 new unit tests for `ResultSetCollector` (set splitting, per-set truncation, empty-set handling, unterminated stream flush).
- `pnpm test`: 2711 passed, including 7 new tests for `createEntriesFromResultSets`.
- `cargo check` clean (no deprecated API), `tsc --noEmit` clean.

## Known limitations

- A result set that produces **zero rows** is dropped: sqlx exposes no column metadata without rows, which makes it indistinguishable from the trailing OK packet of a `CALL`. Same pre-existing limitation as a rowless single query.
- The **batch path** (multiple statements in one run) already receives `additional_results` from the backend, but its UI still shows one result per statement; expanding extra sets there is left as a follow-up.
- Pre-existing gap spotted while translating: `zh.json` is missing several `editor.multiResult.*` keys (`queryPrefix`, `viewTabs`, …) — only the new key was added here.


arnabmunshi-shbl:
Hi @debba 
I have follow the readme.md and I have tried very heard to run this PR on my local desktop (windows 11), but I am unable to do that.
I have only node js installed, then I also installed rust (not sure properly installed or not).
Then I have started the app but getting so much issue in my end.
Could you (@debba) please guide me to run the application on my end to test the bug fixes.

debba:
Hi @arnabmunshi-shbl,

if you'd like to test the fix before it lands in a release, you can build the app on your machine. Tabularis is a Tauri 2 app (Rust backend + React frontend), so you'll need:

1. Visual Studio Build Tools 2022 — https://visualstudio.microsoft.com/visual-cpp-build-tools/ — select the "Desktop development with C++" workload (includes MSVC and the Windows SDK).
2. Rust via https://rustup.rs (the default stable-msvc toolchain is correct).
3. Node.js 22 (the repo pins 22.18.0 in .node-version) — https://nodejs.org
4. pnpm 10: after installing Node, run corepack enable (or npm install -g pnpm).
5. WebView2 runtime — already present on Windows 11, nothing to do.

Then, in PowerShell:

git clone https://github.com/TabularisDB/tabularis.git
cd tabularis
git fetch origin pull/415/head:pr-415
git checkout pr-415
pnpm install
pnpm tauri dev        # runs the app directly, fastest way to test

If you want a standalone binary instead of the dev mode, use:

pnpm tauri build --no-bundle

and you'll find the portable executable at src-tauri\target\release\tabularis.exe (--no-bundle also skips the updater-signing step that would otherwise require our private signing key).

Let us know if the fix works with your real stored procedures!

PR #417, opened 2026-07-02
feat(sidebar): typo-tolerant fuzzy matching for the table & trigger filters

Davydhh:
## Summary
The sidebar "Filter tables…" boxes currently do a strict, case-insensitive substring match. This switches them to **typo-tolerant fuzzy matching** via Fuse.js — so a misspelling like `ordrs` still finds `orders`, and results are ranked best-match-first.

## Changes
- New `src/utils/fuzzy.ts` — a small `fuzzyFilter(items, query, getText)` helper wrapping Fuse.js (`threshold: 0.4`, `ignoreLocation: true`), with tests.
- Replaced the substring filter with `fuzzyFilter` in the **table and trigger** filters across all three sidebar layouts: schema (`SidebarSchemaItem`), multi-database (`SidebarDatabaseItem`), and flat (`ExplorerSidebar`).
- Left the favorites filter (matches name *and* SQL) and the database-selector filter on substring for now — different contexts, easy to switch later.

## New dependency
Adds **`fuse.js` (^7.4.2)** — a small, zero-dependency package (~few KB gzipped) and the most widely used JS fuzzy-search library; it provides the typo tolerance. Flagging explicitly since it's a new runtime dependency — happy to swap in a dependency-free implementation instead if you'd prefer to avoid it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "clsx": "^2.1.1",
     "dagre": "^0.8.5",
     "emoji-picker-react": "^4.19.1",
+    "fuse.js": "^7.4.2",
     "i18next": "^25.10.10",
     "i18next-browser-languagedetector": "^8.2.1",
     "json-edit-react": "^1.29.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -59,6 +59,9 @@ importers:
       emoji-picker-react:
         specifier: ^4.19.1
         version: 4.19.1(react@19.2.4)
+      fuse.js:
+        specifier: ^7.4.2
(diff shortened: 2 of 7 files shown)


debba:
Hi @Davydhh ,
looks good to me, I will merge #412 before.
We have also a Quick Navigator, which can be called using ⌘ + P / CTRL + P .
Do you thin can you apply fuzzy filter also in Quick Navigator?

Davydhh:
Hi @debba,
Yes, I'll do it in a separate pull request (after the merge in order to have the new dep) since it's specific to tables and triggers.

Issue #413, opened 2026-07-02
[Bug]: visible ssh.exe process terminal window when connecting to a DB via ssh tunnel on Windows 11


kennelken:
### Describe the bug

On windows connecting to a DB using SSH tunnel spawns a visible terminal with an ssh.exe process.
This ssh.exe process stays alive even when Tabularis gets closed.

<img width="1566" height="1022" alt="Image" src="https://github.com/user-attachments/assets/e8d219fa-7c9a-435c-a0bf-02fb3149eac1" />

### To Reproduce

1. Open Tabularis
2. Create a new DB connection that uses SSH tunnel
3. Connect to the newly created connection
4. Notice that there is a new terminal window opened automatically with a process of ssh.exe (see an attached screenshot)

### OS Version

Windows 11

### Tabularis Version

v0.13.2

### Relevant Log Output

```shell

```

debba:
Hi @kennelken !
I don't use windows, do you think are you able to create a PR for it?

kennelken:
Hi @debba. Thank you for your response. I will take a look at what i can do.

debba:
Thanks @kennelken ,
this kind of bugfixes are very welcome in Windows area :)

kennelken:
> Hi [@kennelken](https://github.com/kennelken) ! I don't use windows, do you think are you able to create a PR for it?

I created a MR for this issue https://github.com/TabularisDB/tabularis/pull/418

kennelken:
Tested in v0.15.0. Works as expected.

PR #418, opened 2026-07-02
fix(ssh): prevent visible ssh.exe console window on Windows and clean up processes on exit

kennelken:
Fix for issue: https://github.com/TabularisDB/tabularis/issues/413

# Title
`fix(ssh): prevent visible ssh.exe console window on Windows and clean up processes on exit`

## Description
This PR resolves a Windows-specific bug where establishing an SSH tunnel spawned a visible command prompt/terminal window running `ssh.exe` that remained running in the background even after Tabularis was closed.

### 1. Silent Process Spawning on Windows
On Windows, launching a console subsystem executable (like `ssh.exe`) from a GUI subsystem application (without a console) causes the OS to automatically allocate and open a new console window.
- **Fix**: Replaced direct `Command::new("ssh")` calls in the system SSH builder and connection tester with a helper function that applies the Windows-only `CREATE_NO_WINDOW` creation flag (from `std::os::windows::process::CommandExt`).

### 2. Orphaned Process Cleanup
Spawning a `std::process::Child` in Rust does not automatically terminate the child process when the parent exits or when the `Child` struct is dropped.
- **Fix**: Hooked into Tauri's `RunEvent::Exit` lifecycle event inside `lib.rs` to invoke a new `stop_all_tunnels()` cleanup function on application exit, ensuring any active system SSH processes are forcefully terminated.

---

### 1. visible console window fix
1. Run a build/packaged application (or run `pnpm tauri build` and open the release binary directly from Windows Explorer).
2. Create and connect to a database connection configured with an SSH tunnel.
3. **Verify**: No command prompt or `ssh.exe` window pops up on the screen.

### 2. Orphaned process cleanup
1. Open an SSH tunnel database connection in Tabularis.
2. Close the Tabularis application.
3. Open Task Manager or run `tasklist | findstr ssh` in command prompt.
4. **Verify**: The spawned `ssh.exe` process is terminated and no longer running.


debba:
Thanks for this PR, I will merge it now.

PR #421, opened 2026-07-03
feat(navigator): fuzzy search in the Quick Navigator

Davydhh:
## Summary
Follow-up to #417 (fuzzy search for the sidebar filters). The Quick Navigator (⌘P) still did a strict substring match — this routes it through the same shared `fuzzyFilter` helper, so it's now **typo-tolerant** and ranks the closest matches first.

## Changes
- `filterNavigatorItems` now uses `fuzzyFilter` (Fuse.js), searching a combined `name + schema` string — this preserves the previous "match name *or* schema" behaviour, now fuzzy and relevance-ranked.
- Added a typo-tolerance test; the existing name/schema tests pass unchanged.


debba:
Looks good to me, thanks for this.

PR #426, opened 2026-07-03
fix(mcp): serialize multiple databases in list_connections and add list_databases tool

erneztox:
Hey! As discussed in the issue, here is the PR to improve the MCP server database discovery.

### What changed?
1. **Fix `list_connections` array serialization**: In `src-tauri/src/mcp/mod.rs`, changed the `database` property extraction from `.to_string()` (which called `.primary()` and dropped items) to `.as_vec()`. Now the MCP client correctly sees all databases configured in a single connection.
2. **Add `list_databases` tool**: Added a native `list_databases` tool leveraging the existing `driver.get_databases()` trait. This allows the LLM to query and discover all available databases natively through the connection without needing to manually run SQL queries.

### Testing
- Tested locally by passing JSON-RPC requests via stdio to the `tabularis --mcp` binary.
- Validated that `list_connections` outputs the full array.
- Validated that `list_databases` connects to the engine and returns the schemas correctly.
- Code compiles perfectly with `cargo check`.

Let me know if you need any adjustments!


debba:
Hey, thanks a lot for this! I reviewed it locally and I'm happy with it — the fix is correct and the new tool follows the existing patterns nicely.

I verified everything end-to-end against the demo stack (`demo/docker-compose.yml`), driving `tabularis --mcp` over stdio:

- `tools/list` advertises `list_databases` alongside the existing tools ✅
- `list_connections` now returns the full array — the demo MySQL connection correctly shows `["tabularis_demo","blog_demo","perf_demo"]` instead of just the first one ✅
- `list_databases` returns the real database list on both MySQL and Postgres ✅
- Missing `connection_id` → `-32602`, unknown connection → `-32000` ✅
- The audit trail (`ai_activity.jsonl`) records the new tool with correct row counts and error statuses ✅

`cargo check` is clean and the existing test suites pass on the branch.

A few small things I noticed — none of them blocking, but since they're all quick I'd prefer to get them into this PR rather than leaving them behind:

1. **README tools table** — the MCP tools table in `README.md` (and the translated READMEs) doesn't list `list_databases` yet.
2. **AI Activity filter** — the tool filter dropdown in `src/components/settings/ai-activity/AiActivityEventsTab.tsx` hard-codes the tool names, so `list_databases` events show up in the list but can't be filtered. Adding it to `toolOptions` (and to the `AiToolName` union in `src/types/ai.ts` for completeness) would be great.
3. **A small test** — a dispatch-level test for the new tool (e.g. missing/unknown `connection_id` → the right JSON-RPC errors) would be nice to have per our testing conventions.

One unrelated thing my testing surfaced: `keychain_utils.rs` uses `println!`, so `[Keychain] ...` log lines end up on stdout inside the MCP JSON-RPC stream. That's pre-existing (it happens on `main` too), so no need to touch it here — I'll open a separate issue for it.

Thanks again! 🙏

PR #433, opened 2026-07-07
fix(grid): keep result column headers pinned while scrolling

### LINKED ISSUE #411: [Feat]: Sticky Column Headers
### Prerequisites

### 🚀 Feature Proposal

It would be nice to have an option to make the column headers sticky. That means, if I scroll down the list, the header is still visible. It currently stays visible for like 40 rows and then disappears.

### Motivation

Better overview especially with a table with more than 5-10 columns.

### Example

_No response_

thomaswasle:
closes #411

## Summary

- Fixes the result grid's column headers detaching after ~40 rows of scrolling. The `<thead>` used `position: sticky` plus a manually-computed counter-transform to cancel out the virtualized `<table>`'s own `translateY` offset — sticky positioning and transforms compose in CSS, so the counter-transform grew with scroll depth and eventually desynced the sticky calculation, causing the header to fall out of view.
- Replaces the transform-based virtualization with the standard TanStack Virtual table pattern (top/bottom spacer rows instead of an absolutely-positioned, transformed `<table>`), so the table stays in normal document flow and `sticky top-0` works correctly at any scroll depth.
- Adds a **"Sticky column headers"** toggle under Settings → Appearance → General (`stickyColumnHeaders`, default `true`), for anyone who prefers the old non-sticky scrolling behavior.

## Why

Column headers were meant to stay pinned while scrolling but silently detached after roughly 40 rows (~1400px of scroll), making it hard to keep track of which column is which in large result sets.

## Changes

- `src/components/ui/DataGrid.tsx` — removed the double transform on `<table>`/`<thead>`; added `virtualPaddingTop`/`virtualPaddingBottom` spacer `<tr>`s driven by `rowVirtualizer.getVirtualItems()`; `<thead>` now conditionally gets `sticky top-0` based on the new setting.
- `src/contexts/SettingsContext.ts` — added `stickyColumnHeaders?: boolean` to `Settings`, default `true`.
- `src/components/settings/AppearanceTab.tsx` — added a `SettingSection`/`SettingRow`/`SettingToggle` for the new setting.
- `src/i18n/locales/en.json` — added `settings.dataGrid.{title,stickyHeaders,stickyHeadersDesc}` keys (other locales fall back to the English `defaultValue`, consistent with existing untranslated settings strings).

## Test plan

- [x] `tsc --noEmit` and `eslint` clean on all changed files
- [x] `vitest run` — full suite passes (2733 tests); pre-existing unrelated failures (missing `fuse.js`/`sql-formatter` deps in this sandbox) confirmed present on `main` too
- [x] Manually run a query returning 150+ rows and scroll to the bottom — header should stay pinned throughout
- [x] Toggle "Sticky column headers" off/on in Settings → Appearance and confirm the header un-pins/re-pins accordingly


debba:
Awesome!
I will merge it now

PR #436, opened 2026-07-07
feat(datagrid): show column type in header hover tooltip (#435)

### LINKED ISSUE #435: [Feat]: Show column type in a tooltip on header hover (DataGrip-style)
## Summary

Show a column's data type in a tooltip when hovering over its header in the data grid, similar to JetBrains DataGrip.

## Motivation

When browsing a result set / table, it's often useful to know a column's type (e.g. `uuid`, `varchar`, `int4`, `timestamptz`) without opening the table structure view or the sidebar editor. DataGrip shows this inline on header hover, which is fast and unobtrusive.

## Proposed behavior

- Hovering over a column header shows a small tooltip with `<column_name>: <type>`.
- Example: `id: uuid`, `birthdate: date`, `valuation: int8`.
- Type comes from the already-available column metadata returned with the query result / schema introspection (no extra query needed).
- Should work for query results and table browse views, across all supported drivers (Postgres, MySQL/MariaDB, SQLite, ...), using whatever type name each driver reports.

## Reference

DataGrip-style header hover tooltip showing `id: uuid`.

## Notes

Happy to implement this — taking it on.

benedettoraviotta:
## What

Adds a DataGrip-style tooltip on data-grid column headers: hovering a header shows `<name>: <type>` (e.g. `id: uuid`).

Closes #435

## How

- The column data type is read from the already-available `columnMetadata` (`data_type`) via the existing `columnTypeMap` in `DataGrid.tsx` — **no backend changes**.
- The tooltip reuses the existing floating-surface look through the reserved-but-previously-unused `--bg-tooltip` theme token (`bg-tooltip border border-strong rounded-lg shadow-xl`), with the same `group-hover` fade used elsewhere.
- Positioned with `absolute` (not `fixed`) because the sticky `<thead>` carries a `translateY` transform, which would break `fixed` anchoring.
- `pointer-events-none` so it never intercepts header clicks/sort.
- The native sort-hint `title` is suppressed only while the type tooltip is shown, to avoid two overlapping tooltips on hover.

## Scope

Shows whenever column metadata is present (table browse). Arbitrary query results carry no per-column type in `QueryResult`, so no tooltip is shown there — sort-hint title behavior is unchanged for those columns.

## Verification

- `pnpm typecheck` — pass
- `pnpm exec eslint src/components/ui/DataGrid.tsx` — pass
- Verified visually via `pnpm tauri dev` (Postgres table browse).


debba:
Looks good!
Thanks for it :)

PR #447, opened 2026-07-08
feat(connections): add export modes with optional password encryption

pokertour:
The connections export previously always wrote database, SSH and K8s credentials in plaintext behind a simple warning dialog. Replace it with a dedicated export modal offering three modes:

- Encrypted with a password (default): the payload is encrypted with AES-256-GCM using an Argon2id-derived key, in a new export_crypto module with unit tests
- Plain text without passwords: secrets are stripped from the payload
- Plain text with all passwords: previous behavior, with the warning shown inline on the option

Import now detects the encrypted envelope and prompts for the password before decrypting and merging. Plain exports from older versions still import unchanged, and export_connections_payload keeps resolving secrets by default for backward compatibility.


debba:
Hi @pokertour,
Just some adjustments to align with the new ImportModal shipping in the next release, already in `main` (#393). Let me know if you want to run any final tests, otherwise this should be ready to merge.

pokertour:
Hi @debba
Just tested your merge, working fine on my side.
Good to go

PR #455, opened 2026-07-09
fix(mysql): display ENUM allowed values and add dropdown editing (#452)

### LINKED ISSUE #452: [Bug]: MySQL ENUM columns do not show allowed values and ENUM type not available when creating tables
### Describe the bug

When working with a MySQL database in Tabularis:

- Columns defined as ‎`ENUM` do not show any of their allowed values.

- In the generated “Create Table” SQL, the column type is displayed only as ‎`enum` without the list of allowed values.

- Additionally, ‎`ENUM` is not available as a column type option when creating a new table from the Tabularis UI.

This makes it difficult to understand and manage enum-based schemas directly from Tabularis.

### To Reproduce

1. In MySQL, create a table with an enum column, for example:

```sql
CREATE TABLE example (
  id INT PRIMARY KEY AUTO_INCREMENT,
  status ENUM('pending', 'approved', 'rejected') NOT NULL DEFAULT 'pending'
);

```

2. Connect Tabularis to the MySQL database.

3. Open the ‎`example` table in Tabularis.

4. Inspect:

 ▫ How the ‎`status` column is displayed in the table schema / columns list.

 ▫ The generated “Create Table” SQL shown by Tabularis (if applicable).

**Expected behavior**

- The ‎`status` column should display the full enum definition, including all allowed values (for example ‎`ENUM('pending', 'approved', 'rejected')`).

- The generated “Create Table” SQL should include the actual enum values, not just the generic ‎`enum` type.

**Actual behavior**

- The enum column shows only the generic type ‎`enum` with no values.

- The “Create Table” SQL for the enum column is shown as just ‎`enum` (without the allowed values).

### OS Version

macOS Tahoe 26.5.1

### Tabularis Version

0.14.0

### Relevant Log Output

```shell

```

fhriz:
fix: display MySQL ENUM allowed values and add dropdown editing

- Introspect `column_type` from information_schema to capture full ENUM
  definitions (e.g. `enum('pending','approved','rejected')`) instead of
  the stripped `data_type` value. Applies to both `get_columns` and
  `get_all_columns_batch`. Also handles SET types for completeness.

- Add ENUM to the MySQL data type registry so it appears in the column
  type picker when creating new tables.

- DataGridRow: render a <select> dropdown when inline-editing an ENUM
  cell, parsing allowed values from the column type. Nullable columns
  get a NULL option that maps empty selection to null.

- FieldEditor: same dropdown treatment for the row editor sidebar, so
  editing ENUM columns from the sidebar also shows allowed values.

- RowEditorSidebar columns mapping: match column metadata by name
  instead of array index, fixing type mismatches when a Console query
  SELECT order differs from the table's ordinal_position.

- Add isEnumType() and parseEnumValues() helpers to columnTypes utils.

Closes #452


debba:
Hi @fhriz, thanks a lot for this, great foundation. I've pushed a follow-up commit on top of your branch that builds on the ENUM work:

**What I added**
- **SET multi-select editing.** Your driver change already surfaced `set('…')` via `column_type`, so I wired up the frontend for it: `isSetType`/`parseSetValues` plus `parseSetMembers`/`serializeSetMembers` helpers (serialization preserves the column's declared value order), with unit tests.
- **Unified ENUM/SET editor (`EnumSetInput`).** A single reusable component replacing the raw native `<select>`, rendered as a portal dropdown styled to match our shared `Select` component. ENUM = single choice; SET = checkboxes with a chip summary. Used by both the grid cell editor and the row sidebar (`FieldEditor`).
- **Commit fix.** Added `commitEditWithValue` in `DataGrid` so a pick-and-commit within a single click reads the fresh value — the previous change-then-commit path read `editingCellRef` while it was still stale (synced via a passive effect). This is why ENUM picks didn't persist before.
- The grid editor now overlays the cell edge-to-edge to avoid extra padding.
- `dataGrid.emptySet` added to all locales, and a `07-enum-set-demo.sql` demo seed exercising ENUM/SET, escaped quotes, and empty/NULL sets.

Everything passes typecheck, lint, and the full test suite (2769 tests).

Could you take a look and let me know what you think? If there's anything you'd like to adjust or add, go ahead, otherwise I'm happy to proceed with the merge. 🙌

fhriz:
hi @debba awesome! I agree with using reusable component for EnumSetInput

let's merge

PR #456, opened 2026-07-10
fix(mcp): EXPLAIN ANALYZE bypasses read-only and the approval gate

debba:
## Context

Security report received by email from an external researcher (Daniel Mertz). The MCP safety layer (`McpSafety`) guarantees two things about queries arriving from the MCP/agent surface: **read-only mode** and the **write-approval prompt**. Both guarantees could be bypassed.

## The problem

`classify_query_kind` (`src-tauri/src/ai_activity.rs`) classified **any** statement whose leading keyword is `EXPLAIN` as a read (`"select"`), without looking at the options.

But on PostgreSQL `EXPLAIN ANALYZE ...` and `EXPLAIN (ANALYZE ...) ...` **actually execute** the wrapped statement (that's how they gather real timings). So:

```sql
EXPLAIN ANALYZE DELETE FROM some_table;
EXPLAIN ANALYZE DROP TABLE some_table;
EXPLAIN (ANALYZE, BUFFERS) UPDATE some_table SET x = 1;
```

were classified `"select"` and, in `src-tauri/src/mcp/mod.rs`, skipped both gates:

- **Read-only** (`mod.rs:913`): `is_connection_readonly(...) && kind != "select"` → with `"select"` it **does not block**.
- **Approval gate** in `writes_only` mode (`mod.rs:933`): `kind != "select"` → with `"select"` **no prompt**.

Result: arbitrary writes/DDL executed past both guarantees, with no approval.

### Threat model

Relevant when the query text comes from an untrusted source — a connected MCP server or a prompt-injected agent turn — which is exactly the scenario the approval gate exists to contain. Estimated severity **High** (CVSS 3.1 ~7.0) under this model; drops to Medium for a fully-trusted local operator.

## The fix

`EXPLAIN` classification is now **option-aware** (`classify_explain` + helpers):

- **Plain EXPLAIN** (no `ANALYZE`) → `"select"`: it only plans and never touches data. This also covers `EXPLAIN DELETE ...` (Postgres does not execute without ANALYZE), `EXPLAIN (FORMAT JSON) ...`, `EXPLAIN QUERY PLAN ...` (SQLite), `EXPLAIN VERBOSE ...`.
- **EXPLAIN with the `ANALYZE` option** → the **wrapped statement** is classified as if it had been submitted directly (`write` / `ddl` / `unknown`).

Details:
- Handles both the legacy syntax (`EXPLAIN ANALYZE ...`) and the parenthesized form (`EXPLAIN (ANALYZE, BUFFERS) ...`), with `ANALYZE` anywhere in the option list.
- **Word-boundary aware**: `ANALYZE` used as an identifier (e.g. `EXPLAIN SELECT * FROM analyze_runs`) does **not** produce false positives. `contains_keyword` now reuses the new `keyword_index` helper.
- **Fail-closed**: an unbalanced parenthesized option list or an unrecognized inner statement → `"unknown"` (the gates close).
- Also covers executed CTEs: `EXPLAIN ANALYZE WITH ... DELETE ...` → `write`.

No change needed in `mcp/mod.rs`: both gates derive from `kind`, now correct both for the original query (`mod.rs:906`) and for the query edited at approval time (`mod.rs:1018`).

## Tests

Added 6 regression tests in `ai_activity_tests.rs`, including the one requested in the report (`EXPLAIN ANALYZE DELETE` handled identically to a bare `DELETE`), plus parenthesized forms, DDL, CTE-write, `EXPLAIN ANALYZE SELECT` (stays a read) and unbalanced options.

- `cargo test --lib classify` → 37 passed
- `cargo test --lib ai_activity` → 66 passed
- `cargo clippy --lib` → no warnings on ai_activity

## Impact analysis

`classify_query_kind` is called only from the two call sites in `mcp/mod.rs` plus the tests. The change **tightens** classification (moves `EXPLAIN ANALYZE` from `select` to write/ddl/unknown): it is fail-safe — no legitimate read is unblocked, only the executing EXPLAIN is now correctly subjected to the gates.

## Acknowledgements

Many thanks to **Daniel Mertz** ([danielmertz.com](https://danielmertz.com) — @daniel-mertz) for responsibly reporting this vulnerability and for reviewing the fix.

### DIFF
--- a/src-tauri/src/ai_activity.rs
+++ b/src-tauri/src/ai_activity.rs
@@ -398,7 +398,13 @@ pub fn classify_query_kind(sql: &str) -> &'static str {
     let first = first_keyword(&peeled_upper);
 
     match first.as_str() {
-        "SELECT" | "SHOW" | "EXPLAIN" | "DESCRIBE" | "DESC" | "PRAGMA" | "VALUES" => "select",
+        // EXPLAIN needs option-aware handling: plain EXPLAIN only plans, but
+        // EXPLAIN ANALYZE (and EXPLAIN (ANALYZE ...)) *executes* the wrapped
+        // statement. Classifying it blindly as a read would let
+        // `EXPLAIN ANALYZE DELETE …` run past read-only mode and the approval
+        // gate. See `classify_explain`.
+        "EXPLAIN" => classify_explain(&peeled_upper),
+        "SELECT" | "SHOW" | "DESCRIBE" | "DESC" | "PRAGMA" | "VALUES" => "select",
         "INSERT" | "UPDATE" | "DELETE" | "MERGE" | "REPLACE" => "write",
         "CREATE" | "DROP" | "ALTER" | "TRUNCATE" | "RENAME" | "GRANT" | "REVOKE" | "COMMENT" => {
             "ddl"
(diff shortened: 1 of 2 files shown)


daniel-mertz:
Reporter here — thanks for the fast turnaround, Andrea. The option-aware classification (matching on word boundaries, failing closed on unbalanced option lists) plus the regression test gating `EXPLAIN ANALYZE DELETE` identically to a bare `DELETE` is exactly the right fix. LGTM from my side. — Daniel Mertz (danielmertz.com)

Issue #446, opened 2026-07-08
[Bug]: Fix auto paginated color in Tabularis Light


math-krish:
### Describe the bug

When viewing the table in the Tabularis Light theme, the word Auto Paginated, comes in colors that is very difficult to read. Could we use something that goes better with the theme? 
<img width="422" height="174" alt="Image" src="https://github.com/user-attachments/assets/6a5a055d-5af5-4543-8be3-ba7a9137c675" />

### To Reproduce

1. Go to Settings -> Appearance
2. Pick Tabularis Light
3. Go back and select a table you want to view which has enough records to be paginated. 
4. Observe the word auto paginated

### OS Version

macOS 15.7.4

### Tabularis Version

v0.14.0

### Relevant Log Output

```shell

```

debba:
Hello @math-krish,
yeah, you can go ahead and make the changes.

You can create a PR for each issue, or, if they are small changes like this, you can also create a single PR that addresses multiple issues and closes them together.

These kinds of issues are super welcome, so thanks for taking the time to contribute! :)

PR #461, opened 2026-07-11
fix: fixes auto pagination color, dropdown menu font and save menu

### LINKED ISSUE #448: [Bug]: All dropdowns render a different font when opened
### Describe the bug

I checked multiple dropdown, and the font changes after we open it: 
Image

Image

### To Reproduce

1. Open any table, look at the font around the word "Comma"
2. Open the dropdown
3. Notice font change for the options listed

### OS Version

macOS 15.7.4

### Tabularis Version

v0.14.0

### Relevant Log Output

```shell

```

math-krish:
## Summary: 
Fixes #448 and #446 . Also fixes the menu that appears on edit, before confirming your save. Screenshots shown below: 

**Fix 1:** 
Before: 
image
After: 
image

Fix 2: 
Before:
image
After:
image

Fix 3: 
Before: 
image
After:
image

### DIFF
--- a/src/components/ui/ResultEntryContent.tsx
+++ b/src/components/ui/ResultEntryContent.tsx
@@ -143,7 +143,7 @@ export function ResultEntryContent({
             )}
           </span>
           {entry.result.pagination?.has_more && (
-            <span className="px-2 py-0.5 bg-yellow-900/30 text-yellow-400 rounded text-[10px] font-semibold uppercase tracking-wide border border-yellow-500/30">
+            <span className="px-2 py-0.5 bg-accent-warning/15 text-accent-warning rounded text-[10px] font-semibold uppercase tracking-wide border border-accent-warning/50">
               {t("editor.autoPaginated")}
             </span>
           )}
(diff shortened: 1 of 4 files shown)


debba:
@math-krish 
I really love it. Since the window can be resized, I'd just think about how to present the available options when the window width becomes smaller.

This is something that affects the application as a whole, it might make sense to handle it in a separate PR.

What do you think?

math-krish:
Sure @debba!
That makes sense, I'll remove the Fix 2 from this PR then

debba:
@math-krish 

We can merge it as it is for now. At the moment, this isn't handled much better elsewhere, and it's still an improvement over the current behavior. 🙂

We could then open a separate PR to handle collapsing the UI when the window becomes narrower, or even enforce a minimum window width by limiting how much it can be resized.

math-krish:
@debba I just reverted the other change 😂 do I add it back then? It'll be quick, lemme know, I have the commit ready.

debba:
ahah you are faster than me, let me merge it and if you want you can achieve that in another PR

math-krish:
Done, added it back!

PR #464, opened 2026-07-12
feat(frontend): run statement at cursor on Cmd+Enter

maximumbreak:
## Demo : 

https://github.com/user-attachments/assets/b78cba00-741d-4f44-9377-933dcc527795

## Summary

- Cmd+Enter, Ctrl/Cmd+F5, and the Explain button now run/explain the SQL statement the cursor is currently inside when nothing is selected, instead of running the whole file or popping open a "pick a statement" modal — matching TablePlus/DataGrip behavior.
- A subtle, theme-aware background highlight shows which statement is armed to run whenever there's no selection, and disappears the instant you make a selection.
- Whole-file / multi-statement execution is still available by selecting all text first (Cmd+A) — no new UI needed for that.
- Explaining a non-explainable statement (e.g. DDL) now shows a warning alert instead of silently doing something unexpected.
- Fixes two related bugs surfaced by this workflow: running two different statements/tables back to back on the same tab could leave stale primary-key/column metadata from the *previous* statement applied to the *new* one (breaking inline cell editing with a misleading "column not editable" error naming the wrong table), and every pagination/refresh action (including the post-edit-save refresh) re-sent the *entire* editor buffer instead of the single statement that produced the current result — which Postgres rejects as "cannot insert multiple commands into a prepared statement" whenever the buffer holds more than one statement.

## Details

- Added `findStatementAtOffset` to `sqlSplitter`, resolving which statement a cursor offset falls inside using the splitter's existing (previously unused) per-statement `range` data. A cursor in the gap between two statements resolves to the preceding one; a cursor before the first statement does nothing.
- `handleRunButton`/`handleExplainButton` in `Editor.tsx` now resolve the statement at the cursor via a shared helper instead of falling back to the whole file.
- `SqlEditorWrapper` gained an opt-in `dialect` prop that drives the cursor-statement highlight decoration; other consumers of the component are unaffected since they don't pass it.
- `runQuery` now tracks the exact statement it last ran per tab and a per-tab generation counter, so async responses (PK/column metadata fetch, the main query result) from a superseded run are dropped instead of clobbering a newer run's state, and pagination/refresh call sites now re-run "whatever's currently shown" instead of the raw multi-statement editor buffer.
- `QuerySelectionModal` is left in the codebase but no longer reachable from the Run/Explain paths.

## Test plan

- [x] `pnpm test` — 2797 tests pass, including new coverage for `findStatementAtOffset` (statement boundaries, gaps between statements, before/after the statement list, single-statement files)
- [x] `pnpm typecheck` / `eslint` clean
- [x] Manual pass in the running app: multi-statement tab, cursor in different statements with no selection (highlight + Cmd+Enter/F5/Explain all target the right one), select-all still runs everything, editing a cell after running a single statement out of a multi-statement tab, paging through results from that same scenario, and saving an edit — all previously reproduced the "multiple commands in a prepared statement" error and now work correctly

Let me know if you'd like it trimmed down or adjusted before you post it.


debba:
Thanks a lot for this, it works great! I just took the liberty of pushing one small addition on top of your branch: a **Run All** command (Ctrl/Cmd+Shift+Enter in the editor, Ctrl/Cmd+Shift+F5 globally, plus a dedicated entry at the top of the Run dropdown), so running the whole tab stays one keystroke away now that Cmd+Enter targets the statement at the cursor.

Everything else looks good as is, merging right away. 🎉

debba:
One more note: this overlaps with #487, which proposes the same run-statement-at-cursor behavior but gated behind a dedicated setting. I agree with adding that setting so the current picker-modal behavior remains available for those who prefer it — since this PR deliberately left `QuerySelectionModal` in the codebase, reintroducing it behind a toggle is straightforward. I'll follow up in #487 on how to rebase it on top of this once merged.

PR #470, opened 2026-07-13
feat(backup): automatic encrypted connection backups with local and Webdav targets

pokertour:
Adds a dedicated Backup settings tab that periodically writes an encrypted connections export (same AES-256-GCM/Argon2id envelope as the manual export; plaintext backups are deliberately not supported):

- backup trigger: manual, interval, on app close or on app launch; interval presets 6h/12h/daily/weekly plus a custom value, with the next backup time shown in the settings
- destinations: local folder or a WebDAV collection (Nextcloud etc.); uploads via PUT, rotation via PROPFIND + DELETE that only ever touches tabularis-backup-*.json files
- the encryption and WebDAV passwords live in the OS keychain, never in config.json
- retention-based rotation, and a scheduler that re-reads the config every minute so settings changes apply without a restart
- unit tests for the pure helpers: file naming, timestamp parsing, rotation selection, due-time computation, tick pacing and WebDAV listing parsing


debba:
Pushed a small UX polish on top of this branch (e8d0e9d) — @pokertour let me know if you're okay with these changes, happy to adjust or revert any of them:

- **Backup password rows now have an explicit resting state.** When a password is already set, the row shows a masked `••••••••` placeholder with a *Change password* button instead of an always-editable empty field (which looked like no password was set). Clicking it reveals the input with Save/Cancel, and the description warns that backups already written keep the old password. Same pattern applied to the WebDAV password.
- **Reordered the tab to follow the setup flow**: Destination first (target + folder/WebDAV fields), then the Backup settings (trigger, password, frequency, retention), with *Back up now* as the last row.
- **Backup log messages now include the trigger** — e.g. `Connections backup (manual) written to …` / `(interval)`, `(on close)`, `(on launch)` — so runs can be told apart in the Logs tab. This added a `trigger` parameter to `run_backup`/`run_and_log`; all callers are internal to the module.
- Locale strings updated in all languages.

TypeScript/ESLint clean, `cargo check` passes, the 13 backup module tests pass.

debba:
Follow-up (04920c7): added the full Korean translation of the `settings.backup` section as well, so the gap mentioned above is closed, all 10 locales now cover the backup tab.

pokertour:
i'm fine with your change @debba

debba:
I will merge it @pokertour

PR #473, opened 2026-07-14
feat(connections): colored tags and environment classification with production safety warning

### LINKED ISSUE #472: [Feat]: Connection environments (dev/prod) with production safety warning, plus free-form colored tags
### Prerequisites

## Problem

All connections look the same today. When you have a dev, a staging and a production database open side by side, nothing tells you which one you are about to write to — one wrong tab and an UPDATE runs against production. Tools like DBeaver solve this with connection types (dev/test/prod) that drive both a visual identity and a confirmation prompt before dangerous operations.

Connections can already be organized into folders and given an individual color, but a connection lives in exactly one folder and a bare color carries no meaning by itself. There is no cross-cutting way to label a connection along several axes at once (team *and* client *and* purpose) — that's what tags are for.

## Proposal

Two complementary concepts, following the DBeaver model:

### 1. Environment (structured, drives behavior)

A dedicated per-connection field: `development` (default), `staging`, `production` — or none. Set in the connection modal, editable later.

The environment drives **safety behavior** for `production` connections:

- **Confirmation prompt before data writes**: INSERT / UPDATE / DELETE statements and inline grid edits ask for confirmation before executing/committing, e.g. *“You are about to modify **production** data on ‹connection name›. Continue?”* with an optional “don’t ask again for this session” escape hatch.
- **Confirmation prompt before destructive DDL**: DROP / TRUNCATE / ALTER get the same prompt (arguably worth confirming everywhere, but at minimum on production).
- Plain SELECTs are never blocked or prompted.

And a **visual identity** so production is recognizable before you even type:

- The environment tints the connection entry in the sidebar and its open tabs.
- A permanent, clearly visible “PRODUCTION” banner/strip while the active connection is a production one.
- A small environment badge on the connection entry.

### 2. Tags (free-form, organizational)

- User-defined tags with a name and a color (e.g. `team-payments`, `client-x`, `analytics`).
- A connection can have multiple tags; tags are shown as colored chips on the connection list and can be used to filter/group the sidebar.
- Purely organizational: tags never trigger warnings — that is the environment’s job. This keeps the safety mechanism impossible to misconfigure by renaming a tag.

## Out of scope

- Read-only enforcement of production connections (possible follow-up).
- Per-statement policies / allowlists.
- Syncing tags or environments to any remote.

## Open questions

- Should the write detection be statement-based (parse the SQL) or event-based (any query that is not a plain SELECT)? Multi-statement scripts need a decision.
- Does “don’t ask again” apply per session, per connection, or per tab?
- Should destructive DDL prompt on *every* environment by default?

### Motivation

Prevent accidental writes to production: mistyping a tab or running a script against the wrong connection can destroy real data, and today nothing in the UI distinguishes a production connection from a throwaway dev one. A structured environment field gives production connections an unmistakable visual identity and a confirmation step before destructive operations — a proven safety net in tools like DBeaver — while colored tags make large connection lists manageable across teams, clients and purposes.

### Example

_No response_

pokertour:
Closes #472

 ## What

  Two complementary features for organizing connections and protecting production data, following the DBeaver model:

  ### Free-form colored tags
  - `ConnectionTag {id, name, color}` stored in `connections.json` alongside groups; connections carry `tag_ids`.
  - Tag picker in the connection modal (appearance tab): inline creation with the shared accent palette, manage mode to rename/recolor/delete. Names are unique (case-insensitive) and capped at 32 chars; colors validated as hex.
  - Colored chips on connection cards and list rows; tag names participate in the connection search filter.
  - Tags ride along in export/import/backups: selective exports only include used tags, imports merge by id (import wins) then by name (unified onto the existing tag, `tag_ids` remapped), so re-creating "prod" on another machine never duplicates. Orphaned tag ids are tolerated.

  ### Environment classification + production warning
  - Optional `environment` field per connection (`development`/`staging`/`production`), picked in the modal title bar, validated in the backend, preserved across duplicate/import.
  - Production identity: PROD badge on cards/rows, permanent red banner while the active connection is production, red ring on open sidebar entries.
  - Confirmation before any write on a production connection, with SQL preview and a per-connection "don't ask again" that lasts for the session. Detection is conservative: only SELECT/SHOW/DESCRIBE/PRAGMA/EXPLAIN-of-a-SELECT count as read-only — data-modifying CTEs, `EXPLAIN ANALYZE <write>` (which executes on Postgres) and unknown statement types (CALL, SET, …) all prompt. Guarded paths: editor runs (single + multi-statement), staged grid-edit commits, immediate cell edits, row insertion, notebook cells. Stacks with the existing dangerous-query guard.

  ## Tests
  - Rust: unit tests for tag CRUD helpers (validation, dedup, name-merge on import, detach on delete).
  - Vitest: 7 new tests for `isReadOnlyQuery`; full suite green (154 files / 2841 tests).
  - i18n: all 9 locales updated.

Would love some tests and feedback, especially on database other than sqlite and mariadb.


debba:
@pokertour can you resolve merge conflicts?

debba:
Thanks @pokertour for this contribution! This looks like a great starting point that we can expand on in future PRs. I pushed a small commit on top to repair the merge with main (two build errors, a few test mocks, and the ProductionBanner render that got lost in the conflict resolution).

Some ideas for follow-up PRs building on this foundation:

- **Custom environment colors**: let users pick the color for each environment tier, or even define custom tiers beyond development/staging/production.
- **Per-environment action policies**: configure which operations are allowed, blocked, or require confirmation based on the environment (e.g. block DROP/TRUNCATE in production, require confirmation for UPDATE/DELETE without a WHERE clause, allow everything in development).
- **Read-only mode**: an option to open production connections in read-only mode by default.
- **Environment visibility in more places**: connection tab colors, status bar badge, window title, and in the query confirmation dialogs.
- **Tag-based filtering and grouping**: filter the connections list by tag and optionally group by environment.

Merging, thanks again!

PR #478, opened 2026-07-14
feat(users): database user & privilege management (MySQL/MariaDB)

pokertour:
New "Users & Privileges" tab for drivers with the user_management capability:

- list server accounts (locked flag, degraded fallback to CURRENT_USER when mysql.user is not readable)
- create user with optional initial privileges on a chosen scope
- change password, drop user (confirmed)
- privilege editor: one card per scope (global / database / table) with checkboxes reflecting parsed SHOW GRANTS; checking grants, unchecking revokes (revoke-first so ALL PRIVILEGES -> subset works)
- raw SHOW GRANTS output kept for grants the editor cannot model (roles, column-level, proxy)

Extensible by design: privilege catalog comes from the driver, all seven trait methods are forwarded over JSON-RPC so external plugins (e.g. MongoDB) can opt in via capabilities.userManagement; PostgreSQL support is planned as a follow-up. SQL builders validate privileges against per-scope allowlists and escape literals per the server sql_mode; builders and grant parser are unit-tested.


debba:
Hi @pokertour can you resolve merge conflicts and provide translations for all languages?

debba (changes requested):
Thanks for the PR. The overall design is solid and I could verify everything the description claims: all eight commands are registered in `lib.rs`, the JSON-RPC forwarding is complete with the `userManagement` manifest alias, the allowlist plus escaping logic holds up (including the injection test case), passwords never hit the logs, and the degraded `CURRENT_USER()` fallback works as advertised. Unit tests pass (9/9 in `drivers::mysql::users`, 12/12 in `tabCleaner`), typecheck and eslint are clean. I also exercised the feature against a real MySQL 8.4.10 server with seeded accounts covering quoting edge cases, locked accounts, PROXY and dynamic-privilege grants, and a low-privilege user for the fallback path.

Two things need fixing before merge, plus one strong recommendation.

## Blockers

1. **`src/components/users/UserManagementView.tsx` contains a literal NUL byte** on line 46, inside the `scopeKey` template literal. Git treats the whole file as binary, so the main component of this PR (809 lines) shows up as "Binary file not shown" on GitHub and cannot be reviewed in the diff. Please replace the raw byte with the `\u0000` escape sequence; same semantics, and the file becomes text again.

2. **The i18n keys were only added to `en.json` and `fr.json`.** There are 9 locale files; de, es, it, ja, ru, tl and zh are missing the whole `userManagement` section. i18next falls back to English so nothing breaks, but per `.rules/general.md` the other locales should be filled in too.

## Strong recommendation

On MySQL 8, `SHOW GRANTS` for an account holding `ALL PRIVILEGES ON *.*` prints the expanded list of static privileges, which includes `CREATE TABLESPACE`, `CREATE ROLE` and `DROP ROLE`. None of these are in the allowlist in `users.rs`, and since a single unknown keyword makes the whole line unrepresentable, the global scope card renders completely empty for `root` and for any "GRANT ALL" admin. That is the most common account people will open this tab with. Adding those three keywords to `GLOBAL_ONLY_PRIVILEGES` makes the full admin line parse on MySQL 8.x; MariaDB is unaffected since it prints `ALL PRIVILEGES` unexpanded. Verified against 8.4.10. On the bright side, `GRANT SUPER` is still accepted on 8.4 (deprecated but functional), so the current catalog produces no errors, just the empty cards.

## Non-blocking notes

- `.rules/react.md` rule 2 (no synchronous setState in useEffect): `refreshUsers` is called from an effect and sets loading state synchronously (line 230), and the selected-account effect resets `extraScopes`/`addingScope` synchronously (lines 257-259). The second one could be avoided by keying the details panel on the selected account, the same trick you already used for `ScopeCard`.
- `.rules/rust.md` rule 4: the ~230 lines of tests in `users.rs` belong in a sibling `tests.rs` loaded with `#[cfg(test)] mod tests;`.
- The Tauri commands do not check the `user_management` capability server side; gating is UI only. The trait defaults return an error so it is not exploitable, but an explicit check would be nice defense in depth.
- Tiny one: the description says seven trait methods, it is actually eight counting the catalog.

The SQL builders and the grant parser are the strongest part of this PR, nicely tested. Happy to approve once the NUL byte and the locales are sorted.

PR #482, opened 2026-07-15
feat(grid): copy column values as list or SQL IN clause

### LINKED ISSUE #459: [Feat]: Copy Values of a Column
### Prerequisites

### 🚀 Feature Proposal

Copy values of a single column from the selected or visible rows.

Right now row or cell copy is available but not column copy.
Column name copy is available but not column value(s) copy.

### Motivation

It's very common use case to copy list of IDs and paste it in another table's filter query.

### Example

_No response_

pokertour:
Closes #459

## What

Adds two entries to the results grid context menus to copy the values of a single column:

- **Copy column values** — newline-separated, one value per line (`null` for NULL cells, same formatting as the existing cell copy).
- **Copy column values (IN clause)** — a ready-to-paste SQL list: numbers raw (`1, 2, 3`), strings quoted with `''` escaping (`'O''Brien'`), `NULL` for nulls — reusing the same value rendering as the existing *Copy as SQL INSERT*.

Both entries appear in:

- the **cell context menu** (applies to the clicked column), using the selected rows or **all visible rows when nothing is selected** — same fallback as *Copy selected row(s)*;
- the **column header context menu**, below the existing *Copy column name* entries.

## Implementation

- Two pure helpers in `src/utils/clipboard.ts` (`columnValuesToText`, `columnValuesToInClause`), one `copyColumnValues` callback in `DataGrid.tsx`, menu items, en/fr i18n keys.
- New `src/utils/clipboard.test.ts` with 4 vitest cases (newline join, null label, IN quoting, quote escaping).

Note: #481 also creates `clipboard.test.ts`; whichever merges second will have a trivial conflict there (independent test blocks, concatenate).

Manually tested on Windows: both menus, with/without row selection, columns containing NULLs and strings with quotes, pasting the IN list into a query.


pokertour:
Waiting clarification on the issue about the way to export data

debba:
@pokertour  Added my 5cents here: https://github.com/TabularisDB/tabularis/issues/459#issuecomment-5001735570

PR #487, opened 2026-07-15
Adds a setting that lets you run the statement under the cursor direc…

GabrielMalava:
Adds a setting that lets you run the statement under the cursor directly,
instead of showing the "Select Query to Execute" modal when a script has
multiple `;`-separated statements.

When enabled, pressing **Run** / `Ctrl+Enter` / `Ctrl+Cmd+F5` with no text
selected executes only the statement the caret is in — no picker. Selecting
text still runs the selected range, and the setting is **off by default**, so
existing behavior is unchanged.

Running one query out of many is a very common flow. The picker modal adds an
extra step every time. This makes "execute current statement" a one-keystroke
action, like most SQL editors (DataGrip/DBeaver).

## Scope / non-goals

- The context-menu "Execute Selection" action and the Run dropdown are
  unchanged.
- Only affects the no-selection + multiple-statements case.


debba:
Thanks for this — running one statement out of many without the picker is exactly the direction we want, and I agree with your idea of a dedicated setting so the existing behavior stays available.

Heads up though: this overlaps with #464, which is about to be merged and implements the same statement-at-cursor resolution as the *default* behavior — with a `findStatementAtOffset` helper in `sqlSplitter` that reuses the splitter's per-statement ranges, a visual highlight of the armed statement, and fixes for stale metadata / pagination when running single statements out of a multi-statement buffer. It also removes the picker modal from the Run/Explain paths, so this PR will conflict on `Editor.tsx` once it lands.

Rather than closing this, I'd love to keep its real value: the setting. Concretely, after #464 is merged, could you rebase and trim this down to:

- **Keep**: the config plumbing — the `AppConfig` field + `save_config` merge, `SettingsContext`, the `GeneralTab` toggle, and the i18n keys (please also add `ko.json`, it's currently missing from the locale updates).
- **Flip the semantics**: with #464 merged, run-statement-at-cursor becomes the default, so the toggle should read as "run statement under cursor" **ON by default**, and turning it **OFF** restores the old query-picker modal (`QuerySelectionModal` was intentionally left in the codebase for exactly this).
- **Drop**: `statementAtOffset` in `utils/sql.ts` and its tests — #464's `findStatementAtOffset` in `sqlSplitter` covers it.

Happy to help with the rebase if you'd like. Thanks again!

PR #489, opened 2026-07-16
fix(data-grid): open dedicated editor for array cells on double-click

Davydhh:
## Problem

Double-clicking a cell whose value is an array (e.g. a Postgres `uuid[]` /
`text[]` column) opened the inline `<textarea>` editor. That editor renders
`String(array)`, cramming the values into a comma-joined string inside a
~120px-tall scrollable box — awkward to read and edit.

The grid was also internally inconsistent: array cells already *render* as
JSON cells (with the JSON affordance), but *editing* them fell through to the
plain-text path.

## Fix

Route array cells through the same dedicated JSON viewer window that
`json`/`jsonb` columns already use on double-click.


debba:
@Davydhh  Can you resolve merge conflicts?
After that I will merge it.

Davydhh:
@debba done

PR #495, opened 2026-07-18
feat(connections): forward raw connection URIs to plugin drivers

### LINKED ISSUE #494: [Bug]: external plugin drivers cannot receive a full connection URI (mongodb+srv rejected, then decomposed)
### Describe the bug

External plugin drivers cannot receive a full connection URI. Two independent gaps in `src/utils/connectionStringParser.ts` combine to make SRV-based drivers (MongoDB Atlas, and any future driver whose scheme carries semantics) unusable.

**Gap 1 — the protocol is rejected**

`buildProtocolRegistry` only registers two things per driver: `normalizeProtocol(driver.id)` and the protocol extracted from `capabilities.connection_string_example`. A plugin whose id is `mongodb-atlas` therefore only registers `mongodb-atlas://`. Pasting a real Atlas URI fails at `connectionStringParser.ts:171`:

```
Unsupported database driver: mongodb+srv. Supported: mongodb-atlas, mysql, postgres, postgresql, sqlite
```

`PROTOCOL_ALIAS_GROUPS` is a hardcoded list (`postgres/postgresql`, `mysql/mariadb`, `sqlite/sqlite3`), so a plugin cannot register an alias pair such as `mongodb` + `mongodb+srv` even by declaring an example — the example only registers one of the two.

**Gap 2 — the URI is destroyed even when the protocol is accepted**

Once parsing succeeds, `parseConnectionString` decomposes the URI into discrete fields (`connectionStringParser.ts:200-228`) and `toConnectionParams` forwards only those:

```ts
return {
  driver: parsed.driver,
  host: parsed.host,
  port: parsed.port,
  username: parsed.username,
  password: parsed.password,
  database: parsed.database,
};
```

The scheme and the entire query string are discarded. For `mongodb+srv://` this is not recoverable:

- `mongodb+srv` mandates a DNS SRV lookup to discover the replica-set members. A bare `hostname` with no port is not a valid substitute — the driver has no way to know it should resolve SRV records.
- Atlas URIs carry `retryWrites`, `w=majority`, `authSource`, `replicaSet` and `appName` in the query string. All are dropped.
- `url.port` is empty for SRV URIs, so `port` arrives as `undefined` and the plugin falls back to `default_port` (27017), which is wrong for a sharded Atlas cluster.

The observable result is that a connection can appear to be established while database enumeration returns nothing, because the reconstructed connection no longer points at the real cluster.

Grepping current `main` (39f5ab2b) confirms there is no path that preserves the original string — there is no `connection_uri` / `full_uri` field anywhere in `src-tauri/src` or `src/utils`.

### To Reproduce

1. Install an external plugin driver that accepts a full URI. I used a MongoDB Atlas plugin whose `manifest.json` declares `"id": "mongodb-atlas"`, `"default_port": 27017` and no `connection_string_example`.
2. Open **New Connection** and paste a standard Atlas connection string:
   ```
   mongodb+srv://user:pass@cluster0.xxxxx.mongodb.net/mydb?retryWrites=true&w=majority&appName=Cluster0
   ```
3. The modal rejects it with `Unsupported database driver: mongodb+srv`.
4. Work around gap 1 by adding `"connection_string_example": "mongodb+srv://user:pass@host/db"` to the plugin manifest so the protocol registers.
5. Paste the same URI again. It now parses, but the plugin receives only `host=cluster0.xxxxx.mongodb.net`, `port=undefined`, `database=mydb` — no SRV scheme, no query parameters.
6. The connection does not resolve to the real cluster and databases fail to load.

### OS Version

macOS 26.4 (25E246), arm64

### Tabularis Version

v0.14.0 (also verified against `main` @ 39f5ab2b)

### Relevant Log Output

```shell
Unsupported database driver: mongodb+srv. Supported: mongodb-atlas, mysql, postgres, postgresql, sqlite
```

---

### Suggested direction

The core need is that a plugin should be able to declare *"I take the raw connection string, do not decompose it"*, and the host should carry that string through to the driver over JSON-RPC.

Roughly:

1. A capability flag (e.g. `connection_string_passthrough`) that makes `parseConnectionString` skip decomposition and preserve the original URI verbatim.
2. A `connection_uri` field on the connection model, forwarded to the external plugin on connect.
3. Allow plugins to declare multiple protocols rather than deriving a single one from `driver.id` plus one example, so alias pairs like `mongodb` / `mongodb+srv` work.

Because the URI embeds credentials, it should be treated as a secret: stored in the OS keychain rather than in `connections.json`, with the cached secret rolled back if persistence fails. A transient URI is acceptable for **Test Connection**, but a saved connection should require keychain availability.

I have a prototype of this against the v0.14.0 codebase. It predates the capability-driven protocol registry that `main` now has, so it needs to be reworked to fit the current architecture, and I have not yet verified end-to-end that databases enumerate correctly with it. I am happy to open a PR once it is rebased and properly verified — let me know if the direction above is one you would accept.

Robbyfuu:
Closes #494.

Opened as a draft for one reason, stated up front: **the Rust tests have never executed in my environment.** `src-tauri/tests/test_ca.pem` is tracked in git but absent from my worktree, and `pool_manager_tests.rs:466` pulls it in with `include_bytes!`, so `cargo test --lib` fails to compile before reaching anything. I wrote 13 Rust tests that compile cleanly under `cargo check --tests` but have never run. I am opening this as a draft specifically so CI executes them. Everything else below was verified locally.

## The problem

Pasting a real MongoDB Atlas connection string fails. Three separate gaps, all on the connection-string import path:

**The protocol is rejected.** `buildProtocolRegistry` derives protocols from only two sources per driver — `normalizeProtocol(driver.id)` and the scheme of `connection_string_example`. `PROTOCOL_ALIAS_GROUPS` is a hardcoded const. A plugin has no way to register a second scheme, so:

```
Unsupported database driver: mongodb+srv. Supported: mariadb, mongodb,
mongodb-atlas, mysql, postgres, postgresql, redis-rust, sqlite, sqlite3
```

**The URI is destroyed even when accepted.** `parseConnectionString` decomposes the URL and `toConnectionParams` forwards only host/port/username/password/database. The scheme and the entire query string are discarded. For `mongodb+srv://` that cannot be recovered: the scheme mandates a DNS seedlist lookup, and Atlas cluster hostnames publish SRV records but **no A record**, so the decomposed `host:27017` fails at resolution:

```
Server selection timeout: No available servers.
Servers: [ { Address: cluster0.xxxxx.mongodb.net:27017, Type: Unknown,
  Error: I/O error: failed to lookup address information } ]
```

Query parameters (`tls`, `retryWrites`, `w`, `authSource`, `replicaSet`, `appName`) are lost with it.

**A database path is mandatory.** Real Atlas URIs routinely omit it — `mongodb+srv://user:pass@cluster0.xxxxx.mongodb.net/?tls=true&retryWrites=true&w=majority&appName=Cluster0` — but the parser rejects that with "Database name is required in connection string".

## Approach

Two new capabilities: `connection_uri` (bool) to declare that a driver takes the string verbatim, and `connection_uri_schemes` (string[]) to register additional schemes. Both accept camelCase aliases, matching the existing convention for `connection_string` / `connectionString`.

The URI travels on `ConnectionParams` rather than through plugin settings. That is the load-bearing decision, so it is worth stating why: plugin settings are pushed once via the `initialize` RPC in `RpcDriver::new`, which makes them global per plugin — an Atlas cluster and a local `mongodb://localhost` could not coexist. `ConnectionParams` is passed to every driver method, so putting the URI there makes it per-connection. It also means `plugins/driver.rs` needed no production change: serde already serializes the whole struct into each request.

Both new fields are `Option` + `#[serde(default)]` + `skip_serializing_if`, so existing stored connections deserialize unchanged and re-serializing does not introduce the keys. There is a test asserting the serialized output of legacy params contains no `connection_uri` substring. Both new capability fields are `#[serde(default)]`, so existing plugin manifests keep parsing.

Non-passthrough parsing is untouched: passthrough is a property of the resolved driver and branches before the existing local/remote split. The database requirement still applies to every driver that does not opt in.

Note that fixing only the protocol rejection would have been strictly worse than the status quo — the URI would parse, get decomposed, and persist a connection that fails at DNS instead of showing a clear validation error. Both gaps land in the same function for that reason.

## Security

The URI embeds credentials in cleartext inside the string, so it is treated as a secret throughout:

- Stored in the OS keychain under `{id}:connection_uri`, never in `connections.json`. `params_for_persistence` strips it on every write path and leaves only a `connection_uri_in_keychain` marker.
- Saving or updating requires the keychain and fails loudly rather than silently downgrading to plaintext. The frontend mirrors the check, but the backend enforces it independently.
- Test Connection may use an ephemeral in-memory URI with no keychain requirement.
- If `connections.json` persistence fails after the secret was written, both the keychain entry and the in-memory cache roll back. If the rollback itself fails, both errors are reported and the cache entry is dropped rather than left disagreeing with the keychain.
- A marker pointing at a missing keychain entry fails closed with a clear error rather than silently connecting with the decomposed fields.
- Declared schemes cannot displace a protocol another driver already owns. Drivers reach the parser sorted by id, so without that guard a plugin could claim `postgres` and — because passthrough hands the string over verbatim — capture a URI the user pasted intending the built-in driver. This mirrors the guard the alias-expansion pass already had.

Import, export and the MCP surface were each brought onto the same terms: `apply_export_payload` validates and routes an inline URI to the keychain instead of persisting it, `export_connections_payload` resolves it from the keychain when `include_secrets` is set so a restored backup is not left with a marker and no secret, and `resolve_db_params` restores it so an MCP query reaches the driver with the same params the app would send.

## Verification

Run locally against this branch:

- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `pnpm test --run tests/utils/connectionStringParser.test.ts` — 47 passed
- `cd src-tauri && cargo check` — compiles, 2 warnings, both pre-existing and unrelated (`ssh_tunnel.rs:91`, `plugins/driver.rs:34`)

The full JS suite reports 33 failures across `SettingsProvider`, `ThemeProvider` and `useSidebarResize`, all on `localStorage.clear()`. These are pre-existing on `main` — I confirmed by stashing this entire change and reproducing the identical 3 files / 33 tests on a clean tree. None of those files are touched here.

The regression test guarding scheme shadowing was verified to be meaningful rather than decorative: removing the guard makes it fail, restoring it makes it pass.

## Known gaps

- **The Rust tests have not run.** 13 tests across `commands.rs`, `models_tests.rs` and `plugins/driver.rs` compile but are unexecuted, for the `test_ca.pem` reason above. This is the reason for the draft status.
- **No plugin exercises this end to end.** Nothing in this repo declares `connection_uri: true`, so the passthrough path has not been driven against a live cluster. The SRV resolution that motivates the change is the plugin's responsibility.
- **The i18n strings need native review.** One new key across 10 locales. Each quotes that locale's existing `saveKeychain` label verbatim and uses its established keychain terminology, but they are not native-speaker reviewed.
- **The modal changes are covered by typecheck and lint only** — no component test.

Out of scope deliberately: no UI affordance marking a connection as URI-backed, no reveal-stored-URI flow, and no plugin SDK changes under `packages/` — external plugin authors will need the capability documented there before they can adopt it.

Happy to adjust the capability naming or split this if you would rather see the parser and the secret-handling land separately.


Robbyfuu:
Update: the "no plugin exercises this end to end" gap from the PR description is now closed — I drove the full loop against a live Atlas cluster with a real plugin, and it surfaced one bug in this PR, fixed in the new commit.

**Setup.** A MongoDB plugin declaring the new capabilities (`connection_uri: true`, `connection_uri_schemes: ["mongodb", "mongodb+srv"]`) and consuming `connection_uri` verbatim in its connect path: https://github.com/Robbyfuu/tabularis-mongodb-plugin/tree/feat/mongodb-atlas-full-uri. Host built from this branch, production Tauri bundle, macOS arm64.

**What worked immediately:**
- Pasting a real `mongodb+srv://.../?tls=true&retryWrites=true&w=majority&appName=...` URI (no database path) parses, shows the derived host/username for display, and keeps the URI verbatim — the old `Unsupported database driver: mongodb+srv` error is gone.
- Save refused while "Save passwords in Keychain" was off, with the new message; with it on, the connection saved and databases enumerated during creation.
- On-disk state after save is exactly the contract: `connections.json` carries `connection_uri_in_keychain: true` and no URI; the secret exists in the OS keychain under `{id}:connection_uri`; SRV resolution works (the driver talks to the individual `ac-...-shard-00-0N` replica-set members, which only SRV can discover).

**The bug this found (fixed in `5e13eee1`):** reconnecting to a saved connection failed with the decomposed `host:27017` DNS error even though creation-time enumeration had worked. The frontend's reconnect flow fetches `get_connections` — on-disk params, URI stripped by design — and passes those params to `test_connection` / `list_databases`. Both commands already restore the *password* from the keychain when the request params lack one and a `connection_id` is present; the URI needed the identical treatment and had no restore path. Same asymmetry class as the MCP gap already fixed in this branch — the third instance, and the reviews missed it because none of them enumerated frontend-initiated flows that pass stale params by value. Worth keeping in mind for any future field that lives in the keychain rather than `connections.json`.

After the fix: reconnect works, both databases enumerate, collections list, schema inference and data browsing all return real documents from the cluster.

(Plugin-side, for anyone wiring their own driver to this: the host sends `database` as an array for multi-database drivers, the selected database arrives in the request-level `schema` field, and browsing queries quote identifiers — `SELECT * FROM "collection"`. All three tripped the plugin before it handled them.)

debba:
Thanks for the exceptionally well-documented PR, reviewed locally on the branch. TL;DR: **the Rust tests you couldn't run are green here, and I verified the passthrough contract end to end. I think this can come out of draft.**

## Verification

- **Rust suite (your draft blocker):** `test_ca.pem` is present in my worktree, so `cargo test --lib` compiles and runs — **1001 passed, 0 failed**, including all 13 new tests (`commands.rs`, `models_tests.rs`, `plugins/driver.rs`).
- `pnpm vitest run tests/utils/connectionStringParser.test.ts` — 47 passed, as stated.
- `pnpm typecheck` / `pnpm lint` — clean.
- i18n: the new key is present in all 10 locales and the en string quotes the existing `saveKeychain` label verbatim. The Italian one reads natively to me. 👍

## End-to-end passthrough proof

To close the "no plugin exercises this" gap I drove your `tabularis-mongodb-plugin` v0.2.0 binary (from #496) directly over JSON-RPC stdio with a `test_connection` request shaped exactly like the one this PR sends:

- with `connection_uri` pointing at `uri-host.invalid` and the decomposed `host` field pointing at `decomposed-host.invalid`, the plugin dialed **`uri-host.invalid`** — it consumed the URI verbatim;
- without `connection_uri`, it dialed the decomposed host — fallback intact.

Together with the new `rpc_driver_forwards_*` tests, both halves of the contract are demonstrated. The only thing still unexercised is live SRV resolution against a real Atlas cluster, which I'll do as part of manual testing with #496 installed.

## Non-blocking notes

1. **MCP and export paths fail open where the app path fails closed.** `resolve_db_params` (`mcp/mod.rs`) and `export_connections_payload` use `if let Ok(Some(uri))`, silently swallowing keychain errors — via MCP that surfaces as a confusing driver-level failure instead of the clear error the in-app path gives, and on export with `include_secrets` it can produce exactly the marker-without-secret payload the PR set out to prevent. It mirrors the pre-existing password handling on the adjacent lines, so it's consistent — fine as a follow-up rather than in this PR.
2. The `"Stored connection URI is unavailable in the OS keychain"` branch in `restore_runtime_connection_uri` is effectively unreachable (nothing writes `CacheEntry::Absent` for URIs), so a deleted keychain entry surfaces the generic "Failed to read…" message instead. Cosmetic.
3. I traced the `update_connection` preserve/clear/overwrite logic against all three frontend paths (field untouched → preserved, cleared → deleted, re-pasted → overwritten) — it's sound, and the driver-switch guard dropping the URI is a nice touch.

The keychain-or-nothing rule, the rollback on failed persistence, and the scheme-shadowing guard all check out (I confirmed the shadowing test fails when the guard is removed, as you said).

@NewtTheWolf you've been closest to the registry/plugin surface lately — I'd value your opinion before merging, particularly on the capability naming (`connection_uri` + `connection_uri_schemes` vs. something like a single `connection_uri: { passthrough, schemes }` object) and on whether the per-connection-params transport (instead of plugin settings) fits where the plugin API is heading. Roberto offered to adjust naming or split the parser from the secret handling if we prefer.

debba:
Hi @Robbyfuu can you fix merge conflicts?
I will proceed on merge after that

NewtTheWolf:
Thanks for the ping, @debba.

**1. Naming — keep it flat.** `connection_uri` + `connection_uri_schemes` mirrors `connection_string` + `connection_string_example` three lines up in the same struct and keeps the camelCase alias trivial. A nested object would be the only nested capability in `DriverCapabilities`. @Robbyfuu's manifest already writes it flat, unprompted.

**2. Transport — params, not plugin settings.** Settings are a plaintext `HashMap<String, Value>` in `config.json`, one instance per plugin. A per-connection secret belongs where the password lives. Via settings you'd get one URI shared across every Atlas connection, in cleartext.

**Registry: nothing needed** — capabilities come from the bundle's `.tabularium`, not from us. (Checking it did turn up that the `driver` kind schema had drifted behind `DriverCapabilities`; patched separately, unrelated to merging this.)

@Robbyfuu I resolved the conflicts on your branch directly — merge commit, no force push, your commits untouched. All three were against the IAM-auth work that landed upstream. The `mcp/mod.rs` one also closes the Kilo finding: the URI restore is now out of the `save_in_keychain` block. Merged tree is green — `cargo test --lib` 1051 passed, `pnpm vitest run` 3405 passed, typecheck and lint clean.

One non-blocking note: `resolve_db_params` allocates a fresh `CredentialCache` per read, so nothing is actually cached in the MCP path. Fold it into the follow-up with your fail-open `if let Ok(Some(..))` point.

**Approve.** Putting the restore in `find_connection_by_id` covers the ~30 commands that resolve params from disk, leaving only the four that take params from the frontend — all four handled. Root cause fixed once, not patched per call site.

PR #499, opened 2026-07-20
feat(indexes): render functional/expression indexes across drivers

Davydhh:
## Problem

Functional/expression indexes (e.g. `CREATE INDEX ON users (lower(email))`)
were not rendered. Each driver's index query dropped expression columns —
Postgres via an inner join on `pg_attribute` (expression columns have
`attnum = 0` and match no row), MySQL and SQLite because they only read the
plain column name — so mixed indexes lost their expression columns and
all-expression indexes disappeared from the sidebar entirely.

## Fix

Recover the expression text in every driver (Postgres `pg_get_indexdef`,
MySQL `information_schema.STATISTICS.EXPRESSION`, SQLite's `CREATE INDEX` DDL)
and add an `is_expression` flag so DDL export emits expressions raw instead of
identifier-quoting them.

## Notes

- MySQL: `EXPRESSION` exists in 8.0.13+; the driver probes for it and falls
  back gracefully on MariaDB / older MySQL.
- Verified against real Postgres 18, MySQL 8.0, and SQLite (added unit +
  integration tests for the SQLite path and the DDL export).


debba:
Reviewed the changes across all three drivers and ran the tests locally: 21 sqlite driver tests and 65 sqlGenerator tests, all green. No blockers, a few minor notes below.

**What works well**

- Postgres: the rewrite with `unnest ... WITH ORDINALITY` plus a LEFT JOIN is the right approach. The old inner join on `pg_attribute` dropped rows with `attnum = 0`, which hid expression-only indexes entirely. It also fixes a latent bug: `array_position(ix.indkey, a.attnum)` returned the first position for a column that appears twice in an index, while `k.n` is always correct.
- MySQL: probing `information_schema.COLUMNS` for the `EXPRESSION` column (MySQL 8.0.13+) with a fallback for MariaDB and older MySQL is pragmatic, and `Err(_) => false` degrades cleanly. Column indices in the two query variants line up with the code reading them.
- SQLite: `parse_sqlite_index_columns` handles nested parens, all four quoting styles, commas inside function calls, partial index WHERE clauses, and even parens inside quoted table names, all covered by tests. The new `sqlite_master` query uses a bind parameter instead of `format!`, nice.
- Frontend: leaving expression columns unquoted in `renderColumns` is correct, quoting them would turn the expression into a literal identifier. The other consumer (`groupIndexes` for the sidebar) just displays the expression text, which is what you want.

**Minor notes, none blocking**

1. The MySQL probe runs on every `get_indexes` call, so each table expanded in the sidebar costs one extra round trip. The result only depends on the server version, so it could be cached per pool. Micro-optimization, up to you whether it's worth it.
2. If a MySQL STATISTICS row ever came back with both `COLUMN_NAME` and `EXPRESSION` as NULL, `column_name` would end up as an empty string with `is_expression: false`. This shouldn't happen in practice, just noting it for completeness.
3. `sqlite_master.sql` preserves the user's original DDL, so a comment inside the column list (`CREATE INDEX i ON t (a /* note */, lower(b))`) would end up in the parsed token. Very rare and purely cosmetic, I wouldn't bother fixing it now.
4. `seq_in_index` semantics differ across drivers (SQLite 0-based, MySQL/Postgres 1-based). This predates the branch and is harmless since `generateIndexStatements` only uses it for ordering. Same goes for the `format!` in `PRAGMA index_list` and for Postgres INCLUDE columns being rendered as key columns, both pre-existing.

Good to merge as is. The only thing I'd actually consider is caching the MySQL probe if `get_indexes` gets called frequently.

PR #510, opened 2026-07-22
feature: add toggleable right sidebar with row editor panel

aesslinger:
Replaces the fixed-position "Edit Row" overlay with a layout-integrated right sidebar drawer that participates in the flexbox layout. The sidebar follows row selection, is resizable, and introduces a generic panel infrastructure for future extension.

## Motivation

The current row editor opens as a `position: fixed` overlay that:
- Covers content rather than pushing it aside
- Is pinned to a single row and doesn't respond to selection changes
- Requires reopening via context menu each time a different row is needed

This PR introduces a proper right sidebar that behaves like the left Explorer sidebar — a first-class layout citizen with keyboard shortcut, resize persistence, and selection tracking.

## What's New

**Right Sidebar Infrastructure**
- New `RightSidebarContext` with provider and consumer hook
- `RightSidebar` layout component rendered in `MainLayout` as a flex sibling
- `useRightSidebarResize` hook (drag handle, width persisted to localStorage)
- Keyboard shortcut: `⌘+Shift+B` / `Ctrl+Shift+B`
- Generic panel system (`activePanel` type supports future panels)

**Row Editor Panel**
- Extracted from `RowEditorSidebar` → `RowEditorPanel` (pure content, no positioning)
- Pin toggle button — locks sidebar to a specific row (off by default = follows selection)
- Sidebar auto-updates when a single row is selected
- `onChangeRef` pattern prevents stale closures when data refreshes

**Settings (General → Data Editor)**
- *Row editor follows selection* — toggle (default: on)
- *Cell double-click action* — "Inline edit" / "Open in sidebar" / "Both"

**Double-Click on `#` Column**
- Opens the full row in the sidebar (respects readonly guard)

## Architecture Decisions

| Decision | Rationale |
|----------|-----------|
| `onChangeRef` pattern instead of storing closures in state | Prevents stale closures after data refresh / row reorder |
| Stable method refs (`openRowEditor`, `updateRowEditorData`) via `useCallback([])` | Avoids invalidating `rowCtx` useMemo → prevents full grid re-render on sidebar state changes |
| DataGrid unmount cleanup (`close()`) | Prevents stale sidebar state persisting across route navigation |
| `close()` clears all state (`activePanel`, `rowEditorData`, `onChangeRef`) | Prevents stale data appearing when sidebar is re-opened via toggle |
| Context wraps inside `ConnectionLayoutProvider` | Sidebar has access to connection state |

## Files

| Category | Files |
|----------|-------|
| **New** | `RightSidebarContext.ts`, `RightSidebarProvider.tsx`, `useRightSidebar.ts`, `useRightSidebarResize.ts`, `RightSidebar.tsx`, `RowEditorPanel.tsx` |
| **Modified** | `App.tsx`, `MainLayout.tsx`, `DataGrid.tsx`, `DataGridRow.tsx`, `GeneralTab.tsx`, `SettingsContext.ts`, `useGlobalShortcuts.ts`, `shortcuts.json`, all 10 locale files |
| **Deleted** | `RowEditorSidebar.tsx` (replaced by `RowEditorPanel`) |

## Testing

- [x] `pnpm tsc --noEmit` — no errors
- [x] `pnpm run lint` — no errors or warnings
- [x] `pnpm tauri build` — compiles successfully
- [x] Manual: right-click → "Open Sidebar Editor" → sidebar slides in, pushes content left
- [x] Manual: click different rows by `#` column → sidebar follows
- [x] Manual: edit a field, click another row → edits preserved in pending changes
- [ ] Manual: ⌘+Shift+B toggles sidebar
- [x] Manual: drag resize handle → width persists on reload
- [x] Manual: pin button locks to row, unpin re-enables following
- [x] Manual: double-click `#` column → opens full row
- [ ] Manual: double-click data cell with "sidebar" setting → opens in sidebar
- [x] Manual: settings "Cell double-click action" controls behavior
- [x] Manual: close sidebar, run new query, toggle → no stale data
- [ ] Manual: navigate to Settings with sidebar open → sidebar closes cleanly
- [x] Manual: read-only query result → double-click `#` does nothing

## Breaking Changes

None — the `RowEditorSidebar` component was only used internally by `DataGrid`. All external behavior (context menu "Open Sidebar Editor", blob cell buttons, geometry SQL functions) continues to work identically.


aesslinger:
@debba this is a fully unsolicited PR.  Just something I was thinking about as I use this awesome tool.  Curious for feedback and if the idea here isn't in the vision and this gets discarded then there are no hard feelings on my end.

debba:
Hi!
Looks very interesting! It's not clear to me how it works pin row, is it WIP?

aesslinger:
@debba previously when viewing data you only had a static right panel that would open and be pinned to a single record.  Now you can toggle from row to row and the right panel changes as you go, unless you pin it. It works in my testing. So, not really a WIP, but I kept this as a Draft PR in case it isn't a feature that is desired.

debba:
I really love this feature, I understand how it works the pin feature, but didn’t work to me. But I will recheck soon.
Btw very good work, please go ahead 😀

debba:
Hi @aesslinger,

First of all, great work! I really like the direction of these changes. I only have a few pieces of feedback from the manual testing I did:

* **Linux:** the <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>B</kbd> shortcut doesn't seem to work. I expected that, after selecting a cell or a row and pressing the shortcut, the modal would open. Am I misunderstanding how it's supposed to work?
* **Double-clicking a data cell** with the **"Sidebar"** setting correctly opens the sidebar. However, with the **"Both"** mode, the behavior feels a bit inconsistent: sometimes the focus goes to the modal input, while other times it starts inline editing instead. Also, for the **"Open in Sidebar"** mode, automatically focusing the edited column would be a nice quality-of-life improvement.
* **Navigating to Settings** with the sidebar open: on my side, the sidebar remains open even after navigating to **Settings** or **Connections**. I expected it to close cleanly when leaving the data view.

aesslinger:
@debba I'm going to look at each issue individually.  For the first one:

Good catch — the shortcut was only toggling the sidebar's visibility state without populating it with row data. If the sidebar had never been opened via context menu or double-click first, pressing Ctrl+Shift+B would toggle isOpen to true but the component would render nothing (due to a !activePanel guard).

The fix moves the toggle event handling from the provider (which has no knowledge of the current grid state) into the DataGrid component itself.

Now the shortcut:

- Sidebar open → closes it
- Sidebar closed + a single row selected → opens with that row's data
- Sidebar closed + no selection or readonly context → does nothing

This matches the expected behavior of select a row, press the shortcut, see the editor.

aesslinger:
Two issues addressed:

1. "Both" mode focus fighting

When cellDoubleClickAction was set to "Both", both the inline editor and the sidebar were competing for focus — the inline input would grab it immediately, then 150ms later the sidebar's scroll-to-field effect would steal it back (or vice versa depending on timing).

Fix: In "Both" mode, the sidebar opens to show the full row but does not pass a focusField, so it won't attempt to grab focus away from the inline editor. The user gets the inline input focused (expected) with the sidebar visible as a companion view.

2. "Sidebar" mode: repeated same-column double-click didn't re-focus

If you double-clicked the "name" column on row 1, then double-clicked the "name" column on row 2, the sidebar would scroll to the row but not re-focus the field — because focusField was still "name" both times, so React's useEffect didn't re-fire.

Fix: Added a focusTrigger counter that increments on every openInSidebar call with a focusField. The panel's focus effect now depends on [focusField, focusTrigger], guaranteeing it re-fires even for the same column on a different row.

aesslinger:
The sidebar wasn't closing on navigation because the unmount cleanup had a stale closure. The effect used [] deps (fire once on mount), meaning rightSidebar.isOpen was captured as false at mount time. When the DataGrid later unmounted (user navigated to Settings/Connections), the cleanup ran but still saw isOpen === false and skipped the close() call.

Fix: Store rightSidebar in a ref that's updated on every render. The cleanup now reads rightSidebarRef.current — which always has the live value at unmount time — and correctly calls close() when the sidebar is open.

debba:
@aesslinger I really like this a lot! The issues now seem fixed on my end. If you're not testing anything else, this is ready to merge as far as I'm concerned . I'm just waiting for your confirmation.

Also, features like this one are really interesting, please keep proposing them, we love them! :)

aesslinger:
@debba Thanks, I am sure there will be more functionality over time that can take advantage of the right sidebar.  I just don't know the project that well.  All this stems from me as a new user and seeing things I like to have/use vs what is available.  I'll never claim to have the best ideas or the best way to implement them all the time though! I just appreciate that you are responsive to ideas and willing to work with folks like me.

So, from my end, there is nothing more for this PR.  I'll keep watching out for other things I potentially run into.  You can also feel free to poke something my way and I can always try to analyze or even fix/add.

debba:
Sure, if you want you can reach us on Discord, if you'd like to propose something before implementing 😀
Btw thanks a lot for your efforts!

PR #513, opened 2026-07-23
fix(import): scope import to selected database and prevent double exe…

### LINKED ISSUE #512: [Bug]: Running a SQL file launches concurrent duplicate imports and interleaves DDL
## Summary

Running a SQL file once can launch multiple concurrent `import_database`
commands. Their DDL statements interleave, leaving a partially modified
database and producing a rapid sequence of unrelated/repeated errors.

Copying the SQL into a console and using "Run all" works correctly.

## Environment

- Tabularis: 0.16.0 (from AUR)
- OS: Arch Linux
- Database: fresh `mariadb:lts` Docker container with an empty database
- Import format: plain `.sql`

## Steps to reproduce

1. Start a fresh `mariadb:lts` container with an empty database.
2. Connect to it in Tabularis.
3. Select **Run SQL File...**.
4. Select the attached/reproducer SQL file [bug.sql](https://github.com/user-attachments/files/30286738/bug.sql).
5. Confirm the import once.

The file drops and recreates these related tables in dependency order:
`pays`, `regions`, `appellations`, `cepages`, `producteurs`, `vins`,
and `vin_cepages`.

## Expected behavior

Exactly one import job runs. Statements execute sequentially, and either:

- the script completes successfully, or
- execution stops at the first error and reports that error once.

## Actual behavior

A picture is worth a thousand words, so here's a video:
https://github.com/user-attachments/assets/f5423620-6876-455d-be48-864f4808da27

The alert changes rapidly between failures from different statement numbers.
In roughly two seconds we cycled across the following errors:

- statement 11: errno 150, creating `regions`
- statement 15: errno 150, creating `vins`
- statement 12: table `appellations` already exists
- statement 10: table `pays` already exists
- statement 14: errno 150, creating `producteurs`
- statement 11: errno 150, creating `regions` again
- statement 13: table `cepages` already exists

The database is left partially populated.

## Technical findings

A single backend import cannot produce this sequence:
`execute_statements_streaming!` awaits every statement and propagates the first
error with `?`, so one invocation is serial and fail-fast.

The frontend appears able to start the command repeatedly:

- `ImportDatabaseModal` starts the import from an effect depending on
  `startImport`.
- `startImport` depends on `onClose` and `onSuccess`.
- `ExplorerSidebar` supplies those props as inline callbacks, giving them new
  identities when the parent renders.
- The effect cleanup removes only the event listener; it does not prevent or
  cancel the already-started command.
- `React.StrictMode` can trigger the initial duplicate in development.
- Reporting an error through the alert provider causes another render, which
  can restart the effect and continue the cascade.
- The backend cancellation registry allows multiple import task handles under
  the same connection/import key, so duplicate commands run concurrently.

MariaDB DDL implicitly commits, so concurrent imports can interleave their
`DROP TABLE` and `CREATE TABLE` operations despite each task beginning a
transaction. This explains both the malformed-FK and already-exists errors.

## Suggested safeguards

- Make modal auto-start one-shot and StrictMode-safe.
- Stabilize callbacks or avoid using callback identity as an import-start
  dependency.
- Reject a second active import for the same connection/database in the
  backend.
- Scope progress events to an import/job ID so concurrent operations cannot
  overwrite each other's UI state.

gcapellib:
Close #512 

## Problem

Two bugs in the SQL file import, on MySQL/MariaDB connections exposing several databases:

1. **Wrong target database.** Right-clicking a database → *Run SQL file* runs the file against the connection's **primary** database, not the one that was right-clicked. The error makes it explicit:

The modal receives the target database name and displays it in its title, but never forwards it to the backend.

2. **Import runs twice.** The modal's `useEffect` re-fires `startImport` (React strict-mode double mount + `startImport` recreated when its deps change), executing the whole dump twice. On a fresh database this surfaces as `Table 'x' already exists` on the very first statement.

## Fix

- `import_database`: add a `database: Option<String>` parameter and override `params.database` before opening the pool — a line-for-line mirror of the fix already present in `dump_database` (with the same explanatory comment).
- `ImportDatabaseModal`: forward the target database (kept separate from the display label, which can fall back to a generic string), and guard the effect with a `useRef` so the import runs once per opening.

## Testing

Verified on MariaDB (`mariadb:lts`, Docker), Linux. Before: import lands in the wrong DB and/or fails with duplicate-table errors. After: importing a multi-table dump into the right-clicked database succeeds in one pass (7 tables + data loaded).

`npx tsc --noEmit` passes. **Not tested on PostgreSQL** — the change touches the shared `import_database` path, so a maintainer check on the PG import flow (where `schema` and `database` now coexist) would be welcome.

### DIFF
--- a/src-tauri/src/dump_commands.rs
+++ b/src-tauri/src/dump_commands.rs
@@ -443,11 +443,20 @@ pub async fn import_database<R: Runtime>(
     connection_id: String,
     file_path: String,
     schema: Option<String>,
+    database: Option<String>,
 ) -> Result<(), String> {
     let saved_conn = find_connection_by_id(&app, &connection_id)?;
     let expanded_params = expand_ssh_connection_params(&app, &saved_conn.params).await?;
     let expanded_params = expand_k8s_connection_params(&app, &expanded_params).await?;
-    let params = resolve_connection_params_with_id(&expanded_params, &connection_id)?;
+    let mut params = resolve_connection_params_with_id(&expanded_params, &connection_id)?;
+    // Scope the import to the selected database on connections that expose multiple
+    // databases (e.g. MySQL/MariaDB). Without this the connection pool stays bound
+    // to the primary database, so every statement in the dump file is executed
+    // against the wrong database. Mirrors the same fix already applied to
+    // `dump_database`.
+    if let Some(db) = database {
+        params.database = crate::models::DatabaseSelection::Single(db);
+    }
     let driver = saved_conn.params.driver.clone();
     let pg_schema = schema.unwrap_or_else(|| "public".to_string());
     let app_handle = app.clone();
--- a/src/components/layout/ExplorerSidebar.tsx
+++ b/src/components/layout/ExplorerSidebar.tsx
@@ -2561,6 +2561,7 @@ export const ExplorerSidebar = ({ sidebarWidth, startResize, onCollapse, sidebar
           onClose={() => setImportModal(null)}
           connectionId={activeConnectionId}
           databaseName={importModal.database || activeDatabaseName || "Database"}
+          targetDatabase={importModal.database || activeDatabaseName || undefined}
           filePath={importModal.filePath}
           onSuccess={() => {
             if (refreshTables) refreshTables();
--- a/src/components/modals/ImportDatabaseModal.tsx
+++ b/src/components/modals/ImportDatabaseModal.tsx
@@ -109,8 +109,15 @@ export const ImportDatabaseModal = ({
     }
     hasStartedRef.current = true;
     startImport();
+  }, [isOpen, startImport]);
 
-    // Listen to progress events
+  // The progress listener lives in its own effect: if it shared the starter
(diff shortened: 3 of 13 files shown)


debba:
@gcapellib Great work on this one — both diagnoses were spot on, and mirroring the `dump_database` override was exactly the right call. Thanks also for the detailed writeup, it made the review much easier.

I found one side effect of the run-once guard while testing: the `import_progress` listener was subscribed in the same effect as the guard, so whenever that effect re-ran (strict-mode double mount in dev, or `startImport` changing identity mid-import) the cleanup unsubscribed the listener and the early return never re-registered it — the progress bar froze while the import kept running. I pushed 3bc5475 to your branch, which moves the listener into its own effect keyed on `isOpen` only; the guard keeps the run-once behavior.

Verified on MySQL 8.4 (Docker): a multi-table dump imported into the right-clicked database in a single pass, with progress updating correctly in dev. If you want to give it another spin on your MariaDB setup before merge, that would be great.

Re your PG question: the frontend fallback sends the connection's own database, so the override is a no-op there — nothing to worry about.

PR #517, opened 2026-07-23
feat(editor): allow reordering tabs via drag-and-drop

maximumbreak:
## Demo

https://github.com/user-attachments/assets/daedf8a5-c68f-4c40-89ce-7aa55f892735

## Summary

- Console, table, query-builder, and notebook tabs all share one tab bar and one flat tabs array, so tabs of any type can now be dragged to reorder — no artificial restriction to console-only.
- Reordering is scoped to the active connection: dragging only reshuffles that connection's own tabs; tabs belonging to other connections keep their original slots in the underlying array.
- Follows the existing native HTML5 drag-and-drop pattern already used for notebook cell reordering (src/utils/notebookDnd.ts) — same auto-scroll behavior near the edges, same gap-based insertion model, default browser drag ghost with reduced opacity on the source tab, and an insertion-line indicator showing where the tab will land.
- No new persistence code: the existing tab-save effect (EditorProvider) already writes to preferences.json whenever tabs changes, so the new order is picked up automatically.

## Changes

- src/utils/tabDnd.ts (new): moveTab() — pure, connection-scoped array reorder.
- tests/utils/tabDnd.test.ts (new): unit tests covering forward/backward moves, first/last position, no-ops, unknown ids, out-of-range input, and the multi-connection interleaving case.
- src/contexts/EditorContext.ts / EditorProvider.tsx: added reorderTab(fromTabId, insertAt).
- src/pages/Editor.tsx: wired up draggable/drag handlers on tab elements, container-level drop + auto-scroll, and the insertion-line indicator.

## Test plan

- [x] pnpm typecheck — clean
- [x] pnpm lint — clean
- [x] pnpm test — 170 files / 3121 tests passing
- [x] Manually verified in pnpm tauri dev against a live Postgres connection
- [ ] Reviewer: drag tabs across a few connections, confirm order persists after switching away and back


debba:
Looks promising, it's something we'd like to integrate.
I will test it ASAP

debba (changes requested):
Tested this locally on Linux (Arch, WebKitGTK through Tauri). The reorder logic itself is solid: moveTab is pure and well tested (10/10 passing here), the gap index model matches the per-connection filtered tabs the context exposes, and the order persists through the existing save effect. Typecheck and lint are clean on my machine too.

Requesting changes for two things I hit while testing:

1. The drag never starts on Linux because dragstart leaves the data store empty. WebKitGTK (and Firefox) won't initiate a drag without data, so on those platforms the feature silently does nothing. The notebook cell DnD you referenced already handles this in NotebookView. Adding this line in handleTabDragStart fixed it for me:

```ts
e.dataTransfer.setData("text/plain", tabId);
```

2. Reordering now permutes the tabs array, and the editor panes near the bottom of Editor.tsx are rendered in tabs order (all mounted, hidden with display:none, precisely to avoid Monaco remounting). With this PR every drop makes React move live Monaco editor DOM nodes around with insertBefore, for no visual benefit since those panes are invisible anyway. Please render the panes in a stable order (insertion order is fine) so reordering only affects the tab strip. Monaco is fragile around DOM churn, especially on WebKitGTK, and we already had to wrap it in an error boundary.

Unrelated heads up: while testing I hit an editor crash (_throwIfDisposed from setModel), but it traces back to a known upstream bug in @monaco-editor/react that only fires in dev builds with StrictMode (suren-atoyan/monaco-react#794). Not caused by this PR.

For context, this repo moved its chrome-level drags (sidebar, split panes) to src/utils/pointerDrag.ts because native HTML5 DnD can freeze the WebKitGTK compositor. With the setData fix the native path works for tabs on my setup, so I'm fine keeping it, but if black screen reports show up after merge the fallback plan is porting this to startPointerDrag.

src/pages/Editor.tsx
```diff
@@ -549,6 +551,115 @@ export const Editor = () => {
...
+      autoScrollRafRef.current = requestAnimationFrame(tick);
+    };
+    tick();
+  }, []);
+
+  useEffect(() => stopTabAutoScroll, [stopTabAutoScroll]);
+
+  const handleTabDragStart = useCallback(
+    (e: React.DragEvent<HTMLDivElement>, tabId: string) => {
+      e.dataTransfer.effectAllowed = "move";
```
debba, line 620: This is why the feature does nothing on Linux and Firefox: WebKitGTK won't initiate a drag while the data store is empty, so dragstart aborts before any dragover ever fires. Verified locally on Arch that adding setData makes the drag work. The notebook cell DnD this PR references already does the same in NotebookView.

```suggestion
      e.dataTransfer.effectAllowed = "move";
      // WebKitGTK (and Firefox) won't start a drag with an empty data store.
      e.dataTransfer.setData("text/plain", tabId);
```

src/pages/Editor.tsx
```diff
@@ -549,6 +551,115 @@ export const Editor = () => {
...
+    (e: React.DragEvent<HTMLDivElement>) => {
+      e.preventDefault();
+      stopTabAutoScroll();
+      const fromTabId = dragTabId;
+      const insertAt = dropPos;
+      setDragTabId(null);
+      setDropPos(null);
+      setDropIndicatorLeft(null);
+      if (fromTabId === null || insertAt === null) return;
+      reorderTab(fromTabId, insertAt);
```
debba, line 686: Permuting the tabs array has a side effect beyond the strip: the editor panes further down in this file are rendered in tabs order, all mounted and hidden with display:none, precisely so Monaco never remounts. After this call React reorders those keyed pane divs with insertBefore, which detaches and reattaches live Monaco DOM nodes on every drop. The panes are invisible, so their DOM order buys us nothing, and Monaco is fragile around that kind of churn on WebKitGTK (we already wrap it in an error boundary for a reason).

Please decouple the two: keep the strip rendering in tabs order, but render the panes in a stable order, for example insertion order or sorted by tab id. That way a reorder only ever touches the tab strip DOM.

debba:
@maximumbreak thanks a lot for this contribution! The drag and drop feels really smooth and it was a good call to follow the existing notebook cell DnD pattern.

Two fixes landed during the review:

1. `dataTransfer.setData` in the dragstart handler. Without it WebKitGTK and Firefox never start the drag, so the feature did nothing on Linux. You already applied the suggestion, thanks for the quick turnaround.
2. I pushed a small commit that renders the editor panes in a stable order (sorted by tab id), decoupled from the tab strip order. Since all panes stay mounted with display:none, reordering the tabs array was making React move live Monaco DOM nodes on every drop, which Monaco does not love on WebKitGTK.

Tested locally on Linux: typecheck, lint and the full test suite pass, and reordering works as expected in the app.

If you feel like keeping contributing we would be happy to have you around. Come say hi on our Discord: https://discord.com/invite/K2hmhfHRSt

Merging, thanks again!

debba (approved):
Both review points addressed, tested locally on Linux. LGTM.

PR #519, opened 2026-07-23
fix(editor): skip string literals and comments in query param detection

### LINKED ISSUE #458: Query params modal triggers on ":" inside string literals
**Problem**

The editor detects named parameters with a plain regex over the whole SQL text (`extractQueryParams` in `src/utils/queryParameters.ts`). It has no idea about string literals or comments, so a query like:

```sql
SELECT * FROM categories WHERE value = 'x:y'
```

pops up the "Query Parameters" modal asking for a value for `y`, even though `:y` is just part of the data. Worse: if you actually fill in a value, `interpolateQueryParams` happily replaces the text *inside* the quoted string, so the query gets silently corrupted.

Anything with colons in data hits this — URLs, timestamps, `key:value` pairs, JSON snippets stored as text.

**How other tools handle it**

DataGrip and DBeaver both tokenize the SQL before looking for parameters, so a `:name` inside a string literal or a comment is never treated as a bind parameter (DBeaver even lets you turn `:name` detection off entirely). TablePlus mostly sidesteps the issue by not doing aggressive parameter detection at all.

**Proposed fix**

No need for a full SQL parser — a small scanner that walks the query and skips over:

- single-quoted strings (including `''` escapes, and `\'` for MySQL)
- double-quoted and backtick-quoted identifiers
- `-- line` and `/* block */` comments
- Postgres dollar-quoted strings (`$tag$ ... $tag$`)

Then run the existing param regex only on the parts outside those ranges. `interpolateQueryParams` needs the same treatment, replacing matches by position on the original SQL so values never land inside a literal. The `::cast` exclusion already in place stays as is.

Unit tests should cover: `'x:y'`, `::text` casts, params inside comments, dollar quotes, escaped quotes, and a real `:param` sitting next to a string literal.

ymadd:
Fixes #458

## Problem

`extractQueryParams` / `interpolateQueryParams` ran a plain regex over the whole SQL text, so `:y` in

```sql
SELECT * FROM categories WHERE value = 'x:y'
```

popped the Query Parameters modal — and filling in a value silently rewrote the *inside of the string literal*, corrupting the query. Anything with colons in data hit this: URLs, timestamps, `key:value` pairs, JSON stored as text.

## Approach

Instead of adding a second scanner, this reuses the dialect-aware tokenizer that already powers the statement splitter (#225):

- **`scanNonCodeSpans(sql, dialect?)`** (new export from `src/utils/sqlSplitter`) returns the spans of every string literal and comment, driving `scanToken` with the same state updates as the splitter. `tokenizer.ts` internals stay private; the collector is an iterative worklist (see Hardening below).
- **`queryParameters.ts`** runs detection and interpolation on a **length-preserving mask** (non-code spans blanked with spaces) with the existing regex unchanged, so the `::cast` exclusion behaves exactly as before. Interpolation is position-based over the *original* text (mask indices transfer 1:1), so values can never land inside a literal — including after astral characters like `'😀:id'`, which the old full-text replace corrupted.
- **`Editor.tsx`** threads the driver's `sql_dialect` capability (introduced in #225) into all six call sites, and the Params-button probe is memoized since it runs in the render path on every keystroke.

**Fallback semantics.** With an `undefined`/unknown dialect nothing is masked: behavior is byte-for-byte the old one (the optional `dialect` argument keeps both functions backward compatible; the pre-existing tests pass unmodified). Guessing a dialect is measurably worse than not masking — e.g. generic quote rules read MySQL's `'a\'b'` as an unterminated string and would swallow every *real* parameter after it. Masking with the correct dialect only ever removes false positives; masking with a wrong one can hide real parameters.

MySQL executable comments (`/*!50700 ... */`) are **rescanned rather than masked**, so `:params` in the executable payload stay live while literals inside it are still protected. MariaDB's `/*M! ... */` gets the same treatment at the masking layer only — the splitter still folds it as a comment, so MySQL proper never receives a comment-only statement.

## Hardening

- The executable-comment rescan is an **iterative worklist with a depth cap** instead of native recursion: a pasted ~15KB chain of `/*!` markers used to overflow the call stack (reachable from the render path on MySQL connections). Past the cap the tail is left unmasked — the same fail-safe direction as the unknown-dialect fallback.
- Dialect lookup uses `Object.hasOwn`, so prototype member names (`'__proto__'`, `'toString'`) fall back cleanly instead of handing a non-`DialectOptions` value to the tokenizer. `normalizeDialect` had the same latent issue and got the same fix.
- Interpolation ignores parameter names that only resolve through `Object.prototype` (`:toString` no longer splices function source into the SQL).

## Dialect-preset corrections

Building the masking layer surfaced preset gaps that also affect statement splitting; each is fixed with splitter canary tests:

| Dialect | Fix | Reference |
|---|---|---|
| MySQL/MariaDB | `#` line comments recognized (any position, no whitespace requirement). Previously a `;` inside `# note;` mis-split the script. | [MySQL comment syntax](https://dev.mysql.com/doc/refman/8.0/en/comments.html) |
| MSSQL | `"double-quoted"` identifiers added to the quote rules (identifiers under default `QUOTED_IDENTIFIER ON`, strings when `OFF` — non-code either way). | [SET QUOTED_IDENTIFIER](https://learn.microsoft.com/en-us/sql/t-sql/statements/set-quoted-identifier-transact-sql) |
| MSSQL | Block comments now nest, per T-SQL semantics. | [Slash-star comments](https://learn.microsoft.com/en-us/sql/t-sql/language-elements/slash-star-comment-transact-sql) |
| PostgreSQL | A dollar quote glued to an identifier (`foo$tag$`) is no longer treated as a string opener. | [Lexical structure](https://www.postgresql.org/docs/current/sql-syntax-lexical.html) |
| MariaDB | `/*M! ... */` payloads rescanned at the masking layer (see above). | [Comment syntax](https://mariadb.com/docs/server/reference/sql-statements/comment-syntax) |

## Known limitations (follow-up)

- **A dedicated MariaDB dialect**: tokenizer-level `/*M!` statement semantics (executing them as statements is only correct for MariaDB, and both servers currently share the `mysql` dialect).
- **`NO_BACKSLASH_ESCAPES`**: the sql_mode is per-session server state a static lexer cannot know; under that mode `'a\'` is a closed literal. Handling it would require reading `@@sql_mode` through the driver.

I'll open a follow-up issue for these after this lands.

## Tests

59 new cases: 28 in `tests/utils/sqlSplitter/nonCodeSpans.test.ts` (span exactness and sorted/non-overlapping/in-bounds invariants, UTF-16 indices with astral chars, unterminated strings/comments, dialect fallback incl. prototype-key names, per-dialect matrix, `DELIMITER` collision cases, `GO`/slash advancement, the depth-cap boundary, and a regression test for the 15KB `/*!` chain), 27 in `tests/utils/queryParameters.test.ts` (issue repro, boundary cases like `'x'::text` / `/*c*/:id` / `:na/*c*/me`, maskless-fallback equivalence, position-based interpolation incl. unknown-param preservation), and 4 splitter canaries in `tests/utils/sqlSplitter/dialects.test.ts`.

Full suite: **3170 passed**, `tsc --noEmit` clean, `eslint` clean. The pre-existing `queryParameters` tests pass unmodified.

### DIFF
--- a/src/pages/Editor.tsx
+++ b/src/pages/Editor.tsx
@@ -760,7 +760,7 @@ export const Editor = () => {
       if (!(await guardDangerousQuery(textToRun))) return;
 
       // Check for parameters
-      const params = extractQueryParams(textToRun);
+      const params = extractQueryParams(textToRun, activeDialect);
       if (params.length > 0) {
         const storedParams = paramsOverride || targetTab.queryParams || {};
         const missingParams = params.filter(
@@ -785,7 +785,7 @@ export const Editor = () => {
         }
 
         // Interpolate parameters before execution
-        textToRun = interpolateQueryParams(textToRun, storedParams);
+        textToRun = interpolateQueryParams(textToRun, storedParams, activeDialect);
       }
 
       // Automatically open the results panel when running a query — but only
@@ -985,6 +985,7 @@ export const Editor = () => {
       activeDatabaseName,
       addHistoryEntry,
       guardDangerousQuery,
+      activeDialect,
     ],
   );
 
@@ -1000,7 +1001,7 @@ export const Editor = () => {
 
       // Collect all unique parameters across all queries
       const allParams = [
-        ...new Set(queries.flatMap((q) => extractQueryParams(q))),
+        ...new Set(queries.flatMap((q) => extractQueryParams(q, activeDialect))),
       ];
       if (allParams.length > 0) {
         const storedParams =
@@ -1022,7 +1023,9 @@ export const Editor = () => {
           return;
         }
         // Interpolate all queries with the stored params
-        queries = queries.map((q) => interpolateQueryParams(q, storedParams));
+        queries = queries.map((q) =>
+          interpolateQueryParams(q, storedParams, activeDialect),
+        );
       }
 
       const pageSize =
@@ -1168,7 +1171,7 @@ export const Editor = () => {
       });
       updateTab(targetTabId, { isLoading: false });
     },
-    [activeConnectionId, updateTab, patchResultEntry, settings.resultPageSize, activeSchema, t, isMultiDb, activeDatabaseName, addHistoryEntry, guardDangerousQuery],
+    [activeConnectionId, updateTab, patchResultEntry, settings.resultPageSize, activeSchema, t, isMultiDb, activeDatabaseName, addHistoryEntry, guardDangerousQuery, activeDialect],
   );
 
   // Auto-run entry point for navigation-initiated executions (sidebar "open
@@ -2574,7 +2577,7 @@ export const Editor = () => {
   const handleEditParams = useCallback(() => {
     if (!activeTab || !activeTab.query) return;
 
-    const params = extractQueryParams(activeTab.query);
+    const params = extractQueryParams(activeTab.query, activeDialect);
     if (params.length === 0) return;
 
     setQueryParamsModal({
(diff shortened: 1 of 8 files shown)


debba:
Looks perfect to me.
Merging it.

PR #520, opened 2026-07-23
fix(editor): use the Tauri clipboard API for Cut and Copy

gcapellib:
## Problem

In the SQL editor's right-click menu, Cut and Copy silently do nothing — nothing reaches the clipboard, and Cut doesn't remove the text either. Ctrl+C and Ctrl+X work fine, so it's specific to the context-menu actions.

Reproduced on Kubuntu 26.04, Wayland, WebKitGTK 2.52.3.

## Cause

Monaco's built-in Cut/Copy go through `document.execCommand`, which fails inside the WebView on this setup. Confirmed directly in the devtools console with a selection active:

```js
document.execCommand('copy')  // → false
```

This isn't a Tauri configuration issue — wry already calls `set_javascript_can_access_clipboard(true)` on the WebKitGTK settings. It looks like a platform-level limitation, so working around it in the app is the practical option.

The codebase already does exactly that for Paste: `SqlEditorWrapper.tsx` registers a `tauri.clipboardPaste` action using the plugin's `readText`, and filters the built-in out of the context menu, with the comment *"doesn't work in Tauri"*. Cut and Copy just never got the same treatment.

## Fix

Same approach, applied symmetrically:

- register `tauri.clipboardCut` / `tauri.clipboardCopy` using `writeText`
- extend the context-menu filter to hide all three built-ins instead of only the paste one
- with no selection, both fall back to the current line, matching Monaco's original behaviour

## Notes

- The menu filter still relies on Monaco's internal `_getMenuActions`. That was already the case for Paste — this PR doesn't introduce the hack, it extends it — but worth knowing it could break on a Monaco upgrade.
- Only touches the SQL editor. The result grid has its own separate issues (no context menu, no click-and-drag selection) which I've left out of this PR.

Tested on Kubuntu 26.04 / Wayland: Copy, Cut, Paste all work from the context menu, keyboard shortcuts still work, no duplicate entries in the menu. `npx tsc --noEmit` passes.


debba:
It is working very good from my side, merging it

PR #532, opened 2026-07-25
feat(editor): label the Run button with what it will actually run

ymadd:
## Problem

With no selection, `Cmd/Ctrl+Enter` and the Run button execute only the statement under the cursor (#464). The button reads "Run" in every case, so pasting a multi-statement script and pressing Run silently executes one statement and skips the rest — no error, no warning, nothing in the UI saying that only part of the buffer ran.

I hit this against a production database: a 21-statement script where only the last statement (the one the caret landed in after pasting) executed. It surfaced as a foreign key violation, because the 9 parent rows that the remaining statements depended on were never inserted. Without that FK constraint it would have failed silently.

## What this changes

The Run button now says what it will actually run:

| State | Label |
| --- | --- |
| Selection active | **Run Selection** |
| Multi-statement script, no selection | **Run Statement** |
| Single statement / table tab | **Run** (unchanged) |

Behaviour is untouched — this only makes the existing behaviour visible before you commit to it.

Additionally, when the button would run one statement out of several, the tooltip surfaces `Run All (Cmd/Ctrl+Shift+Enter)`. That shortcut is already bound in `SqlEditorWrapper`, but nothing in the UI mentions it — and it is exactly the escape hatch you want at the moment the label warns you.

## Implementation

- `resolveRunTarget()` (`src/utils/runTarget.ts`) resolves the target from selection state, statement count and the `runStatementUnderCursor` setting — a small pure function so the label can't drift from the behaviour it describes.
- `SqlEditorWrapper` gains `onRunContextChange`, reusing the existing cursor-statement highlight pass (`onDidChangeCursorSelection` / `onDidChangeModelContent`). It only fires when the answer changes, so moving the caret doesn't re-render the toolbar on every keystroke.
- `runSelection` / `runStatement` added to all 10 locales.

## Testing

- `pnpm typecheck` clean
- `eslint` clean on the changed files
- 6 unit tests for `resolveRunTarget`
- Checked in the app: the label switches as expected when selecting text, moving the caret between statements, and editing the buffer


debba:
Thanks a lot for this, I really like it. You are right that the button was hiding what it would actually do, and deriving the label from the same conditions the run handler branches on is the right way to keep the two from drifting. The pure helper with its own tests, the Run All hint in the tooltip, all ten locales updated: nothing to ask for on that side.

One thing I would like to see before merging. The label reacts to the selection and to the statement count, but not to where the caret actually sits. In a multi-statement buffer it reads "Run Statement" in every case, including when the caret is still above the first statement (a leading comment, or a couple of empty lines at the top). There the highlight is cleared, `findStatementAtOffset` returns `undefined`, and `handleRunButton` falls back to `queries[0]`, so the button promises the statement under the cursor when there isn't one.

Could `onRunContextChange` report the statement at the caret too? `updateCursorStatementHighlight` already resolves it a few lines earlier, so it would be the same pass with no extra work, and the label would follow the caret and always agree with the highlight.

Running the statement at the cursor is fairly recent here (#464, with #487 making it the default for Run and Explain), so having the highlight and the label tell the same story would round that off nicely.

ymadd:
Thanks for catching this — you're right that the leading gap breaks the story. I took a slightly different route than teaching the label about the caret, though: if the label just followed the cleared highlight it would fall back to a generic "Run", while the button still executes `queries[0]` — the mismatch would survive, just pointing the other way.

Instead I made the resolution total: `findStatementAtOffset` now resolves an offset above the first statement to the first statement, mirroring how the gaps between and after statements already snap to the preceding one (the leading gap was the only offset in a non-empty buffer that resolved to nothing). With that, the highlight follows the caret into a leading comment and shows exactly the statement Run will execute, so the label, the highlight and the behaviour agree at every caret position — no change to what actually runs. It also fixes a small pre-existing inconsistency where Explain was a silent no-op in that same spot; it now targets the same statement as Run.

While in there I also pushed a cleanup commit: `handleRunButton` now dispatches on `resolveRunTarget` instead of mirroring its branching inline (so label and behaviour genuinely share one decision), the run context re-announces itself when a tab becomes active (it could describe the previous tab after a switch), and the statement-split cache is keyed on the model version id instead of the text.

Happy to switch to the `onRunContextChange` approach instead if you'd rather keep the highlight cleared there.

debba:
Thanks for the thorough follow-up — the reasoning is convincing. You're right that teaching the label about the caret would only have flipped the mismatch; making the resolution total is the cleaner fix, and it's consistent with how the gaps between and after statements already snap to the preceding one. Nice bonus that Explain now targets the same statement instead of silently doing nothing there.

The cleanup commit looks good too: dispatching `handleRunButton` on `resolveRunTarget` delivers what the helper promised, and the tab-switch re-announce and the version-id cache key both check out on my read.

No need to switch to the `onRunContextChange` approach — this route covers it better. Nothing else from my side: if you're done, I'll proceed with the merge. Thanks again for the great work on this!

ymadd:
All done from my side, nothing else pending. Thanks for the review, feel free to merge!

PR #537, opened 2026-07-27
feat(i18n): add Brazilian Portuguese (pt-BR) translation

jeffersongoncalves:
## Summary
- Adds full pt-BR UI translation (`src/i18n/locales/pt-BR.json`), registered in `src/i18n/config.ts`
- Adds `README.pt-BR.md` and links it from every other README's language switcher
- Fixes `nonExplicitSupportedLngs: true`, which stripped hyphenated language codes (`pt-BR` -> `pt`) before matching against `supportedLngs`, silently falling back to English for any region-coded locale — confirmed with an isolated i18next repro that this was the actual cause of pt-BR not applying, and that the fix doesn't regress existing detection (`en-US`->`en`, `zh-CN`->`zh`, `fil-PH`->`fil`)

## Test plan
- [x] `pt-BR.json` validated as syntactically valid JSON with identical key structure and interpolation placeholders vs `en.json` (0 mismatches)
- [x] Reproduced the language-switch bug in isolation with i18next directly, confirmed root cause, confirmed fix
- [ ] Manual: select "Português (Brasil)" in Settings and confirm the UI switches


debba (approved):
Thanks a lot for this contribution! I reviewed the code side and the fix on `nonExplicitSupportedLngs` is correct: it was stripping region codes before matching against `supportedLngs`, so pt-BR could never be applied. I also verified it doesn't regress detection for the existing locales (en-US, zh-CN, fil-PH all still resolve correctly).

As for the translation itself, I don't speak Portuguese, so I have no way to verify its correctness... I'll just trust you on that one :)

Really appreciate the clean work on the README language switchers too. Please keep contributing, you're more than welcome here!

PR #541, opened 2026-07-27
feat(ai): add Atlas Cloud provider

binyangzhu000-sudo:
## Summary

- add Atlas Cloud as a first-class AI provider in settings and automatic key detection
- reuse the existing OpenAI-compatible model discovery and chat-completions transport with `https://api.atlascloud.ai/v1`
- support `ATLASCLOUD_API_KEY` and seed currently available text models
- add provider label, detection, default-model, and endpoint-path coverage

## Validation

- `pnpm typecheck`
- `pnpm build`
- `pnpm exec vitest run tests/utils/settings.test.ts tests/utils/settingsUI.test.ts` (83 passed)
- ESLint on the modified TypeScript source files (0 errors)
- `rustfmt --edition 2021 --check src/ai.rs`
- `git diff --check`
- YAML parse and default-model assertion
- Atlas Cloud public `/v1/models` probe (121 models; all three seeded models present)

The focused Rust tests could not finish compiling Tauri dependencies in this environment because the local disk ran out of space. No test assertion failed.

## Scope

No README, documentation, logo, sponsor, credits, or partner-promotion changes.

### DIFF
--- a/src-tauri/src/ai.rs
+++ b/src-tauri/src/ai.rs
@@ -7,6 +7,8 @@ use std::fs;
 use std::time::{SystemTime, UNIX_EPOCH};
 use tauri::{AppHandle, Manager};
 
+const ATLAS_CLOUD_BASE_URL: &str = "https://api.atlascloud.ai/v1";
+
 // --- Data Structures ---
 
 #[derive(Serialize, Deserialize, Debug)]
@@ -271,7 +273,7 @@ fn build_api_url(base_url: &str, endpoint: &str) -> String {
     format!("{trimmed}{endpoint}")
 }
 
-async fn fetch_custom_openai_models(base_url: &str, api_key: &str) -> Vec<String> {
+async fn fetch_openai_compatible_models(base_url: &str, api_key: &str) -> Vec<String> {
     if base_url.is_empty() || api_key.is_empty() {
         return Vec::new();
     }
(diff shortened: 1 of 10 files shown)


debba:
@binyangzhu000-sudo thanks for the work here. The implementation is clean, and I verified the claims in the description locally: `tsc --noEmit` is clean, the two vitest files pass (83 tests), and the Rust tests that could not compile on your side pass here, including the new `test_atlas_cloud_api_urls`.

That said, I am not going to merge it as it stands, and I want to be transparent about why. I am leaving the PR open for now so you have room to reply.

Tabularis already ships an "OpenAI Compatible" provider with a user-configurable base URL. Atlas Cloud is documented as a drop-in replacement for the OpenAI SDK, so it already works today: set the base URL to `https://api.atlascloud.ai/v1`, paste the key, and that is it. No code change is required for anyone to use the service. A dedicated entry would add a brand to the settings list rather than a capability, and that is a line we hold for every OpenAI-compatible service. Otherwise the provider list grows without any real benefit for users.

For context on how this reads from our side: the same change has been opened across a large number of unrelated repositories over the past few days, and the same account has PRs merged inside the AtlasCloudAI organization. That makes this a vendor-led addition rather than something requested by our users, and it is not disclosed in the description. It does not change the technical assessment above, but it does mean the bar for a first-class provider entry is not met.

If Atlas Cloud would like to support the project as a sponsor, that is a different conversation and one we are open to having. In that case a dedicated provider entry, with a proper Atlas Cloud icon instead of the OpenAI one, is something we would consider. A GitHub Discussion or a direct message to the maintainers is the right place to start, and this PR can stay open in the meantime.

Two technical notes for the record, in case this does move forward. The provider reuses `OpenAIIcon`, which would show the OpenAI logo next to "Atlas Cloud" in settings. And the three hardcoded model ids in `ai_models.yaml` are asserted in a Rust test, which ties our test suite to a third-party model catalog that we do not control.

PR #542, opened 2026-07-27
Fix DML submit schema selection

DhruvShah-Dev:
## Summary
- Use the table tab schema when submitting row inserts, updates, and deletes for schema-capable drivers.
- Keep existing multi-database submit behavior by sending the tab value as `database` for those drivers.
- Add unit coverage for data-change scope resolution.

## Validation
- pnpm.cmd vitest run tests/utils/database.test.ts
- pnpm.cmd typecheck
- pnpm.cmd lint
- pnpm.cmd build
- git diff --check


debba:
Thanks for the fix! Nice catch on the DML submit using the sidebar's active schema instead of the tab's own schema. The extraction into getTableDataChangeScope also cleans up the triple duplication in the submit handler, and the tests cover all the branches. Merging.

debba (approved):
Reviewed the diff: the fix correctly aligns DML submits with the tab schema, matching the read path. Tests cover all branches and CI is green.

PR #543, opened 2026-07-27
Fix PostgreSQL key metadata for readonly users

DhruvShah-Dev:
## Summary
- Replace PostgreSQL primary-key column detection with a `pg_catalog` query instead of `information_schema.table_constraints`.
- Reuse the catalog expression across single-table, batch-table, and view column loading.
- Extend coverage so the Postgres FK integration scenario also verifies primary-key detection on the referenced table.

## Why
Issue #96 reports that readonly users can have access to `pg_constraint` while `information_schema.table_constraints` returns no rows. That caused key metadata to be missed even when the required catalog data was visible.

## Validation
- `pnpm.cmd typecheck`
- `pnpm.cmd lint`
- `pnpm.cmd build`
- `git diff --check`
- `pnpm.cmd exec gitnexus detect-changes` reported low risk


debba (approved):
Thanks a lot for this contribution! The fix looks solid: the pg_catalog-based EXISTS expression is correlated on schema, table, and column, handles composite keys via unnest(conkey), and correctly skips dropped columns. I reproduced the original issue locally with a SELECT-only role — information_schema.table_constraints returns no rows for it while the new pg_constraint query correctly reports the primary key — so this genuinely fixes #96. Unit tests (102 passed) and the Postgres integration test both pass on the branch. Merging.

PR #549, opened 2026-07-28
feat(datagrid): Select All rows via Cmd/Ctrl+A and context menu

### LINKED ISSUE #546: [Feat]:Multi-select column
### Prerequisites

### 🚀 Feature Proposal

I'd like to request a feature in Tabularis. Can you add a Select All feature for tables?

This means that when I click on a table, all the data in that table is immediately selected (row by row), making it easier to copy and paste into other applications without having to select each item individually.

Thank you in advance.

### Motivation

Makes it easier to copy and paste across multiple columns

### Example

_No response_

iamthenuggetman:
## Summary

Closes #546

Feature request from Discord: add a **Select All** feature so users can grab all rows of a table at once and copy/paste them into other applications without selecting row by row.

The grid already implemented select-all (`handleSelectAll`) behind a click on the `#` header cell, but the action was undiscoverable — and it silently wrote to the clipboard. This PR surfaces it and splits selection from copying cleanly:

- **Cmd/Ctrl+A** selects all loaded rows. Guarded so it only fires in the last-interacted grid (multi-result panels mount several grids) and never inside text inputs or while editing a cell.
- **Select All / Deselect All** entry in the row context menu (label toggles with selection state); the `#` header tooltip toggles the same way.
- **Copy Selected (N)** and **Copy All (M)** as adjacent, explicit entries in the row context menu — the two copy scopes read as a pair. `Copy All` shows the total when known, a count-less label when not.
- **Copy feedback**: every row-copy path shows a success toast via the app's `ToastProvider` — "Copied N row(s) to clipboard", or "Copied N of M row(s)" when a page-only copy of a larger result happens, so a partial copy is never silent.
- **DBeaver-style multi-column selection** (follow-up request): Cmd/Ctrl+click a column header toggles it, Shift+click range-selects — plain click keeps sorting. Selected columns highlight in header and body; row/column selection are mutually exclusive so copy semantics stay unambiguous. Cmd/Ctrl+C copies the selected columns (all loaded rows) in the active format, and the header context menu offers Select/Deselect Column and Copy Selected Columns (N). On macOS, a Ctrl-held contextmenu event on a header routes to the toggle (Ctrl+click is right-click there).
- **DBeaver-style cell range selection** (follow-up request): plain click focuses the anchor cell, Shift+click extends a rectangular range from the anchor — highlighted like column selection. Cmd/Ctrl+C or the context menu's **Copy Range (R×C)** entry copies just the rectangle in the active copy format and toasts "Copied N cell(s) to clipboard". Row, column, and cell-range selection are mutually exclusive.
- i18n: all new strings added to all 11 locales.

## Semantics

- **Selecting never copies.** The clipboard is only written by an explicit copy action (Cmd/Ctrl+C or the menu).
- **Cmd/Ctrl+C copies the selection instantly** — no modal, no interruption. If the selection covers the loaded page of a larger result, the "N of M" toast discloses the partial scope; upgrading to the full result is the `Copy All (M)` menu item.
- **Copy All re-runs the query unpaginated** and strips the tab's "Total Limit" clause (matching how `count_query` computes totals), so it truly covers every row; on-screen sort order is preserved. When the row total is unknown (no count requested yet), it fetches with a large practical cap and the toast reports the actual row count.
- Row formatting is shared via `formatRowsForCopy` in `utils/clipboard.ts`, so grid-copy and copy-all produce identical output in the active copy format.

## Test plan

- 15 grid tests in `tests/components/ui/DataGrid.test.tsx`:
  - Cmd/Ctrl+A selects all rows after interacting with the grid (without copying)
  - Cmd/Ctrl+A ignored when the grid wasn't interacted with / from editable targets
  - Cmd/Ctrl+C copies the selection instantly and toasts the row count
  - Full-page copy on a paginated result toasts "N of M" and never fetches
  - `#` header cell toggles select-all / deselect-all
  - Context menu shows "Select All (N)", "Copy Selected (N)", and "Copy All (M)" (count-less variant when the total is unknown)
  - Column selection: modifier-click toggle/range, plain click still sorts, exclusivity with row selection, column-only copy, Ctrl+click contextmenu routing on macOS
  - Cell range: Shift+click rectangle highlight, range-only copy with cell-count toast, exclusivity with column selection, anchor move on plain click, Copy Range (R×C) menu entry
- `pnpm vitest run`: full suite passes
- `tsc -b`: clean (note: the repo's `typecheck` script is a no-op — solution-style tsconfig with `files: []`; `tsc -b` is the real gate)
- `eslint` on changed source: clean
- Manually tested against the `demo/` Docker stack (`perf_demo.wide_table`, 50k rows)


mocahyoo:
Thanks for working on this, @iamthenuggetman ! The current "select all" implementation works great on my end.
Would it be possible to also support selecting multiple specific columns (similar to DBeaver's behavior)?  ( •̀ ω •́ )✧

image

iamthenuggetman:
hey @mocahyoo thanks for checking it out. just pushed a new commit with your request in it. give it a try and let me know how it works for you (works on my machine lol)

mocahyoo:
Wooah the update is working great for me

Just a quick idea, would it be okay to add cell range selection too?

iamthenuggetman:
done

debba:
Hi @iamthenuggetman can you resolve conflicts .
After that I will merge it.
Great work :)

iamthenuggetman:
@debba should be good now

debba:
Thank you for your contribution! We really appreciate the time and effort you put into this. If you'd like to contribute more in the future, we'd be happy to have you. Thanks again!

debba (approved):

PR #552, opened 2026-07-28
feat(data-grid): navigate cells with the arrow keys

### LINKED ISSUE #551: [Feat]: DataGrid: move between cells with the keyboard
### Prerequisites

### 🚀 Feature Proposal

The result grid highlights a focused cell, but the focus can only ever be placed by clicking. There is no way to move it with the keyboard — the arrow keys, `Home`/`End` and `PageUp`/`PageDown` do nothing — so reading or editing a table row by row means reaching for the mouse on every cell.

Proposal: let the focused cell be moved with the arrow keys, jump to the first/last column with `Home`/`End`, page through rows with `PageUp`/`PageDown`, and open the focused cell for editing with `Enter` or `F2` (the same action as a double-click).

### Motivation

Every other part of the app is keyboard-driven — the editor runs a query with `Ctrl+Enter`, rows are deleted with `Delete` (#218), a focused cell is copied with `Cmd/Ctrl+C` — but the grid itself is mouse-only once you want to move around it. Reviewing a wide result set, or correcting several values in a row, currently costs a click per cell.

It also leaves a gap flagged in #86: the reporter noted that result cells announce themselves as clickable while `Enter` does nothing on them.

### Example

```
Click a cell, then:
  ↑ ↓ ← →        move one cell
  Home / End     first / last column of the row
  PageUp/Down    one viewport of rows
  Enter or F2    start editing the focused cell
  Escape         cancel, focus stays on the cell so navigation continues
```

ymadd:
## Summary

Closes #551.

The grid already tracked a focused cell (`focusedCell`, added in #161) and used it for `Cmd/Ctrl+C`, but it could only ever be set by clicking — nothing moved it from the keyboard. This adds cell navigation:

| Key | Action |
|---|---|
| `↑` `↓` `←` `→` | move one cell, clamped at the edges |
| `Home` / `End` | first / last column of the row |
| `PageUp` / `PageDown` | one viewport of rows |
| `Enter` / `F2` | open the focused cell for editing (same path as double-click) |

The first key press in a grid that has no focused cell yet enters at the top-left cell.

## Design notes

- **The handler is bound to the scroll container, not to `document`.** The existing `Cmd+C` / `Delete` handler is document-level, which is fine for actions gated on a selection but would make every mounted grid move at once — a notebook renders one grid per SQL cell. Clicking a cell focuses the container so key events reach it, and `tabIndex={0}` means `Tab` reaches the grid as well.
- **Keys are left alone for anything that handles them itself**: text inputs, the FK/BLOB buttons inside cells, and the sortable column headers (`role="button"`). Without this, `Enter` on a focused FK button would be swallowed instead of activating it.
- **Focus returns to the grid when an edit is closed with `Enter`/`Escape`**, so navigation continues; committing by clicking elsewhere deliberately leaves focus where the user clicked.
- **`Enter` resolves the cell value through the same display helpers as the mouse path**, so a cell pending the database DEFAULT opens an empty editor rather than exposing the internal `__USE_DEFAULT__` sentinel.
- The row height was duplicated between the virtualizer estimate and the row markup; it is now `DATA_GRID_ROW_HEIGHT` in `src/utils/dataGrid.ts`.

## Not included

`role="grid"` / `aria-activedescendant` and the rest of the ARIA surface for the focused cell. The container currently suppresses its own focus ring in favour of the cell ring, which is a visual-only affordance. That is worth doing properly as a follow-up on top of #355 rather than half-done here.

## Test plan

- 9 new tests in `tests/components/ui/DataGrid.test.tsx` covering arrow movement, edge clamping, `Home`/`End`, modifier keys being ignored, container focus on click, keys passing through to focusable controls, `Enter` → edit → `Escape` → focus restored, and the DEFAULT-sentinel case. Each was confirmed to fail against the unfixed code.
- `pnpm vitest run`: 187 files / 3368 tests pass
- `pnpm typecheck` and `eslint` on the changed sources: clean
- Manually exercised in `tauri dev` against a MySQL table.

## Note for reviewers

This touches the same anchor in `DataGrid.tsx` as #549 (Select All via `Cmd/Ctrl+A`) and both add tests to `DataGrid.test.tsx`, so whichever lands second will need a small rebase. They also overlap conceptually: #549 tracks the "last-interacted grid" with a `mousedown` listener to scope its document-level shortcut, which this PR's container focus could replace.


debba:
Really like this one, the feature feels solid and the write-up plus test coverage make it easy to review. Thanks!

One suggestion: would it be worth also handling Tab and Shift+Tab for moving across columns in navigation mode? Tab already jumps to the next cell while editing, so having it work the same way outside of edit mode would feel consistent. The trade-off is that capturing Tab makes it harder to move focus out of the grid with the keyboard alone, so it may tie into the ARIA follow-up you mentioned. Curious what you think.

ymadd:
Thanks for taking a look! I dug into this, and I'd hold off on capturing Tab/Shift+Tab in navigation mode for now. You did prompt me to revisit my "Tab exits the grid" framing, though — the current behavior is messier than that.

**Sortable table views already have a long Tab path.** Every sortable column header is a tab stop (`tabIndex={0}`, added in #355) inside the scroll container, so a 30-column table adds roughly 30 header stops before focus leaves the grid, plus any focusable cell controls. (Query-result and notebook grids don't add those sortable-header stops, since they don't pass `onSort`.) Capturing Tab on the container would also remove the normal forward-Tab path to those headers.

**Capturing Tab before the ARIA layer would risk an accessibility regression.** While navigating cells, DOM focus stays on the container and the active cell is represented only by a CSS ring, so cell changes aren't exposed or announced to assistive technology. Arrow navigation is visual-only today, but it doesn't replace the browser's normal Tab behavior; capturing Tab would. Addressing that properly means choosing an `aria-activedescendant` or roving-tabindex model and making it work with row virtualization, where off-window cells unmount.

So I'd fold this into the ARIA follow-up. If we adopt the APG grid pattern, the grid should become a coherent single-tab-stop composite — and in navigation mode the APG convention is Tab between composites, arrow keys within, so Tab may well remain the exit key.

Your consistency point did uncover an adjacent issue, though: in *edit* mode, Tab wraps from the last cell back to row 0, so it never stops at a boundary. And Shift+Tab is handled by the same branch without checking `shiftKey`, so it also moves forward. Both behaviors predate this PR. I'd prefer to fix them in a small follow-up — stop wrapping at the outer boundaries and add a proper reverse Shift+Tab — to keep this PR focused, but I'm happy to fold them in here if you'd prefer.

debba:
Completely agree with your reasoning — folding the Tab capture into the ARIA follow-up is the right call, and handling the edit-mode wrap/Shift+Tab fixes in a separate small PR keeps this one nicely focused. Merging this in the meantime. Thanks a lot for the thorough investigation and the great write-up!

debba (approved):
LGTM — solid implementation and excellent test coverage.

PR #553, opened 2026-07-28
Fix PostgreSQL visual query identifier quoting

DhruvShah-Dev:
## Summary
- pass the active driver into Visual Query Builder SQL generation
- quote generated PostgreSQL table and column identifiers when needed
- add regression coverage for reserved identifiers such as `user` and `order`

## Root cause
The Visual Query Builder generated raw table and column references. PostgreSQL treats unquoted reserved words such as `user` as keywords, so generated SQL could fail for valid table names.

## Validation
- `pnpm.cmd vitest run tests/utils/visualQuery.test.ts`
- `pnpm.cmd typecheck`
- `pnpm.cmd lint`
- `pnpm.cmd build`
- `git diff --check`
- `pnpm.cmd exec gitnexus detect-changes`


debba:
Thanks for this, and thanks for taking the time to add the driver plumbing rather than patching a single call site. I reviewed it locally on the branch and ran the generated SQL against a real PostgreSQL 16 instance.

**Verification**

- `pnpm typecheck` clean, `eslint` clean on the changed files
- `pnpm vitest run tests/utils/visualQuery.test.ts tests/utils/identifiers.test.ts`: 84 passed
- Full suite: 3427 passed across 190 files, no regressions
- Blast radius is small: `generateVisualQuerySQL` has a single consumer (`VisualQueryBuilder.tsx:155`), the other functions are internal plus tests, and every new parameter is optional, so nothing breaks for existing callers

**What I confirmed against a live database**

I seeded a schema with reserved table names, mixed case identifiers created quoted, plain lowercase tables and a name containing a hyphen, then executed the SQL your code actually generates. Three failures that were real before this PR are now gone:

- `FROM order t1` was a syntax error, and `FROM user t1` silently resolved to the `USER` keyword and then failed with `column t1.id does not exist`
- `FROM AccountEventLog t1` failed with `relation "accounteventlog" does not exist`
- `t1.AccountId` failed with `column t1.accountid does not exist`

There is a fourth fix worth calling out because the description does not mention it. The old ORDER BY branch quoted the whole `t1.order` string as a single identifier, producing `ORDER BY "t1.order"` and failing with `column "t1.order" does not exist`. That affected every PostgreSQL ORDER BY on a column that needed quoting, so this is arguably the most valuable part of the change and it deserves a line in the summary.

One correction on the stated root cause: qualified reserved column references already parse fine in PostgreSQL. `SELECT t1.order, t1.select FROM "user" t1` returns rows without any quoting, because the grammar accepts reserved words as attribute names after a dot. So `t1."order"`, the case the summary leads with and that four of the new tests assert, is cosmetic rather than a fix. It is harmless and arguably more readable, but the tests would carry more weight if they targeted the table name, mixed case and ORDER BY paths that genuinely failed.

**Notes for follow up, none of them blocking**

1. `generateHavingClause` (`visualQuery.ts:328`) still emits `c.column` raw. It is reachable from the aggregate checkbox in the WHERE panel, which feeds the same `t{N}.{col}` values, so `HAVING COUNT(t1.AccountId) > 0` still fails with `column t1.status does not exist`. It looks like a one line change using the `formatGeneratedColumnRef` helper you already added.
2. Column aliases stay unquoted (`visualQuery.ts:150` and `:160`). A user typed alias containing a space produces `AS Total Count` and a syntax error. Reserved word aliases are fine, since PostgreSQL accepts reserved words as AS labels, so this is only the space and mixed case case. It predates the PR but it is the same bug class you are closing here.
3. `generateTableList` is now unused, since `generateFromClause` inlines its own quoted version, but it is still exported and tested and it returns unquoted names. A future caller would silently reintroduce the bug. Worth either deleting it or routing it through `formatTableRef`.
4. `shouldQuoteIdentifiers` matches the exact string `"postgres"` and ignores the manifest `capabilities.identifier_quote` that `getQuoteChar` already supports. I verified that a driver id of `"postgresql"` produces no quoting at all. This is pre existing, but it matters for the ongoing PostgreSQL plugin work: if the plugin id is not literally `postgres`, this fix quietly stops applying.
5. Two small things in `formatTableRef`: the `if (!driver) return tableName` guard is redundant because `formatSqlIdentifier` already returns the input unchanged for non PostgreSQL drivers, and the `.split('.')` schema handling is unreachable today since `node.data.label` is always a bare table name (`VisualQueryBuilder.tsx:209`). Not a problem, just dead weight. Related and out of scope: the FROM clause still never qualifies with `activeSchema`.

Merging with those notes. Points 1 and 3 are the ones I would like to see picked up in a follow up.

Issue #563, opened 2026-07-29
[Bug]: Tables with reserved-keyword names (order, user, group, etc.) disappear from the list


gustavomelo-dotgroup:
### Describe the bug

First off: awesome tool, really enjoying using Tabularis day to day.

Ran into a weird one though — some tables just don't show up in the sidebar/tree view. After digging a bit, I noticed a clear pattern: it's always tables whose name matches a SQL reserved keyword (like order, user, group, index...). If I rename the table to something non-reserved, it shows up again right away.

Not a Tabularis-only thing either — I reproduced the same behavior in Valentina Studio and another client. My guess is the internal schema introspection query (listing tables/columns) isn't quoting identifiers, so when a table name happens to match a reserved word, that query fails silently and the table just gets dropped from the list with no error shown.

Would be great to see this fixed (probably just needs proper double-quoting, or the right delimiter per driver, when building those metadata queries). That small detail would make the tool even more solid.

Thanks for all the work on this project! 🙌

### To Reproduce

1. Create (or use) a SQLite database and add a table using a reserved SQL keyword as its name, e.g.:
CREATE TABLE "order" (id INTEGER PRIMARY KEY, total REAL);
2. Open that database in Tabularis.
3. Look at the sidebar/tree view — the order table is missing, even though it exists in the database (you can confirm it's there by running SELECT * FROM "order"; in the SQL Editor, or via sqlite3 yourdb.db ".tables" in the terminal).
4. Now rename it to something non-reserved:
ALTER TABLE "order" RENAME TO orders;
5. Refresh the schema in Tabularis — orders now shows up fine in the tree view.

Same pattern reproduces with other reserved words (user, group, index, etc.), and across different backends (tested on SQLite; likely affects Postgres/MySQL too since they have their own reserved word lists).

Suggested fix: wherever the schema introspection query is built (looks like it'd be in src-tauri/, in the SQLite/driver code that lists tables and runs PRAGMA table_info(...) or queries sqlite_master), make sure the table name is wrapped in double quotes (or the driver-appropriate quoting) before being interpolated into the query. That should stop the silent failure and let reserved-word-named tables show up like any other table.

### OS Version

Ubuntu 25.04

### Tabularis Version

v0.17.0

### Relevant Log Output

```shell

```

PR #568, opened 2026-07-29
fix(mysql): include MariaDB SYSTEM VERSIONED tables in table list

gustavomelo-dotgroup:
## Summary
- `get_tables()` for the MySQL/MariaDB driver filtered `information_schema.tables` on `table_type = 'BASE TABLE'` only.
- MariaDB reports tables created `WITH SYSTEM VERSIONING` (temporal tables) as `table_type = 'SYSTEM VERSIONED'`, not `'BASE TABLE'`.
- Those tables were silently excluded from the query result, so they never appeared in the table explorer/tree — with no error shown anywhere, making it look like the tables didn't exist.
- Fix: include both `'BASE TABLE'` and `'SYSTEM VERSIONED'` in the `table_type` filter.

This looks like it may be the underlying cause of #563 for MariaDB users specifically (that issue's repro was on SQLite, and speculated MySQL/Postgres might also be affected — for MySQL/MariaDB the actual root cause we found is this `table_type` filter, unrelated to reserved-keyword identifier quoting).

## Test plan
- Reproduced against a real MariaDB 10.4.34 database with 134 `BASE TABLE` + 78 `SYSTEM VERSIONED` tables (212 total).
- Before fix: `get_tables()` returned 134 tables.
- After fix: `get_tables()` returns all 212 tables, verified both by running the query directly and by running the compiled app (debug and release builds) against the database — table explorer now lists all 212 tables.
- `cargo check` passes with no new warnings.

- [x] Verified query returns correct count directly against MariaDB
- [x] Verified in the running app (debug build) — sidebar shows "TABELAS (212)"
- [x] Verified in the running app (release build)
- [x] `cargo check` clean


debba (approved):
Thanks for this! The fix is correct and nicely scoped. I reviewed it locally and verified the behavior against a real MariaDB instance (10.11) with a mix of regular tables, a `WITH SYSTEM VERSIONING` table, a view and a sequence: with the old filter the versioned table silently disappeared from the list, with your change it shows up as expected, while views and sequences stay correctly excluded. Also confirmed the query is harmless on MySQL 8.4, since it never reports `SYSTEM VERSIONED` as a table type.

Good catch on the root cause, and the writeup made it really easy to review.

We'd love to see you contribute again, feel free to pick up anything else that catches your eye. And if you want to hang out with the rest of the community, come join us on Discord: https://discord.com/invite/K2hmhfHRSt

PR #575, opened 2026-07-30
feat(plugins): extend RpcDriver for BLOB, materialized views, and type mappings

aesslinger:
Extends the plugin RPC adapter to forward three method categories that were previously unsupported, enabling full PostgreSQL feature parity for plugin drivers.

### 1. BLOB Methods Forwarded
`save_blob_to_file` and `fetch_blob_as_data_url` are now forwarded to plugin drivers via JSON-RPC. Plugins return base64-encoded data; the host handles file I/O. Plugins that don't implement these methods return `-32601` and the host falls back to the existing "not supported" error (no behavioral change for existing plugins).

### 2. Materialized View Methods Forwarded
`get_materialized_views`, `get_materialized_view_columns`, `get_materialized_view_definition`, and `refresh_materialized_view` are now forwarded. Same fallback pattern as above.

### 3. `type_mappings` Manifest Field
Adds an optional `type_mappings: HashMap<String, String>` field to `PluginManifest`. The `RpcDriver` uses this to resolve `map_inferred_type()` locally (this method is synchronous and cannot issue an RPC call). Example: a PostgreSQL plugin declares `{"DATETIME": "TIMESTAMP", "JSON": "JSONB"}`.

## Context

These are prerequisite changes for the PostgreSQL plugin migration initiative (see `.github/planning/postgres-plugin-migration.md` in the test infrastructure PR that follows). Without them, a plugin driver cannot achieve full feature parity with the built-in PostgreSQL driver.

## Testing

- Unit tests added for all new forwarding methods (success + method-not-found fallback)
- Unit tests for `map_inferred_type` with and without manifest mappings
- All existing tests pass unchanged (`cargo test`)
- Backward-compatible: existing plugins without these methods continue to work


aesslinger:
Closing — will split into separate PRs for RPC extensions and test infrastructure.

PR #576, opened 2026-07-30
feat(plugins): extend RpcDriver for BLOB, materialized views, and type mappings

aesslinger:
Extends the plugin RPC adapter to forward three method categories that were previously unsupported, enabling full feature parity for plugin drivers.

Ref #16

### 1. BLOB Methods Forwarded

`save_blob_to_file` and `fetch_blob_as_data_url` are now forwarded to plugin drivers via JSON-RPC. The plugin receives the file path and is responsible for writing the file directly (since it runs on the same machine as the host). Plugins that don't implement these methods return `-32601` and the host falls back to the existing "not supported" error (no behavioral change for existing plugins).

### 2. Materialized View Methods Forwarded

`get_materialized_views`, `get_materialized_view_columns`, `get_materialized_view_definition`, and `refresh_materialized_view` are now forwarded. Same fallback pattern as above.

### 3. `type_mappings` Manifest Field

Adds an optional `type_mappings: HashMap<String, String>` field to `PluginManifest`. The `RpcDriver` uses this to resolve `map_inferred_type()` locally (this method is synchronous and cannot issue an RPC call). Example: a PostgreSQL plugin declares `{"DATETIME": "TIMESTAMP", "JSON": "JSONB"}`.

## Documentation Updates

- `plugins/manifest.schema.json`: `type_mappings` field added (fixes validation for plugins using it)
- `plugins/PLUGIN_GUIDE.md`: All 6 new RPC methods and `type_mappings` fully documented with params, results, and contracts
- `src/types/plugins.ts`: `type_mappings` added to TS `PluginManifest` interface

## Context

These are prerequisite changes for the PostgreSQL plugin migration initiative. Without them, a plugin driver cannot achieve full feature parity with the built-in PostgreSQL driver for BLOB handling, materialized view management, and type inference during paste/import operations.

## Testing

- 22 unit tests covering all new forwarding methods (success path + method-not-found fallback for every method)
- All existing tests pass unchanged (`cargo test`)
- TypeScript compiles cleanly (`pnpm tsc --noEmit`)
- Backward-compatible: existing plugins without these methods continue to work (graceful `-32601` fallback)


debba (commented):
Hey @aesslinger, thanks a lot for this one. I checked out the branch, read through the whole diff, and ran everything locally. This is really solid work: the fallback behavior is carefully aligned with the trait defaults, the tests verify both the exact payloads sent over RPC and the `-32601` fallback paths, and the decision to resolve `map_inferred_type()` locally through the manifest (since the trait method is synchronous and can't issue an RPC call) is exactly the right call. I also confirmed the forwarding is actually reachable, since plugins can declare `materialized_views: true` in their capabilities. Full suite is green on the branch for me: 1051 passed, 0 failed.

I also built a small test plugin implementing the new methods (plus a "legacy" variant that returns `-32601`) and drove it over JSON-RPC stdio with the exact payloads the `RpcDriver` sends. Everything behaves as advertised, including the graceful degradation for old plugins. Nice job on backward compatibility.

A few things I'd like to see addressed before merge, none of them about the code itself:

1. **`plugins/manifest.schema.json` needs the new field.** The schema has `additionalProperties: false` at the top level, so a plugin declaring `type_mappings` in its manifest will currently fail validation in the registry tooling. Adding the property to the schema should be enough.

2. **`PLUGIN_GUIDE.md` should document the new surface.** The six new RPC methods and the `type_mappings` manifest field aren't mentioned yet. Since this PR is the prerequisite for the PostgreSQL plugin migration, the guide is the public contract plugin authors will rely on.

3. **The description doesn't match the code for `save_blob_to_file`.** The PR says "plugins return base64-encoded data; the host handles file I/O", but the implementation forwards `file_path` to the plugin and discards the result, so it's the plugin that writes the file. That's a perfectly fine design given the plugin runs on the same machine, but the description (and ideally the guide) should state the actual contract so plugin authors implement the right thing.

Two small nits, take them or leave them: `get_materialized_view_columns` is the only new method without a fallback test, and the TS `PluginManifest` interface in `src/types/plugins.ts` could gain an optional `type_mappings` field for completeness (the frontend doesn't consume it today, so no urgency). Also a tiny note on the description: it mentions 21 new tests, but I count 13, the other 8 in that filter are pre-existing.

Overall: approve once the schema and docs are updated. Really happy to see the plugin RPC surface reaching parity, this unblocks a lot.

aesslinger:
Hey @debba, thanks for the thorough review and for actually spinning up a test plugin to verify the behavior — that's awesome.

All 6 items addressed in the latest push:

**1. `manifest.schema.json` updated** — Added `type_mappings` as an optional object property with `additionalProperties: { "type": "string" }`. Plugins declaring this field will now pass registry validation.

**2. `PLUGIN_GUIDE.md` updated** — Added full documentation for:
- All 4 materialized view methods (`get_materialized_views`, `get_materialized_view_columns`, `get_materialized_view_definition`, `refresh_materialized_view`) with params, result shapes, and fallback behavior
- Both BLOB methods (`save_blob_to_file`, `fetch_blob_as_data_url`) with the corrected contract (see below) and example payloads
- A new "Type Mappings" section explaining the manifest field, rules, and a PostgreSQL example

**3. Fixed the BLOB contract description** — Both the PR description and the PLUGIN_GUIDE now correctly state that the plugin receives `file_path` and writes the file directly. The host doesn't handle file I/O — this is the right design since the plugin runs locally on the same machine.

**4. Added the missing fallback test** — `rpc_driver_materialized_view_columns_falls_back_when_method_missing` now exists. Every new method has both a success test and a `-32601` fallback test.

**5. TS interface updated** — `type_mappings?: Record<string, string>` added to `PluginManifest` in `src/types/plugins.ts`. No frontend code consumes it today, but it's there for completeness and future use.

**6. Test count corrected** — 22 plugin driver unit tests total (13 new in this PR + 9 pre-existing that were in the filter). Updated the PR description.

Let me know if there's anything else you'd like adjusted.

debba:
LGTM, @NewtTheWolf is there something related to tabularium?

NewtTheWolf:
Pushed review fixes directly to the branch (maintainer edit):

- `manifest.schema.json`: added the `materialized_views` capability — the new docs instruct declaring it, but `additionalProperties: false` rejected it
- `PLUGIN_GUIDE.md`: capability table row added; matview fallback note now matches the actual behavior (getters → empty, definition/refresh → error)
- `RpcDriver::map_inferred_type` now uppercases the input before lookup, as documented ("case-insensitive"), + test case
- `fetch_blob_as_data_url` test mock now returns the documented `BLOB:<size>:<mime>:<base64>` wire format (built-ins produce this via `encode_blob_full`), not a `data:` URL

`cargo test --lib plugins::driver::tests`: 25 passed. Thanks for the PR — solid forwarding work, this was all edge polish.

PR #577, opened 2026-07-30
feat: PostgreSQL plugin reaches parity + safety with built-in driver (#614)

### LINKED ISSUE #614: Hardcoded driver === "postgres" checks in core break the standalone PostgreSQL plugin
## Context

While testing the standalone PostgreSQL plugin from TabularisDB/tabularis-postgresql-plugin#3 installed in a real Tabularis instance, a set of host-side issues emerged: several conditions in the core are still hardcoded on `driver === "postgres"`. They made sense while the driver was builtin, but they no longer hold when PostgreSQL runs as a plugin registered under the `postgresql` id.

The plugin itself behaves correctly (its 82/82 parity is protocol-level, nothing below is a defect of the plugin repo). These are all host-side gaps, and they should be cleaned up before #577 removes the builtin driver, since at that point they stop being edge cases and become the default experience. The investigation was run using Claude; findings below.

### 1. Identifier quoting (`src/utils/identifiers.ts:36`, `shouldQuoteIdentifiers`)

With driver `"postgresql"` it returns `false`, so `formatSqlIdentifier` returns identifiers unquoted. Mixed-case or reserved-word columns produce broken or silently wrong SQL in:

- Visual Query Builder (`src/utils/visualQuery.ts:74,83,91,92`)
- Filter bar WHERE clauses (`src/utils/filterBar.ts:156`)
- Table toolbar structured filters (`src/utils/tableToolbar.ts:123`) and ORDER BY quoting (`formatSortClause`, `src/utils/tableToolbar.ts:105`, called 6 times from `TableToolbar.tsx`)
- Monaco autocomplete insert text (`src/utils/autocomplete.ts:163`)
- Column header sorting (`src/pages/Editor.tsx:2161,2168`)
- Column name replacement in toolbar inputs (`src/components/ui/TableToolbar.tsx:267,337`)

### 2. SSL mode options in the connection form (`src/components/modals/NewConnectionModal.tsx:2958-2972`)

The most serious one. The SSL tab is shown because the plugin declares `supports_ssl: true`, but since the driver string is not `"postgres"`, the dropdown falls into the MySQL-style branch (`disabled/preferred/required/verify_ca/verify_identity`, default `"required"`). The plugin only recognizes the Postgres values: its `needs_tls` matches `require | verify-ca | verify-full`. Result: a connection saved with "Required" or "Verify CA" connects **in cleartext with no error**. Silent failure with security implications.

### 3. MCP backend (`src-tauri/src/mcp/mod.rs:559, 847, 934`)

`conn.params.driver == "postgres"` decides the `"public"` default schema. Plugin connections get `None`, so MCP tools behave differently than with the builtin driver on the same database.

## Working by accident (fragile)

- Direct `quoteIdentifier`/`getQuoteChar` calls (`src/utils/identifiers.ts:9-17`, used by `foreignKeys.ts:96`, `SidebarColumnItem.tsx:60`, `TriggerEditorModal.tsx:123`): unknown drivers fall back to `"` as quote char, which happens to be correct for Postgres. A future MySQL-like plugin would hit the inverse problem here.
- The editor dialect path is correct by design: `Editor.tsx:234` reads `activeCapabilities.sql_dialect` (the plugin declares `"postgres"`), and the splitter/formatter consume that capability, not the driver id.

## Cosmetic only

- `src/utils/driverUI.tsx:27` (icon switch on `manifest.icon`; the plugin declares none, so it gets the generic icon)
- `src/utils/connections.ts:114,217` (default port and label fallbacks)
- `NewConnectionModal.tsx:2524` (3-column instead of 4-column grid)

## Suggested direction

The fix is mostly already available: the manifest exposes `identifier_quote` and `sql_dialect`, and `getQuoteChar` already accepts a `PluginManifest`. Making `shouldQuoteIdentifiers` and the SSL branches capability/dialect-driven instead of driver-string-driven (or normalizing driver to dialect in one place, e.g. `DatabaseProvider`) would cover items 1 and 2. For item 3 the backend can look at `manifest.engine` or a capability instead of the driver id.

It may also be worth folding these into the "manual smoke test" checklist tracked in TabularisDB/tabularis-postgresql-plugin#3, which is exactly the class of gap that checklist is meant to catch.

aesslinger:
Ref #16, closes #614

## Summary

PostgreSQL now ships as a standalone plugin ([`TabularisDB/tabularis-postgresql-plugin`](https://github.com/TabularisDB/tabularis-postgresql-plugin)) with full parity to the built-in driver, plus the host-side fixes needed for the plugin to actually be a safe, correct, first-class alternative — not just parity in isolation. This closes out the migration's Phase 0/1 work as a shippable unit.

**What changed here:**

- **[#614](https://github.com/TabularisDB/tabularis/issues/614) fixed** — several core checks were hardcoded on `driver === "postgres"` and silently misbehaved for the plugin's `"postgresql"` driver id: broken identifier quoting, an SSL-mode dropdown that could silently connect in cleartext, wrong MCP schema defaults. Fixed across 9 commits, each independently verifiable — full breakdown and justifications on the issue thread, including a security-relevant gap found and closed during adversarial review of the fix itself before this was pushed.
- **In-tree plugin copy removed** (`plugins/postgres-plugin/`) — stale relative to the extracted repo's beta releases; the extracted repo is now the source of truth, with its own security audit, CI hardening, and 5-platform release builds.
- **24-item manual smoke test + full `pnpm test`/`cargo test` regression pass — done.** Manual testing against the real plugin binary in the actual desktop UI surfaced one genuine bug outside #614's scope: `execute_query` returned `null` for PostgreSQL enum values instead of the real label. Filed and fixed in the plugin repo ([`tabularis-postgresql-plugin#7`](https://github.com/TabularisDB/tabularis-postgresql-plugin/issues/7)), plus a new parity test here so the bug class can't silently reappear — parity suite is now 83/83 against the real plugin binary (`v1.0.0-beta.3`).
- **Synced with `upstream/main`** (three times — 106 commits, then 15 more, then [PR #588](https://github.com/TabularisDB/tabularis/pull/588)'s merge) with zero unresolved conflicts each time; full regression suite reverified green after every sync. See the comment below for how the #588 overlap was resolved.
- **CI golden-file tautology fixed** — `pg-integration.yml` was setting `REGENERATE_GOLDEN=1` unconditionally, which writes fresh output *before* comparing against it, so the assertion could never catch drift. Regeneration is now `workflow_dispatch`-only and opt-in.
- Cleaned up superseded/out-of-scope planning docs that had accumulated on this long-lived branch.

### Sign-off checklist

| Item | Status | Where |
|---|---|---|
| Security audit | ✅ Done | Plugin repo PR #3 — 4 real gaps found & fixed |
| CI hardening | ✅ Done | Plugin repo — manifest validation, `cargo audit`, release-binary smoke test, live-db integration test |
| Cross-platform build | ✅ Done | Plugin repo `release.yml` — 5-platform matrix, smoke-tested |
| #614 host-side driver checks | ✅ Done | See [issue #614](https://github.com/TabularisDB/tabularis/issues/614) for the full breakdown |
| 24-item manual smoke test | ✅ Done | Real plugin binary, real desktop UI, every result cross-checked directly against the database via `psql` — not just the UI's own success indicators |
| `pnpm test` / `cargo test` regression | ✅ Done | Frontend and backend suites green (only pre-existing, unrelated failures, confirmed present on unmodified `main`) |

---

## Not in this release

- **Phase 2** (sequences, JSONB editing, extension type system — issue #16's remaining scope) happens in the plugin repo directly, tracked at [`tabularis-postgresql-plugin#9`](https://github.com/TabularisDB/tabularis-postgresql-plugin/issues/9).
- **Phase 3** (removing the built-in driver) is a full-team-consensus decision, not something this PR decides. Tracked at [tabularis#631](https://github.com/TabularisDB/tabularis/issues/631), which also covers deleting the `postgres_integration` test suite and this PR's `pg-integration.yml` workflow once the built-in driver they protect is actually gone.
- **Registry publication** (`plugins/registry.json` entry + a local-file install path for `install_plugin`) — the plugin isn't installable via `Settings > Plugins` yet. Out of scope here; not yet tracked by an issue.

---

## How to Validate Locally

```bash
# Start PG 16
docker run -d --name pg-parity -p 54320:5432 \
  -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=password \
  -e POSTGRES_DB=testdb postgres:16

# Seed
bash tests/fixtures/seed_postgres.sh

# Run baseline + golden tests (builtin only — must pass)
cd src-tauri && cargo test --test postgres_integration -- --include-ignored --test-threads=1
```

To re-check cross-repo parity against a real plugin build (manual, not CI):

```bash
# In a checkout of TabularisDB/tabularis-postgresql-plugin
cargo build --release

# Back in this repo
POSTGRES_PLUGIN_BIN=/path/to/tabularis-postgresql-plugin/target/release/postgresql-plugin \
  cargo test --test postgres_integration parity -- --include-ignored --test-threads=1
# Expect 83/83
```

---

## Depends On

- PR #576 (RPC driver extensions) — same commits included here


aesslinger:
## `upstream/main` synced — [PR #588](https://github.com/TabularisDB/tabularis/pull/588) merged as a prerequisite, conflict resolved per @debba's instructions

[PR #588](https://github.com/TabularisDB/tabularis/pull/588) (`fix-postgres-vqb-followups`) merged to `main` and has now been merged into this branch. As flagged in [my earlier comment on that PR](https://github.com/TabularisDB/tabularis/pull/588#issuecomment-5280231249), it touched the same `shouldQuoteIdentifiers` function in `src/utils/identifiers.ts` that #614's fix changed here, so the merge produced real conflicts in 3 files (`src/utils/identifiers.ts`, `tests/utils/identifiers.test.ts`, `tests/utils/visualQuery.test.ts`; `src/utils/visualQuery.ts` auto-merged but needed a follow-up pass — see below).

[@debba's resolution decision](https://github.com/TabularisDB/tabularis/pull/588#issuecomment-5283010222): merge #588 first (its HAVING/alias-quoting fixes are net-new, real bugs users hit today, no reason to hold them behind this larger migration), then resolve the conflict here by keeping the `sql_dialect`-based capability check as the long-term version of `shouldQuoteIdentifiers`, **making sure the literal-string fallback still covers `"postgresql"`** so #588's original fix isn't lost, and widening the new helper functions' `driver` params in the same pass.

**What was done, matching each instruction:**

1. **Kept the `sql_dialect`-based check as primary.** `shouldQuoteIdentifiers` still checks `capabilities.sql_dialect` first when a manifest/capabilities object is available — unchanged from this branch's #614 fix.
2. **Fixed the string fallback to cover `"postgresql"`.** The initial merge resolution (mine, before catching this) had left the no-manifest fallback as bare `driver === "postgres"` — which would have silently dropped #588's fix for any caller that only has a driver id string in scope, no manifest. Caught this against debba's comment before committing and fixed it to `driver === "postgres" || driver === "postgresql"`. Updated the one test that had asserted the old (wrong) behavior.
3. **Widened the new helpers in the same pass.** `formatAggregateArgument`, `formatHavingColumnRef`, `formatAlias` (all net-new from #588), and `generateHavingClause`'s `driver` param were typed `string | null | undefined` after the auto-merge — narrower than this branch's `DriverArg` union (`string | PluginManifest | DriverCapabilities | null | undefined`). Widened all four to `DriverArg` so they inherit the same capability-driven quoting as everything else in the module.
4. **Verified end-to-end**, not just by type-checking: added a throwaway test (run, confirmed passing, then deleted — never committed) proving a postgres-dialect plugin manifest produces byte-identical `HAVING` clause output to the bare `"postgres"` string via `generateHavingClause`, and that `"mysql"` correctly stays unquoted. This is the exact scenario #614 exists to fix — a plugin driver id that isn't the literal string `"postgres"` — now proven to work for #588's new HAVING/alias code too, not just the pre-existing quoting paths.

**Verification after resolving:**
- `pnpm tsc --noEmit` — clean
- `pnpm vitest run tests/utils/identifiers.test.ts tests/utils/visualQuery.test.ts` — 110/110 passing (both PRs' test cases preserved, nothing dropped from either side)
- Full frontend suite: 3717/3750 passing (33 failures are pre-existing/environmental — `localStorage.clear()` undefined in this local test environment — independently confirmed present on a clean, unmodified `upstream/main` checkout before this merge, unrelated to either PR)
- `pnpm tsc --noEmit` + full suite re-run post-commit to confirm the committed state matches what was verified pre-commit

aesslinger:
@debba — this is ready for your final review.

**Summary:** PostgreSQL now ships as a standalone plugin with full parity to the built-in driver, plus the host-side fixes ([#614](https://github.com/TabularisDB/tabularis/issues/614)) needed for the plugin to be a safe, correct, first-class alternative — not just parity in isolation.

**Status:**
- #614 fixed (9 commits, each independently verifiable — identifier quoting, SSL-mode dropdown, MCP schema defaults, plus a security-relevant gap found and closed during self-review)
- 24-item manual smoke test done against the real plugin binary in the actual desktop UI, every result cross-checked directly against the database
- Full `pnpm test` / `cargo test` regression pass green (only pre-existing, unrelated failures)
- Synced with `main` three times, most recently to pull in [#588](https://github.com/TabularisDB/tabularis/pull/588) — that conflict was resolved per your instructions on that PR (kept the `sql_dialect`-based check, restored the `"postgresql"` string fallback, widened the new HAVING/alias helpers to the same capability-driven type); details in [this comment](https://github.com/TabularisDB/tabularis/pull/577#issuecomment-5283205488)
- CI green (`test`, `test-postgres`), no merge conflicts with `main`

Deferred/out-of-scope items (Phase 2, Phase 3 deprecation decision, registry publication) are tracked separately — see the "Not in this release" section in the PR description.

Let me know if you'd like anything else looked at before merging.

debba:
@aesslinger Great work overall. I verified this PR locally against PostgreSQL 16 and the actual `v1.0.0-beta.3` Linux release binary: the cross-driver parity suite is genuinely **83/83 green**, and the full frontend/backend suites pass.

Before approval, I think these items should be addressed:

1. **The parity harness can false-green when the plugin path is wrong.** In `src-tauri/tests/postgres_integration/parity.rs`, setting `POSTGRES_PLUGIN_BIN` to a nonexistent file logs a warning, skips the plugin, and runs only against the builtin driver. I reproduced this with:
   ```bash
   POSTGRES_PLUGIN_BIN=/tmp/definitely-missing-plugin \
     cargo test --test postgres_integration parity_tests::parity_get_databases \
     -- --include-ignored --test-threads=1
   ```
   The test passes. If the variable is explicitly set, a missing/unstartable binary should fail immediately.

2. **Missing golden files are silently accepted.** `assert_golden()` in `golden_utils.rs` returns successfully when a fixture does not exist. Outside explicit regeneration mode, a missing/renamed golden file should fail so accidental deletions and filename typos cannot leave CI green.

3. **Four React hook dependency warnings remain in `Editor.tsx`.** The callbacks now consume the full `activeCapabilities` object but only list `activeCapabilities?.schemas` in their dependency arrays (around lines 345, 1145, 1540, and 2175). ESLint reports four `react-hooks/exhaustive-deps` warnings; this also conflicts with `.rules/react.md` rule 1 and can retain stale quoting capabilities.

4. **Formatting/warning cleanup:** `cargo fmt --all -- --check` currently fails on PR-added/modified Rust code, the PostgreSQL integration target emits 50 unused-import/variable warnings, and `git diff --check` reports an extra blank line at EOF in `src-tauri/src/mcp/tests.rs`.

Verification performed:
- `pnpm test -- --run`: **3750/3750 passed**
- `cargo test --lib`: **1142 passed, 4 ignored**
- PostgreSQL 16 integration suite: **181/181 passed**
- Actual plugin `v1.0.0-beta.3` parity: **83/83 passed**
- `pnpm exec tsc --noEmit`: passed

The functional work looks solid; my request-changes recommendation is about making the new safety gates reliably fail when their prerequisites or fixtures are missing, plus the repository-rule cleanup above.

debba:
@aesslinger One additional, non-blocking question: do you want to keep the `.github/planning/` documents in the repository after this PR merges as historical/project documentation, or would you prefer to remove the completed migration plans before merge? They would remain available in Git history either way, while removing them would keep the current tree focused on active documentation. I do not have a strong preference, but it would be useful to make that choice explicit.

aesslinger:
@debba — thank you for the thorough review, and sorry that #3 and #4 needed to be caught by you at all. Both were gaps in my own verification discipline (I was running `tsc`/`cargo build`/`cargo test` at every checkpoint but not `eslint`/`cargo fmt --check` alongside them) rather than anything ambiguous or hard to check — I've corrected that going forward and will run the full lint/fmt/warning gate every time, not just the compile/test gate.

Here's how each of your 5 items was addressed:

**1. Parity harness false-greens on a bad `POSTGRES_PLUGIN_BIN`.**
`try_plugin_driver()` in `parity.rs` now panics when the env var is set but the path doesn't exist, instead of silently falling back to builtin-only. Verified with your exact repro command — it now fails as expected instead of passing.

**2. `assert_golden()` silently accepted missing golden files.**
Now panics with the actual captured output included in the message (so adding a new golden case doesn't require a second run with `REGENERATE_GOLDEN=1` just to see what to commit). Verified by temporarily renaming `get_schemas.json`, confirming the test now fails instead of skipping, then restoring it and rerunning the full 181-test suite clean.

**3. 4 `react-hooks/exhaustive-deps` warnings in `Editor.tsx` (lines 345, 1145, 1540, 2175).**
All 4 dependency arrays now list the full `activeCapabilities` object instead of just `activeCapabilities?.schemas`, matching what each callback body actually reads. Verified with `npx eslint`/`pnpm lint` — zero warnings, was 4.

**4. Formatting/warnings.**
- The entire new `postgres_integration/` test suite (38 files, never run through `cargo fmt` before) is now formatted, and 15 files' unused `DatabaseDriver` imports are removed — `cargo build --all-targets` now reports 0 warnings, was 50.
- The EOF blank line in `mcp/tests.rs` is stripped (`git diff --check` clean).
- One line in `commands.rs` (`dialects.insert(...)`) wrapped to satisfy rustfmt.
- **Found but not fixed here, flagged as a possible followup:** `commands.rs`, `driver_trait.rs`, `mysql/mod.rs`, `postgres/mod.rs`, `sqlite/mod.rs`, and `mcp/mod.rs` each carry pre-existing `cargo fmt` drift unrelated to this branch — confirmed present on a clean `upstream/main` checkout before this PR, so it's not something this PR introduced. I deliberately didn't reformat those whole files since it felt out of scope for a fix PR (would bury our actual changes in unrelated diff noise), but wanted to surface it rather than silently leave it. Want a separate repo-wide `cargo fmt` PR to clear that debt, or leave it as-is for now?

**5. `.github/planning/` docs.**
Removed — Phase 1 planning is done and Phase 2/3 now live in linked issues ([`tabularis-postgresql-plugin#9`](https://github.com/TabularisDB/tabularis-postgresql-plugin/issues/9), [#631](https://github.com/TabularisDB/tabularis/issues/631)) rather than static docs on this branch. Still in git history.

**One new item surfaced during a final safety pass, filed as a follow-up rather than blocking this PR:** [#639](https://github.com/TabularisDB/tabularis/issues/639) — the SSL-mode migration this PR adds only runs from the two GUI-process Tauri commands, never from the `--mcp` server process, which reads the same `connections.json` and can establish real connections using an unmigrated stale `ssl_mode` value. It's a pre-existing gap (the MCP process has never run *any* connection migration, including the older SSH one) that this PR's fix doesn't happen to close — not a regression this PR introduces. Given the scope of fixing it properly (extracting a path-based migration core usable from both the Tauri-command and MCP-bootstrap contexts), I think it's worth deciding post-merge rather than expanding this PR further. Your call on priority/timing.

Everything above is pushed. Full verification after all fixes: `pnpm lint` clean, `pnpm tsc --noEmit` clean, `cargo build --all-targets` 0 warnings, `cargo fmt --check`/`cargo clippy` clean on every file this PR touches, backend 1135+181+83 all green, frontend 3717/3750 (33 pre-existing/unrelated, confirmed present on clean `upstream/main`).

debba:
Everything looks good to me now. I’d proceed with the merge and then start looking into #639, unless you’d prefer to take care of it yourself. Thanks for addressing all the feedback so thoroughly!

aesslinger:
> Everything looks good to me now. I’d proceed with the merge and then start looking into #639, unless you’d prefer to take care of it yourself. Thanks for addressing all the feedback so thoroughly!

I think merge this and then I can start on the follow-ups.

PR #580, opened 2026-07-31
fix: support parameters in explain plan

### LINKED ISSUE #565: [Bug]: Explain does not work with Parameters in the query
### Describe the bug

When clicking "Explain Plan" when you have parameters in your query, you get an error message

Image

### To Reproduce

in the console put a query with parameters in it:
```
select * from test where company = :chosen_name
```

Click "Explain Plan".

Error appears.

Image

### OS Version

Windows 11

### Tabularis Version

0.17.0

### Relevant Log Output

```shell

```

DhruvShah-Dev:
## Summary
- Reuse the existing query parameter modal when Visual Explain is opened for SQL with missing parameters.
- Reuse saved tab parameter values when they are already complete.
- Save submitted values back to the tab before opening the interpolated Explain query.

## Validation
- `pnpm.cmd typecheck`
- `pnpm.cmd lint`

Fixes #565

### DIFF
--- a/src/components/modals/QueryParamsModal.tsx
+++ b/src/components/modals/QueryParamsModal.tsx
@@ -108,7 +108,7 @@ const QueryParamsForm = ({ parameters, initialValues, onSubmit, onClose, mode }:
                 <Save size={16} fill="currentColor" />
               )}
               {mode === "explain"
-                ? t("editor.explain")
+                ? t("editor.visualExplain.buttonShort")
                 : mode === "run"
                   ? t("editor.run")
                   : t("common.save")}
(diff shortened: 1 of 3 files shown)


debba:
Thanks @DhruvShah-Dev, great work! Reusing the run flow's parameter handling for Explain Plan is exactly the right approach, and saving the submitted values back to the tab makes the two flows consistent.

I pushed one small fix on top: the submit button used `t("editor.explain")`, but that key does not exist in any locale, so the button rendered the raw key string. I switched it to the existing `editor.visualExplain.buttonShort` key ("Explain Plan"), which is already translated in all 11 locales.

Verified locally against MySQL: a parameterized query that previously failed EXPLAIN with ERROR 1064 now prompts for values and opens the plan correctly. Merging.

PR #586, opened 2026-08-01
fix(editor): keep query results when switching between connections

### LINKED ISSUE #292: [Bug]: Query results disappear when switching away from and back to a connection
### Describe the bug

With two connections open, query results are lost when you switch to another connection and then switch back. The query tab still shows its SQL, but the result grid comes back empty.

### To Reproduce

1. Open two connections (e.g. PostgreSQL **and** MySQL).
2. In the PostgreSQL editor, run a query → results appear. ✅
3. Switch to the MySQL connection and run any query.
4. Switch back to the PostgreSQL connection.
5. The PostgreSQL result grid is now empty — the results are gone (the SQL in the tab is still there).

**Expected:** the PostgreSQL tab still shows its previous result set.
**Actual:** the result grid is empty; the SQL is preserved, only the results vanish.

### Likely cause

The tab-loading effect re-runs on **every** connection switch and reloads that connection's tabs from file storage, replacing the in-memory tab objects:

- [`EditorProvider.tsx` — load effect keyed on `[activeConnectionId]`](https://github.com/TabularisDB/tabularis/blob/4bd6e1dd4a7e7d5d5a3ebc24a930c19c3cfca4b3/src/contexts/EditorProvider.tsx#L44-L113)
- [the clobber that replaces this connection's live tabs with the loaded ones](https://github.com/TabularisDB/tabularis/blob/4bd6e1dd4a7e7d5d5a3ebc24a930c19c3cfca4b3/src/contexts/EditorProvider.tsx#L76-L77)

But `saveTabsToStorage` deliberately strips transient state (results, errors) before persisting:

- [`editor.ts` — tabs are cleaned with `result: null` before saving](https://github.com/TabularisDB/tabularis/blob/4bd6e1dd4a7e7d5d5a3ebc24a930c19c3cfca4b3/src/utils/editor.ts#L64-L88)

So switching back to an already-loaded connection overwrites its live tabs (which hold the results) with result-less storage copies. Possible fix: skip the storage reload when the connection's tabs are already in memory (e.g. guard on `prev.some(t => t.connectionId === activeConnectionId)`), or merge the loaded tabs while preserving the in-memory `result`.

### OS Version

Linux (CachyOS)

### Tabularis Version

v0.13.0

iamthenuggetman:
## Summary

Closes #292

With two connections open, switching to another connection and back wiped the first connection's query results: the tab kept its SQL, but the result grid came back empty.

**Root cause** (as suspected in the issue): the tab-loading effect in `EditorProvider` re-runs on *every* `activeConnectionId` change and replaces the connection's live tabs with the storage copies — and tabs are persisted result-stripped (`cleanTabForStorage` removes `result`, `error`, `executionTime`, etc.). So switching back to an already-loaded connection clobbered live tabs holding results with result-less storage copies.

## Fix

Skip the storage reload when the connection's tabs are already live in memory (`EditorProvider.tsx`). During a session the in-memory tab state is the source of truth — the continuous save effect (`tabs`/`activeTabIds` change → `saveTabsToStorage`) already keeps storage in sync for the next app start, so nothing is lost by not reloading.

This implements the issue's first suggested approach (guard on existing tabs). The merge-instead-of-reload alternative was unnecessary: the only reason to read storage is to restore tabs that aren't in memory, which by definition only happens on the first visit to a connection.

Side benefit: switching back to a loaded connection no longer flashes the loading state or re-runs the storage round-trip at all.

## What doesn't change

- First visit to a connection (including app start) still loads tabs from storage, including the legacy notebook-tab migration.
- Closing all of a connection's tabs and revisiting still loads from storage (in-memory guard finds no tabs) and creates the initial tab.
- Cross-connection notebook open (`pendingNotebookRef`) is unaffected: with the target connection already loaded, the deferred open resolves immediately instead of waiting out a redundant load.

## Test plan

- New regression test in `tests/contexts/EditorProvider.test.tsx` simulating the issue's exact repro: load conn-1's tab from storage → set a live result → switch to conn-2 (initial tab created) → switch back. Asserts the tab keeps its query *and* its result, and that `load_editor_preferences` ran only once for conn-1. **Verified it fails without the fix and passes with it.**
- `pnpm vitest run`: full suite passes (3431 tests)
- `tsc -b`: clean
- `eslint` on changed source: clean
- Manually verified with the `demo/` Docker stack (MySQL + PostgreSQL connections): results survive switching away and back; tabs still restore correctly from storage after an app restart.

### DIFF
--- a/src/contexts/EditorProvider.tsx
+++ b/src/contexts/EditorProvider.tsx
@@ -58,6 +58,14 @@ export const EditorProvider = ({ children }: { children: ReactNode }) => {
       return;
     }
 
+    // This connection's tabs are already live in memory. Storage copies are
+    // saved result-stripped, so reloading them here would wipe the query
+    // results of every tab whenever the user switches connections (#292).
+    if (tabsRef.current.some((t) => t.connectionId === activeConnectionId)) {
+      setIsLoading(false);
+      return;
+    }
+
     const loadPreferences = async () => {
       setIsLoading(true);
       try {
--- a/tests/contexts/EditorProvider.test.tsx
+++ b/tests/contexts/EditorProvider.test.tsx
@@ -506,3 +506,112 @@ describe("EditorProvider", () => {
     expect(resultConn2.current.tabs).toHaveLength(0);
   });
 });
+
+describe("EditorProvider connection switching (#292)", () => {
+  const storage: Record<
+    string,
+    { tabs: Record<string, unknown>[]; active_tab_id: string | null } | null
+  > = {};
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+    for (const key of Object.keys(storage)) delete storage[key];
+    vi.mocked(invoke).mockImplementation(
+      (cmd: string, args?: { connectionId?: string }) => {
+        if (cmd === "load_editor_preferences") {
+          return Promise.resolve(
+            args?.connectionId ? (storage[args.connectionId] ?? null) : null,
+          );
+        }
+        if (cmd === "save_editor_preferences") {
+          return Promise.resolve(undefined);
+        }
+        return Promise.resolve(null);
+      },
+    );
+  });
+
+  it("keeps query results when switching away and back to a connection", async () => {
+    storage["conn-1"] = {
+      tabs: [
+        {
+          id: "tab-1",
+          title: "Console",
+          type: "console",
+          query: "select * from users",
+          page: 1,
+          activeTable: null,
+          pkColumns: null,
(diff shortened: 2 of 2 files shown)


debba (approved):
Thanks for the excellent work on this one. The root cause analysis is spot on, the guard is minimal and in exactly the right place, and the regression test genuinely catches the bug (I verified it fails without the fix). Full suite passes locally and I confirmed manually with the demo Docker stack that results now survive switching between connections, with tab restore after restart still working as expected. Much appreciated!

Issue #483, opened 2026-07-15
[Feat]: Export data with column anonymization


pokertour:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Add an optional anonymization step to the file export: in the export dialog, let the user assign a per-column rule before writing the file. Proposed rules for a v1:

- **Null / fixed value** — replace with `NULL` or `***`
- **Partial masking** — e.g. `j***@***.com`, keep first/last chars
- **Deterministic hash (pseudonymization)** — HMAC with a per-export key: the same input always produces the same output, so joins on anonymized keys (e.g. `user_id`) still work across exported tables

Possible follow-ups: synthetic data (seeded faker for names/emails/addresses), generalization (birth date → year, zip → region), name-based heuristics to pre-suggest rules for columns like `email`, `phone`, `name`, `ip`.

Technically this fits nicely in the existing streaming export pipeline: a transform layer between the row stream and the CSV/JSON sinks, so it would work for all drivers (including plugins) with no driver changes. The database dump could adopt it later as a second step.

### Motivation

Very common need: sharing production data with staging environments, external developers, support tickets or analytics without leaking personal data (GDPR / privacy). Today users have to post-process exports manually.

Note on wording: this should be presented as *column anonymization / pseudonymization*, not "GDPR compliance" — true anonymization under GDPR is a legal assessment (quasi-identifiers etc.) that no tool can guarantee by itself.

### Example

Export dialog → "Anonymize columns" (optional) → per-column rule selector → file is written with transformed values, e.g. `email` → `a3f9…@masked`, `name` → `***`, `user_id` → HMAC hash (stable across tables).

---

Happy to work on a PR if there's interest — opening this first to discuss scope and UX.

debba:
Feel free to open a PR for this :)

PR #587, opened 2026-08-01
feat(datagrid): mask sensitive columns in the results grid

### LINKED ISSUE #485: [Feat]: Mask sensitive columns in the results grid
### Prerequisites

### 🚀 Feature Proposal

Complementary to #483 (export anonymization), but on the **display** side: automatically mask the values of sensitive columns in the results grid (`••••••` / blurred), so screen sharing, demos, support sessions or working in public places don't leak personal data.

**Configuration (Settings → Privacy):**
- Global toggle: enable/disable column masking.
- A default list of **generic column-name patterns** (editable): `password`, `email`, `phone`, `ssn`, `token`, `secret`, `api_key`, `iban`, `credit_card`, … Matching is case-insensitive on the column name.
- **Per-connection overrides**: add or exclude specific `table.column` entries — so a false positive (e.g. a non-sensitive `email_count`) can be unmasked, and a column the patterns miss (e.g. `contact_info`) can be masked for that database only.

**Reveal interactions (both):**
- **Per cell**: click (or an eye icon) temporarily reveals a single masked cell;
- **Per column**: an eye toggle in the column header reveals/masks the whole column at once.

**Scope: display only.** Copy and export keep returning the real values — masking is protection against shoulder-surfing and screen shares, not data anonymization (that's #483). This keeps existing copy/export workflows predictable.

### Motivation

Anyone who shares their screen (pair debugging, demos, support calls, streaming) with production data visible has hit this: one `SELECT *` on a users table and emails/phones are on screen. Same privacy angle as #483, complementary scopes: #483 protects data leaving the app, this protects data displayed by the app.

### Example

`SELECT * FROM users` → `password` and `email` columns render as `••••••` with an eye icon in the header; clicking a single cell reveals it; Settings → Privacy lets you toggle the feature, edit the patterns, and manage per-connection `table.column` exceptions.

---

Happy to work on a PR if there's interest.

iamthenuggetman:
Closes #485

## What

Columns whose name matches a sensitive pattern (password, email, token, ssn, …) now render as a `••••••` placeholder in the results grid instead of the real value.

- **Per-cell reveal**: masked cells show an eye button that reveals just that cell; a revealed cell gets an eye-off button to re-mask it
- **Per-column reveal**: masked column headers show an eye toggle that reveals / re-masks the whole column
- **Edit guard**: masked cells can't be edited (double-click, Enter, F2) until revealed, and the hover tooltip is suppressed so it can't leak the value
- Reveal state is grid-local and resets when the result data changes

**Masking is display-only**: copy (⌘/Ctrl+C, context menu) and export keep the real values, as requested in the issue. Write-path anonymization for exports is tracked separately in #483.

## Settings

New **Settings → Privacy** tab:

- On/off toggle for masking (default: on)
- Editable column-name patterns (one per line, case-insensitive substring match)
- Per-connection include/exclude overrides as `table.column` entries — `exclude` wins over `include`, which wins over name patterns

## Implementation

- `src/utils/columnMasking.ts` — pattern/overrides matching logic
- `DataGrid.tsx` / `DataGridRow.tsx` — masked rendering + reveal toggles
- `SettingsContext.ts` — `columnMaskingEnabled` / `columnMaskingPatterns` / `columnMaskingOverrides` settings
- `PrivacyTab.tsx` + tab registration in `Settings.tsx`
- i18n strings for all 11 locales

## Tests

- 8 unit tests for the matching logic (`tests/utils/columnMasking.test.ts`)
- 5 DataGrid tests: placeholder rendering, column reveal/re-mask, per-cell reveal/re-mask, copy keeps the real value, edit guard on masked cells
- Full suite green (3443 tests), `tsc -b` and eslint clean


debba:
First of all, I really like this feature. The display-only approach with per-cell and per-column reveal is exactly what #485 asked for, and the test coverage is appreciated.

Before merging I'd like to request a few changes, mostly about consistency with the rest of the app:

1. **Use the shared `Select` component.** The Privacy tab introduces a plain native `<select>` with its own styling, while the rest of the settings UI uses the shared `Select` component. Please switch to that one so the look and behavior stay consistent.

2. **Tab position.** In the settings sidebar I'd place Privacy right after Appearance instead of right after General.

3. **Per-connection overrides in the connection modal.** Since the include/exclude overrides are per-connection, it should also be possible to manage them from a Privacy tab inside the connection modal, not only from the global settings. Please reuse what the settings tab already provides rather than duplicating it.

The core masking logic looks solid, so nothing to change there. Happy to merge once these points are addressed!

iamthenuggetman:
thanks @debba i've made the requested tweaks and will be pushing shortly.

debba:
Looks ok to me.
Maybe in the future we can add some features for enabling / disabling mask only a specific connection.
For now I will merge it, thanks a lot!

PR #596, opened 2026-08-04
feat(connections): opaque plugin-specific extra fields for ConnectionParams

fuleinist:
Follow-up to the remaining core-side work from https://github.com/TabularisDB/tabularis/pull/593#issuecomment-5172749627 — generic support for plugin-owned connection fields, not just DynamoDB.

## What

- **`extra: HashMap<String, String>`** on core `ConnectionParams` (Rust) + `extra?: Record<string, string>` (TS). Opaque to the host: persisted verbatim through `save_connection`/`update_connection`, forwarded to driver plugins as part of `params`, omitted from persisted JSON when empty (`serde(default, skip_serializing_if)` — legacy JSON deserializes to an empty map).
- **New plugin slot `connection-modal.extra_fields`**, rendered below the host/port section of the connection form. Context exposes `driver`, the current `extra` map, and `setExtraField(key, value)` so a plugin UI can edit its own fields (empty value clears the key).
- **`updateExtraField`** pure helper in `src/utils/connections.ts` with 8 unit tests (set / merge / immutability / clear-on-empty / blank-key / trim).
- Host API version **0.1.0 → 0.2.0** (`HOST_API_VERSION`, plugin-api `API_VERSION` + `package.json`) — the slot map is observable to plugin bundles.
- `PLUGIN_GUIDE.md` slot table updated.

## Verification

- `vitest`: **PASS (3490) FAIL (0)** (incl. new helper tests)
- `tsc --noEmit`: clean
- `eslint` on all touched files: 0 errors
- `cargo clippy --all-targets`: clean, no new warnings in touched files
- plugin-api `check:sync`: OK
- `rustfmt` on touched files: no new diffs vs upstream style

Note on local Rust test execution: on this Windows box the compiled test binary fails at load with `STATUS_ENTRYPOINT_NOT_FOUND` from an old system `comctl32.dll` missing `TaskDialogIndirect` — reproduced identically on pristine upstream code, so it's a local-environment issue unrelated to this change; CI (Linux) runs the full Rust suite.

## Companion PR

DynamoDB plugin side (consumes `extra["region"]`): TabularisDB/tabularis-dynamodb-plugin#59

### DIFF
--- a/packages/plugin-api/package.json
+++ b/packages/plugin-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tabularis/plugin-api",
-  "version": "0.2.0",
+  "version": "0.1.1",
   "description": "Public API surface for Tabularis plugin UI extensions.",
   "license": "Apache-2.0",
   "homepage": "https://github.com/TabularisDB/tabularis/tree/main/packages/plugin-api",
--- a/packages/plugin-api/package.json
+++ b/packages/plugin-api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tabularis/plugin-api",
-  "version": "0.1.0",
+  "version": "0.2.0",
   "description": "Public API surface for Tabularis plugin UI extensions.",
   "license": "Apache-2.0",
   "homepage": "https://github.com/TabularisDB/tabularis/tree/main/packages/plugin-api",
--- a/packages/plugin-api/src/version.ts
+++ b/packages/plugin-api/src/version.ts
@@ -2,7 +2,7 @@
  * API version of this package. Must match the version field of package.json.
  * Bump when the host API shape changes in a way that plugin bundles can observe.
  */
-export const API_VERSION = "0.2.0";
+export const API_VERSION = "0.1.1";
 
 /**
  * Minimum Tabularis host version that exposes an API compatible with this package.
--- a/packages/plugin-api/src/version.ts
+++ b/packages/plugin-api/src/version.ts
@@ -2,7 +2,7 @@
  * API version of this package. Must match the version field of package.json.
  * Bump when the host API shape changes in a way that plugin bundles can observe.
  */
-export const API_VERSION = "0.1.0";
+export const API_VERSION = "0.2.0";
 
 /**
  * Minimum Tabularis host version that exposes an API compatible with this package.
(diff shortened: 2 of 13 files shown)


packages/plugin-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tabularis/plugin-api",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
fuleinist:
```suggestion
  "version": "0.1.1",
```

packages/plugin-api/src/version.ts
```diff
@@ -2,7 +2,7 @@
  * API version of this package. Must match the version field of package.json.
  * Bump when the host API shape changes in a way that plugin bundles can observe.
  */
-export const API_VERSION = "0.1.0";
+export const API_VERSION = "0.2.0";
```
fuleinist:
```suggestion
export const API_VERSION = "0.1.1";
```

src/contexts/PluginSlotProvider.tsx
```diff
@@ -24,7 +24,7 @@ interface PluginSlotProviderProps {
  * Keep in sync with `packages/plugin-api/src/version.ts` (API_VERSION).
  * Bump when the host API shape changes in a way plugin bundles can observe.
  */
-const HOST_API_VERSION = "0.1.0";
+const HOST_API_VERSION = "0.2.0";
```
fuleinist:
```suggestion
const HOST_API_VERSION = "0.1.1";
```

debba (approved):
Reviewed and tested this locally on Linux.

Full run on the branch: vitest 3490 passed across 194 files, `cargo test --lib` 1089 passed / 0 failed / 4 ignored, `tsc --noEmit` clean, eslint clean on the touched files, and clippy quiet in `models.rs` and `plugins/driver.rs`. So the Rust suite you couldn't run on that Windows box is green here, nothing to worry about on that front.

I checked the plumbing instead of just reading the diff. `params_for_persistence` clones the map so it survives save and update, `duplicate_connection` keeps it, every plugin RPC ships the whole params struct so drivers get `extra` on each call, and the edit path spreads the stored params back into the form, so opening and re-saving a connection doesn't quietly drop plugin fields. Legacy `connections.json` files still load fine and stay byte identical while the map is empty.

To exercise the new slot for real I wrote a small stub UI extension against `connection-modal.extra_fields` and drove it on a live MySQL connection: prefill from `extra`, per key edits, clear on empty, and the key disappearing from the persisted JSON once the last entry is gone all behave exactly as documented.

Clean, self contained design. Keeping driver specifics out of the core schema is the right call. Thanks!

PR #597, opened 2026-08-04
Fix quick navigator database selection

### LINKED ISSUE #591: Object palette reads the database list from connection params instead of the live selection
The object command palette derives its database list from the saved connection parameter, while every other consumer uses the list the provider actually resolved. The two drift apart.

`src/hooks/useCommandPaletteObjectItems.ts:66-69`:

```ts
const configuredDatabases = useMemo(
  () => getDatabaseList(connection?.params.database ?? ""),
  [connection?.params.database],
);
```

The sidebar instead reads `connectionData.selectedDatabases` (`src/components/layout/ExplorerSidebar.tsx:140`), which the provider keeps current. Neither `setSelectedDatabases` (`src/contexts/DatabaseProvider.tsx:566`) nor `refreshDatabaseSelection` (`src/contexts/DatabaseProvider.tsx:149`) writes back to `params.database`.

### Consequences

**1. All-databases mode (#572): the object palette is empty.**

In that mode `params.database` is empty by design — the real list comes from `get_available_databases` at connect time (`src/contexts/DatabaseProvider.tsx:676-687`). So `configuredDatabases` resolves to `[]`, and `getNavigatorItems` builds its groups exclusively from that array:

```ts
// src/utils/quickNavigator.ts:125-129
} else if (isMultiDb) {
  groups = configuredDatabases.flatMap((group) => {
    const data = databaseDataMap[group];
    return data ? [{ group, data }] : [];
  });
```

An empty array yields no groups and therefore no items. The eager-load effect (`src/hooks/useCommandPaletteObjectItems.ts:85-89`) iterates the same empty array, so nothing recovers it.

**2. Stale list after a selection change.**

Changing the database selection in the sidebar, the manual refresh added in #530, or the pruning of dropped databases added in #524 all update `selectedDatabases` only — the palette keeps offering the old set.

**3. Divergent branch condition.**

The palette uses `isMultiDatabaseCapable(capabilities)` (capability only), while the sidebar uses `usesMultiDatabaseLayout(capabilities, selectedDatabases)` — capability *and* a non-empty selection (`src/utils/database.ts:50-55`). The two can take different layout branches for the same connection.

### Suggested fix

Read `connectionData.selectedDatabases` instead of `getDatabaseList(connection.params.database)`, and align the multi-database check with `usesMultiDatabaseLayout`.

### Notes

Pre-existing, not introduced by #545 — filed separately as agreed in that PR's review rather than growing its diff.

Found by reading the code; reproducing it needs a MySQL/MariaDB connection with an empty database field, since schema-based drivers take a different branch.

DhruvShah-Dev:
## Summary
- use the provider's live selectedDatabases list for quick navigator multi-database items
- align quick navigator multi-database branching with usesMultiDatabaseLayout
- add regression coverage for stale database data being ignored

## Testing
- pnpm.cmd exec vitest run tests/utils/quickNavigator.test.ts --pool=forks --maxWorkers=1
- pnpm.cmd typecheck
- pnpm.cmd lint
- pnpm.cmd build

Fixes #591


debba (changes requested):
Hey @DhruvShah-Dev, I reviewed this one locally too. The fix itself is exactly right, and honestly this is the better of your two takes on #591: you memoized `isMultiDb` and used it consistently in the click handlers as well, so table refs get quoted the same way the rest of the app decides multi-db layout. All 9 tests pass on your branch, including the new regression test for stale database data, which is a nice touch.

The problem is timing: the command palette refactor merged on Aug 10 (778b5ccb, ff516445) deleted `QuickNavigatorModal.tsx` entirely, so this branch is now conflicting and the main file it patches no longer exists.

The good news is that the bug is still alive on current main and your fix ports over cleanly:

- The logic moved to `src/hooks/useCommandPaletteObjectItems.ts`, which still does `getDatabaseList(connection?.params.database)` (around line 95). On current main, `ConnectionData` in the provider already carries `selectedDatabases` per connection, so you can read `connectionData?.selectedDatabases` right there. It's actually a nicer fit than the old modal, because the hook is per-connection.
- While you're at it, `isMultiDatabaseCapable(connectionData?.capabilities)` in that hook should become `usesMultiDatabaseLayout(capabilities, selectedDatabases)` to match the sidebar and the editor.
- Your `quickNavigator.ts` param rename (`configuredDatabases` to `selectedDatabases`) and the regression test still apply almost as-is. The util was reshaped into a groups/flatMap structure on main, but the semantics are unchanged, so it's just a mechanical adjustment.

One more thing: #613 includes a subset of this same fix, so both PRs claim #591. I'd consolidate on this one since it's the more complete version, and drop the navigator changes from #613 (see my review over there).

So: rework on top of current main targeting `useCommandPaletteObjectItems.ts`, keep the test, and this should go in quickly. Thanks for the patience with the moving target, the palette refactor landed right under your feet.

debba (approved):
Thanks for porting the fix to the command palette flow. The implementation now uses the live per-connection selected database list, aligns multi-database detection with the rest of the app, and includes regression coverage ensuring stale cached database data is ignored. Verified locally: 7/7 targeted tests pass, typecheck passes, and lint passes. Looks good to merge.

PR #603, opened 2026-08-05
Editors whose tab was not active on first render never landed in edit…

GabrielMalava:
…orsRef, so Run fell back to executing the entire script instead of the statement under the cursor.


debba:
I reviewed this locally and the fix looks correct.

The root cause checks out: `@monaco-editor/react` keeps the `onMount` callback provided when the editor first mounts, so editors belonging to initially inactive tabs never get registered in `editorsRef`. Passing `onMount` to every editor fixes that, and the active-tab guard correctly prevents background editors from stealing focus. The safer fallback for multi-statement scripts is also a good improvement.

Local verification passed:

- TypeScript typecheck
- ESLint for `src/pages/Editor.tsx`
- Full Vitest suite: 220 files and 3,707 tests passed

My only non-blocking note is that the PR does not add a regression test for the exact flow it fixes: a tab that is inactive on initial render, then becomes active, followed by Run executing the statement under the cursor. Adding coverage for that interaction would make the fix safer against future regressions.

Looks good to merge from my side.

PR #606, opened 2026-08-06
feat(export): per-column anonymization in the export dialog

### LINKED ISSUE #483: [Feat]: Export data with column anonymization

iamthenuggetman:
Closes #483

## What

Adds the optional anonymization step to the file export proposed in the issue, as a transform layer between the row stream and the CSV/JSON/Markdown sinks — so it works for every driver (built-in **and** plugin) with no driver changes.

Export menu → **Anonymized export…** opens a dialog where each column of the result set can be assigned a rule:

- **NULL / fixed value** — replace with a real `NULL` or a literal (`***`)
- **Partial mask** — keep first/last N chars, mask the middle. Email-shaped values keep their shape: `john@example.com` → `j***@***.com`
- **HMAC pseudonym (pseudonymization)** — HMAC-SHA256 with a per-export key: the same input always produces the same output, so joins on pseudonymized keys (e.g. `user_id`) still work across tables exported with the same key. The key is shown in the dialog — reuse it across exports for stable pseudonyms, or regenerate it for a fresh set.

`NULL` values always pass through untouched — anonymizing them would invent data the source never had.

Per the issue's wording note, the feature is presented as *column anonymization / pseudonymization*, with an explicit hint in the dialog that it is not a "GDPR compliance" guarantee.

## Implementation

- `src-tauri/src/export/anonymize/` — rule types, `RowAnonymizer` (transform applied to streamed rows before the sinks), unit tests
- `export.rs` — `export_query_to_file` accepts an optional `anonymize` spec; `stream_to_sink` applies it. `hmac` added as a direct dep (already in the tree transitively)
- `ExportAnonymizeModal.tsx` — the rule-assignment dialog (follows the modal styling rules; Escape handling, per-export key field with regenerate)
- `Editor.tsx` — new "Anonymized export…" item in the export dropdown; the spec flows through `handleExportCommon`
- i18n strings for all 11 locales

## Tests

- 14 Rust unit tests: fixed/NULL replacement, null passthrough, unlisted columns, HMAC determinism + key/input sensitivity, non-string value handling, partial-mask edges (short strings, multibyte, email shapes), serde payload match with the frontend
- 2 frontend tests for the key generator
- Full suites green: 3530 vitest, 46 cargo export tests, `tsc -b` / eslint / clippy clean
- Manually tested against the demo databases: CSV export with partial-masked email, fixed-value name, and HMAC `user_id`; re-export with the same key produces identical pseudonyms, regenerating the key produces fresh ones

## Out of scope (possible follow-ups from the issue)

Synthetic data (seeded faker), generalization (birth date → year, zip → region), and name-based heuristics to pre-suggest rules — the rule engine and dialog are structured so these slot in as additional rule types later.


iamthenuggetman:
Closing for now — holding this one back for a later round. Branch stays intact; will reopen when ready.

PR #612, opened 2026-08-07
feat(datagrid): paste cells from clipboard (Cmd/Ctrl+V)

### LINKED ISSUE #611: [Feat]: Paste cells from clipboard into the data grid
### Is your feature request related to a problem?

Cells, rows, and ranges can be copied from the data grid (Cmd/Ctrl+C, context menu), but there is no way to paste data back into the grid. Filling several cells means editing them one by one.

### Describe the solution you'd like

Spreadsheet-style paste (DBeaver/Excel-like), staged through the existing pending-changes flow (commit/rollback):

- Cmd/Ctrl+V pastes at the current selection — cell range top-left, top of the row selection, or the focused cell — plus a context-menu "Paste" entry.
- TSV from spreadsheet apps and the grid's own CSV copy formats (comma/semicolon/pipe, optional header row) should round-trip.
- A single copied value fills the selected range.
- Data is clipped at the grid edges; nothing is written directly to the database — pasted values are staged and go through the normal apply/rollback flow.

### Additional context

I have an implementation ready and will open a PR shortly.

ymadd:
Closes #611

## Summary

The data grid supports copying cells/rows/ranges but not pasting. This PR adds spreadsheet-style paste as **staged edits** — pasted values go through the existing pending-changes flow (apply/rollback), never directly to the database.

## Behavior

- **Cmd/Ctrl+V** pastes at the selection, in priority order: cell range top-left → right-clicked cell (context-menu "Paste") → top of row selection → focused cell.
- **Parsing**: tab-separated cells win (spreadsheet convention). Multi-line text without tabs is parsed as CSV with double-quote escaping, preferring the configured CSV delimiter — so the grid's own copy formats (comma/semicolon/pipe) round-trip. A single line without tabs is always one value, so free text like `hello, world` lands in one cell.
- A leading header row is dropped only when it matches the grid's column names **positionally from the paste anchor** (round-trips the "export column names" option without swallowing external data that merely mentions column names).
- A single copied value **fills the whole selected range / selected rows**.
- The matrix is clipped at the grid edges. Existing rows require an identifiable row key (keyless grids: only pending-insertion rows accept a paste). Aliases and computed result columns are skipped — same guard as inline editing.
- Pasting a cell's original value back clears its pending change, same as inline editing.

## Prerequisite fix (first commit)

`handlePendingChange` / `handlePendingInsertionChange` computed the next pending state from a `tabsRef` snapshot; with N staged cells in one tick, React batches the updates and only the last cell survived. `updateTab` now also accepts an updater function and both handlers compute inside it. Multi-cell paste is the first caller that hits this, but it hardens every rapid-succession staging path.

## Known limitations

- The grid's CSV copy output does not quote embedded delimiters/newlines (`rowToCSV`), so values containing the delimiter don't survive a copy→paste round-trip. Robust CSV serialization is follow-up material.
- Cells containing quoted newlines (Excel multi-line cells) are split on the newline before quote handling.
- A single-line, no-tab copy of one row (headers off) pastes as one value by design — single-line splitting can't be distinguished from free text with commas.
- `NULL` copies as the literal string `null`, so pasting it into another cell stages that string, not SQL NULL.
- Rows are not auto-created on overflow: pasting more rows than exist below the anchor clips at the last row (add pending-insertion rows first).

## Testing

- `tests/utils/dataGrid.test.ts`: 23 new cases for `parsePasteMatrix` / `stripHeaderRow` / `computePasteTargets` (TSV/CSV/quotes/delimiter detection/positional header stripping/range fill/clipping).
- Manually verified against a MySQL database: range/row/single-cell paste, single-value fill, header round-trip, alias-column skip, keyless fallback, toast counts.
- `tsc --noEmit`, `eslint`, and the vitest suite are green (the 3 failing files on `main` — ThemeProvider/SettingsProvider/useSidebarResize — fail identically without this change).

## i18n

3 new keys (`pasteCells`, `pastedCells`, `pasteNotEditable`) added to all 11 locales.


debba:
Hey @ymadd, thanks for this one, really solid work.

I went through it locally today: checked out the branch, ran the full suite (3552 tests green, including your 23 new paste cases), tsc and eslint clean, and tested the paste flows by hand against a SQLite db (range paste, header round-trip, keyless table, alias columns, quoted CSV, and the no-op detection when you paste the original value back on its own cell). Everything behaved exactly as described in the PR. The functional update fix on `updateTab` is a nice catch too, the stale snapshot problem with multi-cell staging is real and this hardens it properly.

A few things before merging:

1. The branch is quite behind main by now and there's a conflict in `DataGrid.tsx`, so it needs a rebase. This is the only real blocker.

2. Paste skips the blob/JSON routing that inline editing does. Inline editing opens the sidebar for blob cells and the viewer for JSON, while paste stages a raw string into any physical column, blobs included, and you only find out at apply time. Fine for a first version, but I'd like a follow-up that skips blob columns the same way aliases are skipped.

3. The "single value + row selection" fill writes into every column, primary key included. Pasting `42` with 3 rows selected stages a PK change on all of them. It's coherent with spreadsheet behavior but it's the sharpest edge of the PR. I would exclude PK columns from the row-fill case, or at least call it out somewhere.

4. Minor: a clipboard read failure only goes to console.error. A toast would be more consistent with the rest of the grid.

5. Also minor: right-click paste ignores the clicked cell when a cell range is active (the range top-left wins). It's documented in the code comment so I'm fine with it, just noting it for the record.

Rebase it and I'm happy to approve. Points 2 and 3 can be follow-ups.

ymadd:
Thanks for the thorough review @debba — much appreciated, especially the hands-on testing!

All addressed and pushed:

- **(1) Rebased onto main.** The `DataGrid.tsx` conflict was main's new generated-columns guard — while resolving it I also made paste skip generated columns and un-revealed masked cells, matching the inline-editing guards that landed on main since this branch was cut.
- **(3) Agreed, fixed.** Primary key columns are now excluded from the single-value row-selection fill. The exclusion derives from column metadata's `is_pk` (not the row-identity fallback), so keyless tables still fill normally. Explicitly selected cells (range / focused cell) and multi-cell matrix pastes still accept PK values.
- **(4) Done.** Clipboard read failures now surface an error toast (`dataGrid.pasteReadFailed`, all 11 locales).
- **(2)** Will file a follow-up issue for skipping BLOB columns once this merges, and pick it up myself.
- **(5)** 👍 Leaving as documented.

Re-verified after the rebase: full suite green (the 3 suites that fail to load do so on main as well), `tsc`/`eslint` clean, and re-tested the paste flows by hand against a MySQL table including the new PK-exclusion and keyless fill paths.

debba:
@ymadd Do you think it would make sense to support pasting as new rows as well?

One possible approach would be to route the same Paste command based on the clipboard shape: partial matrices would keep updating existing cells, while full-width rows could become staged insertions. Alternatively, a separate explicit **“Paste as new rows”** action would avoid ambiguous cases.

Would you prefer to explore this within the current implementation or in a follow-up PR? If a follow-up sounds better, I’m happy to merge this PR as it is in the meantime.

ymadd:
Great idea — I'd take that on, but as a follow-up PR with an explicit **"Paste as new rows"** action in the context menu, rather than routing Cmd/Ctrl+V by clipboard shape. Shape alone isn't a reliable signal of intent: a full-width matrix can still be meant to update existing rows, while a perfectly valid insertion may omit generated, auto-increment, or defaulted columns. Keeping Cmd/Ctrl+V as an update operation keeps it predictable, and the explicit action gives the append case a clear home — still staged through the pending-changes flow.

It also crosses the current DataGrid/Editor boundary: the grid can edit existing pending-insertion rows and trigger the creation of a single one via `onDuplicateRow`, but creating N insertion rows from a pasted matrix needs a new bulk callback owned by the editor state — easier to review on its own.

So yes please — happy to have this merged as is. I'll file the follow-up issues right after (paste-as-new-rows, plus the BLOB-column guard you flagged) and pick both up.

debba:
Merging it right now .

PR #613, opened 2026-08-08
Fix database UI query issues

### LINKED ISSUE #591: Object palette reads the database list from connection params instead of the live selection

### LINKED ISSUE #608: [Bug]: Query error text cannot be selected or copied
### Describe the bug

Image

When a SQL query fails, the error message displayed in the result panel cannot be selected with the mouse or copied using Ctrl+C.

Expected behavior: Error text should be selectable and copyable, ideally with a “Copy error” button.

OS: Windows
Database: MySQL

### To Reproduce

1. Connect to a MySQL database.
2. Open a SQL query editor.
3. Run an invalid query that returns a database error, for example:

   SELECT * FROM orders WHERE payment_status = 'paid';

   when the `payment_status` column does not exist.

4. The error message is displayed in the result/error panel.
5. Try to select the error message using the mouse.
6. Try to copy it using Ctrl+C.
7. The error text cannot be selected or copied.

### OS Version

Windows 10

### Tabularis Version

v0.18.0

### Relevant Log Output

```shell

```

### LINKED ISSUE #610: [Bug]: Detect JSON in columns bugs out updating json in text cells in mariadb.
### Describe the bug

When enabling the option `Detect JSON in text columns` on a MySql connection to a MariaDB database (version 10.3.31), jsons in text fields show correctly, but when trying to update them, we get this error:

`Update failed: error returned from database: 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'JSON) WHERE `id` = ?' at line 1`

Disabling that option fixes the issue.

### To Reproduce

Have a text column with jsons in a MariaDB database with version 10.3.31. Then try to edit the field.

### OS Version

ArchLinux (flatpak version)

### Tabularis Version

18.0.0

### Relevant Log Output

```shell

```

DhruvShah-Dev:
## Summary
- Use the live selected database list in the object palette and align its multi-database branch with the sidebar.
- Make query error text selectable and add an explicit copy action.
- Render MySQL/MariaDB JSON update values as JSON text instead of CAST(... AS JSON), which MariaDB rejects.

Closes #591.
Closes #608.
Closes #610.

## Validation
- pnpm.cmd vitest run tests/components/modals/ErrorModal.test.tsx tests/utils/quickNavigator.test.ts
- pnpm.cmd typecheck
- pnpm.cmd lint
- pnpm.cmd build
- git diff --check
- cargo test mysql_json_update_value --manifest-path src-tauri/Cargo.toml --lib --no-run
- GitHub CI test job passed


debba (changes requested):
Hey @DhruvShah-Dev, thanks for this one! I did a deep local review: ran your tests on the branch and spun up real MariaDB 11 and MySQL 8.4 containers to verify the JSON fix. Here's where I landed.

**The good news: the MariaDB JSON fix is solid.** Verified against real databases:

- the old shape `SET col = CAST('{"ok":true}' AS JSON)` fails on MariaDB 11 with error 1064, which is exactly the bug in #610
- the new plain-string shape works on MariaDB for both TEXT and JSON columns, and on MySQL 8.4 the value stays typed JSON (`JSON_TYPE` = OBJECT)
- MySQL still rejects invalid JSON server-side (error 3140), so dropping the CAST doesn't lose any validation

Both new Rust tests pass on your branch, and extracting `push_mysql_update_value` matches the existing `push_pk_condition` pattern in `mod.rs`. This part is merge-ready as far as I'm concerned.

**The problem: the branch is now conflicting with main.** The command palette refactors merged on Aug 10 (778b5ccb, ff516445) deleted `QuickNavigatorModal.tsx` entirely. The bug you're fixing in #591 is still real on current main, but it now lives in `src/hooks/useCommandPaletteObjectItems.ts` (around line 95), which still derives `configuredDatabases` from `getDatabaseList(connection?.params.database)` instead of the live selection. So that part needs to be reimplemented in the new hook rather than mechanically rebased. Heads up: the hook is per-connection via `connectionDataMap`, while `selectedDatabases` in `DatabaseProvider` tracks the active connection only, so it's not a one-to-one port.

**About #608:** the issue is about the error text in the query result panel, which is rendered by `src/components/ui/ErrorDisplay.tsx`. That component is untouched here and is still unselectable (we have a global `user-select: none` in `index.css`). `ErrorModal.tsx` is actually dead code right now: its only mount in `Editor.tsx` has an `isOpen` state that nothing ever sets to true, on your branch as well as on main. Also note we already shipped copy + selectable text in `AlertModal` via #618. So the right home for the #608 fix is `ErrorDisplay.tsx`.

Minor nit: the copy button uses `dataGrid.copied`, but since #618 we have `common.copied` in every locale, better to use that one for consistency.

My suggestion: split this PR in three.

1. The Rust JSON fix + tests as a standalone PR, that one can go in quickly.
2. The #591 fix reimplemented in `useCommandPaletteObjectItems.ts` on top of current main.
3. The #608 fix moved to `ErrorDisplay.tsx` (and `ErrorModal.tsx` could probably just be deleted as dead code while you're at it).

Thanks again for digging into these!

PR #624, opened 2026-08-11
Hide SQLite virtual table internals

### LINKED ISSUE #622: [Bug]: hidden columns of virtual tables show up as editable fields in the New Row modal
### Describe the bug

Follow-up to #581. The SQLite driver now reads column metadata with `PRAGMA table_xinfo` instead of `table_info`, which was needed to expose generated columns. The side effect is that `table_xinfo` also returns the hidden columns of virtual tables (`hidden = 1`), which `table_info` used to leave out.

Only `hidden` values 2 and 3 are flagged as `is_generated`, so these hidden columns are treated as plain writable columns. The data grid is unaffected because `SELECT *` never returns them, but the New Row modal builds its form from the column metadata, so on an fts5 table it now lists the internal `docs` and `rank` columns as fillable fields.

This is not just cosmetic: in fts5 the column named after the table is the command interface, so a value written there can trigger special commands (e.g. `'delete-all'` wipes the whole index).

The fix should be a one-liner: skip rows with `hidden = 1` when mapping `table_xinfo` results in `sqlite_column_from_row` / `get_columns` / `get_all_columns_batch` (`src-tauri/src/drivers/sqlite/mod.rs`). That restores the old `table_info` behavior for virtual tables while keeping generated columns visible.

### To Reproduce

1. Create a SQLite db with a virtual table: `CREATE VIRTUAL TABLE docs USING fts5(title, body);`
2. Open it in Tabularis and select the `docs` table
3. Open the New Row modal
4. `docs` and `rank` are listed as editable fields, only `title` and `body` should be there

### OS Version

Arch Linux

### Tabularis Version

v0.19.0 (main, after #581)

### Relevant Log Output

```shell

```

DhruvShah-Dev:
## Summary
- Filters SQLite `table_xinfo` rows with `hidden = 1` out of column metadata.
- Keeps generated columns visible by continuing to expose `hidden = 2` and `hidden = 3` as generated metadata.
- Adds FTS5 coverage for direct and batch SQLite column metadata lookups.

## Root Cause
`PRAGMA table_xinfo` includes hidden virtual-table columns that `table_info` did not return. Those rows were mapped as normal writable columns and surfaced in the New Row modal.

Fixes #622

## Validation
- `git diff --check`
- `cargo test test_get_columns_excludes_hidden_virtual_table_columns --lib -- --nocapture` timed out locally after 3 minutes during Rust execution/build.


debba (approved):
Reviewed this locally, checked out the branch and tested against a real db. LGTM.

**What I checked**

- The `hidden = 1` filter is applied in `sqlite_column_from_row` and all three call sites (`get_columns`, `get_all_columns_batch`, `get_view_columns`) now use `filter_map`. Grepped the branch, there are no other places mapping `table_xinfo` rows, so nothing was missed.
- Confirmed the premise against a real database. On an fts5 table, `PRAGMA table_xinfo` returns the internal `docs` and `rank` columns with `hidden = 1`, while generated columns report `2` (virtual) and `3` (stored). So the filter removes exactly the fts5 internals and keeps generated columns visible with `is_generated: true`, same as before.
- This is more than cosmetic btw: in fts5 the column named after the table is a command interface, so having it in the New Row modal meant a user could accidentally run commands like `'delete-all'`. Good catch on #622.

**Tests**

You mentioned the test timed out on your machine. It runs fine here on the branch:

```
cargo test --lib drivers::sqlite::tests
12 passed; 0 failed (0.02s after build)
```

That includes the new `test_get_columns_excludes_hidden_virtual_table_columns`, which covers both the direct and the batch lookup against a real fts5 table.

**Manual testing**

Seeded a db with an fts5 table, a table with virtual + stored generated columns, a plain table and a view:

- fts5 New Row modal: only `title` and `body`, no `docs` or `rank` anymore
- generated columns: still shown and still flagged as generated, not writable
- plain table and view: unchanged

**Notes (non blocking)**

- `PRAGMA table_xinfo('{}')` with the raw table name predates this PR, but a table name containing `'` would break the query. Worth a separate issue.

PR #629, opened 2026-08-13
feat: support plugin-owned connection fields and secrets

debba:
## Summary

- add optional per-plugin presentation overrides for common connection fields
- add plugin-owned connection secrets backed by the OS keychain
- expose only secret metadata and tri-state changes to UI extensions
- hydrate secrets only for runtime driver calls and keep them out of persisted connection parameters
- preserve secret lifecycle across save, update, test, duplicate, delete, import, export, and MCP flows
- document the additive manifest and UI extension contracts

## Compatibility

All new manifest, model, request, and UI context fields are optional or defaulted. Existing manifests, saved connections, UI extensions, and Password-based plugins keep their current behavior.

Stored plugin secrets are never returned to the frontend as plaintext. Existing values are represented only by `hasStoredValue`, and an unchanged field remains untouched.

## Tabularium prerequisite

Tabularium must be updated to accept `connection_fields` in the driver manifest schema before plugins using this property can be published. The current schema served by `registry.tabularis.dev` rejects it as an additional property.

This PR updates the host-side schema and documentation, but it does not update the operational Tabularium registry schema. That registry change must be deployed before publishing the dependent BigQuery plugin PR.

Dependent plugin PR: https://github.com/TabularisDB/tabularis-bigquery-plugin/pull/2

## Validation

- `cargo test --all-targets`: 1137 passed, 4 ignored; integration suite 9 ignored
- `cargo test --all-targets --no-run`
- `pnpm exec vitest run`: 220 files, 3715 tests passed
- `pnpm build`
- `pnpm lint`
- `pnpm exec tsc --noEmit`
- plugin API sync and build
- create-plugin build
- `git diff --check`

`cargo clippy --all-targets -- -D warnings` still reports existing workspace-wide lint debt unrelated to this change. The new code does not add the MSRV incompatibility found during review.

### DIFF
--- a/packages/create-plugin/src/cli.ts
+++ b/packages/create-plugin/src/cli.ts
@@ -7,7 +7,7 @@ import { scaffold } from "./scaffold";
 import { titleCase, validateDbType, validateName, validateQuote } from "./validate";
 
 const PACKAGE_VERSION = "0.1.0";
-const PLUGIN_API_VERSION = "0.1.0";
+const PLUGIN_API_VERSION = "0.2.0";
 const MIN_TABULARIS_VERSION = "0.9.20";
 
 function main(argv: string[]): number {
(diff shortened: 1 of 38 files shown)


debba:
@NewtTheWolf please can you check the tabularium part?

NewtTheWolf:
I also checked the live Tabularium registry schema: `connection_fields` is still not present in the `driver` kind schema there. Please add it on the Tabularium side too, otherwise plugins using this field still cannot be published even though the Tabularis host side is ready.

PR #635, opened 2026-08-14
Fix: Keep "Manage databases" and "Refresh" controls visible when a single database is selected

gcapellib:
Currently, when a single database is selected, the UI switches to a specific view that hides the "Manage databases" and "Refresh" buttons. Users are forced to return to the root connection configuration to select other databases, which degrades the UX.

Solution
This PR unifies the UI behavior by keeping the multi-database layout active even when only one database is selected, while ensuring immediate access to its tables.

Changes in DatabaseProvider.tsx:

Keep controls visible: Adjusted thresholds from > 1 to >= 1 for savedDbList.length and selection.length. The view now displays DATABASES (1) with the management buttons always accessible.

Auto-expand: To compensate for replacing the old single-view layout, the single database is automatically set as activeSchema (firstDb). It expands by default to directly show its tables (saving an extra click).


debba:
Hi @gcapellib ,
can you fix the merge conflicts?

PR #643, opened 2026-08-14
fix: run the SSL-mode migration from the --mcp server process (#639)

### LINKED ISSUE #639: SSL-mode migration (#614) never runs for the --mcp server process
## Summary

The stale `ssl_mode`-spelling migration added in [#614](https://github.com/TabularisDB/tabularis/issues/614) (`migrate_postgres_ssl_mode_spelling`, `src-tauri/src/commands.rs`) only runs from the two GUI-process Tauri commands that load connections — `get_connections` and `get_connections_with_groups`. It is never invoked from the `--mcp` server process.

## Why this matters

The SSL-mode dropdown bug that #614 fixed could, before the fix, save a postgres-dialect (plugin-driven) connection with a MySQL-style `ssl_mode` value (`"required"`, `"verify_ca"`, etc.) that the plugin's `needs_tls()` doesn't recognize — silently connecting in cleartext. The migration exists specifically to rewrite those already-persisted values going forward.

`tabularis --mcp` ([documented, user-facing](https://tabularis.dev/wiki/mcp-server), not experimental) reads connections from the exact same `connections.json` (`paths::resolve_connections_path`, confirmed identical path resolution to the GUI process) and uses the `ssl_mode` value directly to establish real database connections via `resolve_db_params` → the driver's `execute_query`/etc. `register_drivers_for_mcp()` (`src-tauri/src/mcp/mod.rs`) never calls any connection migration — not the new SSL one, and not the pre-existing `migrate_ssh_connections` either.

**Concrete exposure:** a connection saved with a stale `ssl_mode` value before this PR's dropdown fix, if never reopened in the GUI (so `get_connections` never runs against it), connects in cleartext indefinitely when accessed via `tabularis --mcp` — e.g. an AI agent using MCP to query that connection. The GUI-side fix provides no protection for this path.

## Root cause / why it wasn't just fixed inline

`migrate_postgres_ssl_mode_spelling` takes an `AppHandle<R>` and is called from Tauri command context. `register_drivers_for_mcp()` has no `AppHandle` — it resolves paths directly via `paths::get_app_config_dir()` / `config::load_config_from_disk()`. Wiring the migration into the MCP startup path requires extracting a path-based core (no `AppHandle` dependency) that both the two existing Tauri-command call sites and `register_drivers_for_mcp()` can call — not a one-line fix.

## Scope note

This is **not a regression introduced by #614** — the MCP process has never run *any* connection migration, including the older pre-existing SSH migration (`migrate_ssh_connections`). It's a pre-existing gap in how the MCP process bootstraps that #614's fix doesn't happen to close, surfaced during a final safety-review pass on [PR #577](https://github.com/TabularisDB/tabularis/pull/577) before requesting re-review.

## Suggested fix

1. Extract a path-based (no `AppHandle`) core migration function that both `migrate_ssh_connections`/`migrate_postgres_ssl_mode_spelling` (Tauri-command side) and `register_drivers_for_mcp` (MCP side) can call.
2. Call it from `register_drivers_for_mcp()` before the MCP server starts handling requests.
3. Consider whether this same gap applies to any other non-GUI entry point that reads `connections.json` directly (worth a broader audit, not just this one migration).

aesslinger:
Closes #639

## Summary

The stale `ssl_mode`-spelling migration added in #614 only ran from the two GUI-process Tauri commands that load connections (`get_connections` / `get_connections_with_groups`). It was never invoked from the standalone `tabularis --mcp` server process, which reads the same `connections.json` and uses `ssl_mode` directly to establish real database connections. A connection saved with a stale MySQL-style `ssl_mode` value before #614's dropdown fix shipped stayed silently cleartext forever via MCP, even after the GUI-side fix landed, as long as it was never reopened in the GUI.

## What changed

Four commits, each independently buildable/testable:

1. **`refactor: extract SSL-mode migration into a path-based shared module`** — moves the migration logic out of `commands.rs` into a new `connection_migrations.rs` module. The orchestration function is split into a path-based core (`migrate_postgres_ssl_mode_spelling_at_path`, no `AppHandle` dependency) plus a thin `AppHandle` wrapper left in `commands.rs` for the GUI's cache-invalidation need. Pure extraction, no behavior change — except the return type improves from `Result<()>` to `Result<bool>` so callers can tell "nothing needed migrating" apart from "migrated and saved."

2. **`fix: guard the SSL-mode migration against a concurrent-write race`** — this fix makes MCP a writer of `connections.json` for the first time (previously read-only there). Since MCP clients spawn `tabularis --mcp` as an independent subprocess that can run alongside the (single-instance) GUI app, and this codebase's connection persistence layer has no file locking, adds a content-based compare-and-swap: the file's raw content is checked immediately before writing, and the migration skips itself for this run (rather than overwriting) if it changed since the initial read. Safe because the migration is idempotent and self-terminating — the next process to load connections will retry.

3. **`fix: run the SSL-mode migration from the --mcp server process`** — wires the shared core into `register_drivers_for_mcp()`, appended after plugin loading completes (same ordering constraint the GUI path already has, since dialect resolution depends on the driver registry). Errors are swallowed the same way the existing GUI call site already does — migration failure must never block server startup.

4. **`refactor: avoid a redundant file read in the SSL-mode migration guard`** — found during local review: the guard read `connections.json` twice before comparing (once directly, once again inside `persistence::load_connections_file`). Not a correctness bug — traced every ordering and a concurrent edit landing in the gap still gets caught by the final comparison — but an unneeded extra syscall and a slightly wider window than necessary. Splits `load_connections_file`'s parsing logic into a new `persistence::parse_connections_file(&str)` so the migration can parse the content it already read, without touching any of that function's ~20 other call sites.

## Explicitly out of scope

- **`migrate_ssh_connections`** has this same gap (never called from MCP either) but is meaningfully more complex — a separate `ssh_connections.json` file plus OS keychain credential migration. Left as a follow-up decision, not fixed here.
- **Real file locking** across the whole persistence layer — the correct long-term fix for the concurrency risk, but would touch ~20 existing write call sites, not just this one. The compare-and-swap guard added here only protects this specific new write path.

Full design rationale, alternatives considered, and why each was rejected are documented on #639 before implementation began.

## Verification

- `cargo build --all-targets` — 0 warnings
- `cargo clippy --all-targets` — 0 findings in touched files (129 pre-existing elsewhere, confirmed present on a clean `main` checkout)
- `cargo fmt --check` — clean on all touched files
- `cargo test --lib` — 1138/1138 (4 pre-existing environment-only skips)
- `cargo test --test postgres_integration --include-ignored` — 181/181
- Real-plugin parity (`POSTGRES_PLUGIN_BIN` against a live `v1.0.0-beta.3` binary) — 83/83
- Manual end-to-end: built the actual `tabularis --mcp` binary and ran it against an isolated `$HOME` with the real installed plugin binary and a hand-written `connections.json` covering all 3 driver cases. Confirmed the exact expected outcome: the `postgresql`-driven connection's stale `required` rewrote to `require` (with the `[Migration] Rewriting stale ssl_mode spelling on 1 postgres-dialect connection(s)` log line), while `mysql` and builtin `postgres` connections were left untouched. Second run confirmed idempotent (no log line, no further changes).


aesslinger:
## Local verification notes

Ran a local review pass against `.rules/rust.md` and verified every claim in the PR description against the actual code (not just skimmed the diff):

**Rules compliance** — all satisfied: pure helpers extracted to a dedicated sibling module (`connection_migrations.rs`), tests in a sibling `_tests.rs` file loaded via `#[cfg(test)] pub mod`, all 3 pure functions have direct unit coverage, and the one type-signature change (`Result<()>` → `Result<bool>`) is behavior-preserving for both existing GUI call sites (traced through — cache invalidation happens under the exact same condition before and after).

**Blast radius** — full grep of every moved/new function name confirms nothing outside the touched files references them; no external callers missed.

**One nuance found and fixed**: `migrate_postgres_ssl_mode_spelling_at_path` was reading `connections.json` twice before comparing (once directly, once again inside `load_connections_file`). Traced every possible timing of a concurrent edit — not a correctness bug, since the final `content_before` vs `content_now` comparison still catches it — but an unnecessary extra read. Fixed in the 4th commit by splitting `load_connections_file`'s parsing into a reusable `parse_connections_file(&str)`, with zero signature change for its ~20 other call sites.

**Manual end-to-end verification**: built the real `tabularis --mcp` binary and ran it against an isolated `$HOME` with the actual installed PostgreSQL plugin binary and a hand-written `connections.json` covering all 3 driver cases (plugin with a stale value, MySQL with the same-but-correct value, builtin Postgres). Confirmed via the process's own log output and a file diff:

| Connection | Before | After | Result |
|---|---|---|---|
| `postgresql` (plugin) | `required` | `require` | ✅ rewritten |
| `mysql` | `required` | `required` | ✅ correctly untouched |
| `postgres` (builtin) | `required` | `required` | ✅ correctly untouched (by design) |

Second run produced no `[Migration]` log line and no further changes — confirmed idempotent.

No blockers found. All test/build/lint claims in the PR description were independently reproduced, not taken on faith.

debba:
Reviewed this locally on the branch, ran the tests and reproduced the manual E2E with real data.

What I checked:

- The extraction is a genuine pure move. I diffed the old body in `commands.rs` against the new module and the only behavioral changes are the documented ones (the `bool` return and the CAS guard). The GUI wrapper invalidates the cache only when something was actually written, which also means a CAS skip doesn't invalidate. Correct.
- Path consistency: the MCP call site uses the same `resolve_connections_path(&get_app_config_dir())` expression the MCP module already uses for all its reads, and `get_config_path` on the GUI side funnels into the same helper, so both processes hit the same file.
- Ordering and protocol hygiene: the migration is awaited inside `register_drivers_for_mcp()` before the server starts serving, and it logs to stderr, keeping the stdout JSON-RPC stream clean.
- `load_connections_file` keeps its exact signature, so its ~30 callers are untouched.

Tests on the branch: `cargo test --lib connection_migrations` 12/12, full `cargo test --lib` 1151 passed, 0 failed (4 env-only skips).

For the E2E I built the binary and staged an isolated `$HOME` with the postgresql plugin actually installed (release build plus a `.tabularium` declaring `sql_dialect: "postgres"`) and a 4-case `connections.json`. First run logged `Rewriting stale ssl_mode spelling on 2 postgres-dialect connection(s)`: `required` became `require` and `verify_ca` became `verify-ca` on the plugin connections, while mysql and builtin postgres were left untouched. Second run: no log line, file byte-identical. Idempotency confirmed.

A few non-blocking notes:

1. `connections_file_changed_concurrently` is a bare `!=` behind a function. Fine since it carries the mtime rationale and its own test, just noting it.
2. The residual TOCTOU between the second read and the write is real but acknowledged, and the idempotent-retry argument holds. Acceptable given the file-locking follow-up.
3. The serde round-trip normalizes the file on write and would drop fields unknown to the writing binary. Same behavior the GUI path always had, but MCP clients can pin an older binary than the installed GUI, so it gets slightly more likely now that MCP writes. Worth a line in the follow-up issue.
4. Leaving `migrate_ssh_connections` out is the right call, the keychain part makes it a different-shaped change.

LGTM.

aesslinger:
Thank you for the thorough review — really appreciate the depth here, especially reproducing the manual E2E independently with your own isolated `$HOME`/plugin setup rather than just taking the description's claims at face value.

On note 3 (serde round-trip dropping unknown fields, more likely to matter now that MCP writes too): filed as a follow-up, #668, along with the related file-locking gap we'd already flagged and deferred during design. Not blocking this PR — just tracked for later.

debba:
Great, thanks @aesslinger .
Merging it.

PR #644, opened 2026-08-14
ci: cache Rust dependencies in the test job (#640)

### LINKED ISSUE #640: ci.yml's test job has no Rust build caching — ~6min of every run is redundant cargo compile
## Problem

`ci.yml`'s `test` job takes a median of **~10.3 minutes** per run, and roughly **80% of that** (median 487s / ~8.1 min) is spent in a single step: `Test` (`pnpm run test:all`, which runs `pnpm test:rust && pnpm test:coverage`). Within that step, the Rust half (`pnpm test:rust` → `cd src-tauri && cargo test`) consistently takes **~6-6.5 minutes** — and almost all of that is **compiling the dependency graph from scratch**, not running the actual tests.

This job has **no caching at all** for the Rust build — no `actions/cache`, no `swatinem/rust-cache` — so every single push/PR run recompiles all ~9,500 lines of `Cargo.lock`'s dependency graph before running a single test.

## How this was verified

**Step-level timing across a month of history.** Pulled all 348 successful `ci.yml` runs from the last 30 days via the GitHub API, sampled every 10th run (35 runs, evenly spread across the month) for step-level timing:

| Step | median | mean | % of median total job |
|---|---|---|---|
| Install Rust/Tauri dependencies (apt) | 30s | 58s | ~5% |
| **Test** (`pnpm run test:all`) | **487s (~8.1 min)** | 478s | **~80%** |
| Total job | 616s (~10.3 min) | 627s | 100% |

(One run I'd initially flagged for a slow apt step turned out to be an outlier — 414s vs. the 30s median. Across the full sample, apt is not the bottleneck; `Test` consistently is.)

**Sub-step breakdown within `Test`, from raw job logs (3 representative runs, full timestamps):**

| Sub-phase | Run A | Run B | Run C |
|---|---|---|---|
| `pnpm test:rust` (cargo compile + link + run) | 6m27s | 6m10s | 6m27s |
| `pnpm test:coverage` (vitest, 222 files / 3726 tests) | 2m38s | 2m28s | 2m15s |

**Within `pnpm test:rust` itself** (one run's raw log, second-by-second):
- Compiling dependencies from scratch: **3m51s**
- Final linking / binary finalization: **2m16s**
- Actually running the 1130 backend unit tests: **19s**

So of the ~6.5 minutes spent on `pnpm test:rust`, only ~19 seconds is the tests themselves — the rest (~6m7s, ~95%) is cold compilation that happens identically on every run regardless of whether `src-tauri/` code even changed in that particular commit.

**Confirmed no existing caching mechanism is applied here**, and confirmed no existing issue/PR already covers this — searched issues/PRs by multiple query variants and the GitHub search API directly for `cache cargo`, `CI workflow slow`, `actions/cache`, `ci.yml`, `test:rust`; the closest matches (#599, #540) are unrelated (AppImage bundling fix, and GitHub Actions SHA-pinning respectively).

## Proposed fix

**This repo already has the answer to this in `build.yml`** — no new tooling needed:

```yaml
      - name: Cache Rust dependencies
        uses: swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
        with:
          workspaces: src-tauri
```

This exact step (same pinned SHA, already vetted via #540's action-pinning pass) is used in `build.yml`'s release-build job. Adding the equivalent step to `ci.yml`'s `test` job — before the `Test` step, after checkout — should let `cargo test` reuse cached build artifacts on unchanged dependencies, cutting the ~6 minutes of redundant recompilation to near-zero on a cache hit.

Suggested placement in `.github/workflows/ci.yml`, right after "Install Rust/Tauri dependencies" and before "Install dependencies" (pnpm), so the cache is warm before `pnpm run test:all` invokes cargo.

## Secondary, lower-priority note

`pnpm test:coverage` (vitest, ~2-2.5 min) is comparatively reasonable for 3726 tests across 222 files and isn't flagged here as needing changes — this issue is scoped to the Rust compile-caching gap specifically, since that's the dominant, clearly-fixable cost.

## Alternative path, if caching is rejected

If Rust build caching (via `swatinem/rust-cache` or `actions/cache`) isn't something the team wants to adopt for `ci.yml` — e.g. concerns about cache staleness/size on a shared runner, or a preference to keep this job's dependency footprint minimal — a smaller, independent win is available: **run `pnpm test:rust` and `pnpm test:coverage` as two parallel jobs instead of one sequential step.** GitHub Actions jobs run concurrently by default, so wall-clock for that phase would drop to roughly `max(6.5min, 2.5min) ≈ 6.5min` instead of the current `6.5 + 2.5 = 9min` — an ~2.5 minute reduction, at the cost of duplicating the shared setup (checkout, apt install, `pnpm install`) across two jobs instead of one (each ~30s-1min per the median above, so still a net win).

This alternative is explicitly **not additive with the caching fix above** — it's worth doing only if caching is off the table. Once/if caching lands and `pnpm test:rust` drops from ~6.5min to near the ~19s actual test-run time, `pnpm test:coverage` (~2.5min) becomes the dominant cost on its own, and splitting two steps where one takes 19s and the other 2.5min buys almost nothing. Treat these as alternative approaches to the same problem, not a combined roadmap.

aesslinger:
Closes #640

## Summary

`ci.yml`'s `test` job had no Rust build caching, so every push/PR run recompiled the entire `Cargo.lock` dependency graph from scratch before running a single test — ~6.5 of the job's ~10.3 median minutes, of which only ~19s was the actual test run. Full analysis (a month of CI history, step-by-step timing breakdown) is on #640.

## What changed

Adds the exact same `swatinem/rust-cache` step (same pinned SHA, same `workspaces: src-tauri` config) already in use in `build.yml`'s release job — no new tooling, just applying an already-vetted pattern to a job that was missing it.

@debba [confirmed this approach](https://github.com/TabularisDB/tabularis/issues/640#issuecomment-5294056851) on the issue before this PR was opened.

## Verification

- YAML syntax validated
- Diffed against `build.yml`'s existing usage to confirm identical action reference and configuration — no drift
- **Measured directly on this PR's own CI runs** — run 1 populated the cache (necessarily a miss, nothing to restore yet), run 2 hit it:

  | Step | Run 1 (cache miss) | Run 2 (cache hit) | Delta |
  |---|---|---|---|
  | Cache Rust dependencies | 2s | 34s (restoring ~1.14 GB) | — |
  | **Test** (`pnpm test:rust && pnpm test:coverage`) | **8m28s** | **3m37s** | **−4m51s** |
  | **Total job** | **~11m45s** | **~6m29s** | **−5m16s (~45% faster)** |

  Confirmed via the cache action's own log: `Cache hit for: v0-rust-test-Linux-x64-...`, `Cache restored successfully`, `full match: true`. The `Test` step's drop matches #640's own analysis almost exactly — that step's ~6.5 minutes of Rust time was ~19s of actual test execution and the rest cold compile, and cold compile is what the cache now skips on a hit.
- Run 1: https://github.com/TabularisDB/tabularis/actions/runs/31816668408
- Run 2 (cache hit): https://github.com/TabularisDB/tabularis/actions/runs/31817948418

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,11 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libwebkit2gtk-4.1-dev libsoup-3.0-dev build-essential curl wget file libssl-dev libgtk-3-dev libayatana-appindicator3-dev librsvg2-dev libfuse2
 
+      - name: Cache Rust dependencies
+        uses: swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
+        with:
+          workspaces: src-tauri
+
       - name: Install dependencies
         run: pnpm install
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
 
       - name: Lint
         run: pnpm run lint
-        
+
       - name: Build
         run: pnpm run build
 


aesslinger:
## Cache confirmed working — before/after from this PR's own CI runs

Ran two CI runs on this branch: run 1 necessarily populates the cache (nothing to restore yet), run 2 hits it.

| Step | Run 1 (cache miss) | Run 2 (cache hit) | Delta |
|---|---|---|---|
| Cache Rust dependencies | 2s | 34s (restoring ~1.14 GB) | — |
| **Test** (`pnpm test:rust && pnpm test:coverage`) | **8m28s** | **3m37s** | **−4m51s** |
| **Total job** | **~11m45s** | **~6m29s** | **−5m16s (~45% faster)** |

Confirmed via the cache action's own log output on run 2:
```
Cache hit for: v0-rust-test-Linux-x64-8505e8d8-1942e370
Cache Size: ~1139 MB (1194682537 B)
Cache restored successfully
Restored from cache key "v0-rust-test-Linux-x64-8505e8d8-1942e370" full match: true.
```

The `Test` step's drop lines up almost exactly with #640's own analysis — that step's ~6.5 minutes of Rust time was ~19s of actual test execution and the rest cold compile, and cold compile is what the cache skips on a hit.

Run 1 (miss): https://github.com/TabularisDB/tabularis/actions/runs/31816668408
Run 2 (hit): https://github.com/TabularisDB/tabularis/actions/runs/31817948418

PR #645, opened 2026-08-14
fix: getDriverIcon never checked for a URL/data: manifest icon (#632)

### LINKED ISSUE #632: External plugin icon URLs never render outside the database picker — getDriverIcon has no URL/data-URI branch
## Summary

An external plugin's `.tabularium` `icon` field (documented as accepting a hosted PNG/SVG URL or `data:` URI — see the [Tabularium manifest docs](https://registry.tabularis.dev/docs/plugin-development)) only renders correctly in **one** of the 8 places the app shows a driver/connection icon. Everywhere else, every external plugin — regardless of what its `icon` field is set to — silently falls back to the generic `Plug` icon.

Found while preparing `tabularis-postgresql-plugin` for its first Tabularium registry submission ([tabularis-postgresql-plugin#16](https://github.com/TabularisDB/tabularis-postgresql-plugin/pull/16) added `icon: "https://raw.githubusercontent.com/.../postgresql-icon.svg"`, confirmed working in the database picker, confirmed **not** working in the connection card).

## Root cause

Two independent icon-rendering code paths exist, and only one of them checks for a URL:

**`src/components/modals/connection/EngineCard.tsx`'s `renderIcon` (correct)** — used only by the "Choose a database" picker:
```tsx
function renderIcon(rep: CatalogueDriver) {
  const icon = rep.icon ?? "";
  if (/^https?:\/\//.test(icon) || icon.startsWith("data:")) {
    return <RegistryDriverIcon src={icon} size={24} fallback={<Database size={20} />} />;
  }
  if (rep.isBuiltin) {
    return getDriverIcon({ icon, color: rep.color ?? undefined } as PluginManifest, 22);
  }
  return <Database size={20} />;
}
```

**`src/utils/driverUI.tsx`'s `getDriverIcon` (missing the check)** — used by every other icon-rendering site:
```tsx
export function getDriverIcon(manifest: PluginManifest | undefined | null, size = 14): ReactNode {
  const iconName = manifest?.icon || "";

  // Brand icons for built-in drivers
  switch (iconName) {
    case "postgres": return <PostgreSQLIcon size={size} />;
    case "mysql": return <MySQLIcon size={size} />;
    case "sqlite": return <SQLiteIcon size={size} />;
  }

  // Legacy lucide icon names
  switch (iconName) {
    case "network": return <Network size={size} />;
    case "database": return <Database size={size} />;
    case "folder-open": return <FolderOpen size={size} />;
    default: return <Plug size={size} />;
  }
}
```

`getDriverIcon` never inspects the *value* of `manifest.icon` beyond comparing it against 6 hardcoded literal strings. Since no external plugin's `icon` can ever equal `"postgres"`/`"mysql"`/`"sqlite"` (those are reserved built-in lookup keys mapped to inline `<PostgreSQLIcon>`/`<MySQLIcon>`/`<SQLiteIcon>` React components baked into the app bundle) or one of the 3 legacy lucide names, **every external plugin's icon falls through to `default: return <Plug size={size} />`**, regardless of whether `icon` is a URL, a `data:` URI, or anything else. This is a format-independent gap — switching the manifest's `icon` value between a hosted URL, a base64 `data:` URI, or any other representation would not change the outcome, since the function never reaches a branch that inspects the string beyond exact-match against those 6 literals.

## Where this is visible

`getDriverIcon`/`getConnectionIcon` (which just adds per-connection override handling on top of `getDriverIcon`) are called from 8 sites; only `EngineCard.tsx` has its own separate, correct handling:

- `src/components/layout/sidebar/ConnectionGroupItem.tsx`
- `src/components/layout/sidebar/OpenConnectionItem.tsx`
- `src/components/modals/VisualExplainModal.tsx`
- `src/components/modals/NewConnectionModal.tsx`
- `src/components/connections/ConnectionListItem.tsx`
- `src/components/modals/NewConnectionModal/AppearanceSection.tsx`
- `src/components/connections/ConnectionCard.tsx`
- `src/components/modals/connection/EngineCard.tsx` (correct — has its own inline URL check, duplicated rather than shared)

## Impact

Every currently-published external plugin on the Tabularium registry (`csv`, `elasticsearch`, `mongodb`, `mongodb-atlas`, `duckdb`, `google-sheets`, `clickhouse`, `db2`, `cloudflare-d1`, `dameng`, `firestore` — checked via `registry.tabularis.dev`'s `/api/plugins`) is affected identically: none of them can show branded connection-card/sidebar/connection-list icons no matter what they set `icon` to in their manifest. This isn't specific to any one plugin or icon format choice.

## Suggested fix

Extract the URL/`data:` detection + ``-rendering logic already in `EngineCard.tsx`'s `renderIcon` into a shared helper, and have `getDriverIcon` check it before falling through to the literal-string switch — e.g.:

```tsx
export function getDriverIcon(manifest: PluginManifest | undefined | null, size = 14): ReactNode {
  const iconName = manifest?.icon || "";

  if (/^https?:\/\//.test(iconName) || iconName.startsWith("data:")) {
    return <RegistryDriverIcon src={iconName} size={size} fallback={<Plug size={size} />} />;
  }

  // ...existing literal-string switches unchanged
}
```

This would also let `EngineCard.tsx` drop its duplicated inline check and call the shared `getDriverIcon` directly, matching its own comment ("Priority: brand SVG icon → lucide icon → generic fallback") which already implies URL-based icons should be a first-class case.

## Reproduction

1. Any plugin manifest with `"icon": "https://.../some-icon.svg"` (e.g. `tabularis-postgresql-plugin`, `tabularis-mongodb-plugin`, `tabularis-dynamodb-plugin` — several already do this)
2. Open **Add Connection → Choose a database** — the icon renders correctly (goes through `EngineCard`'s `renderIcon`).
3. Complete the connection and save it — the connection card, sidebar entry, and connection list all show the generic `Plug` icon instead (all go through `getDriverIcon`/`getConnectionIcon`).

aesslinger:
Closes #632

## Summary

Every external plugin's icon fell through to the generic `Plug` icon in 7 of the 8 places the app renders a driver/connection icon, regardless of what its `.tabularium` `icon` field was set to. `getDriverIcon` (`src/utils/driverUI.tsx`) only ever compared the icon string against 6 hardcoded literals — a URL or `data:` URI (a documented, valid manifest value per the Tabularium docs) could never match any of them and always fell to the default `Plug` case. Only `EngineCard.tsx`'s own separate, duplicated icon-rendering logic (used solely by the "Choose a database" picker) had a working URL check.

Every currently-published external plugin on the registry is affected identically — confirmed in the issue.

## Priority order

@debba [confirmed the intended 3-tier priority](https://github.com/TabularisDB/tabularis/issues/632#issuecomment-5295682155) on the issue: **Connection Custom Icon > Manifest Icon (if present) > Plugin Icon**.

- `getConnectionIcon` already enforced the outer tier (per-connection `appearance.icon` override checked before falling back to `getDriverIcon`) — unchanged here.
- This PR completes the middle tier (manifest-supplied URL/`data:` icon), which was previously dead code — no manifest value could ever reach it.
- The inner tier (built-in brand SVG / legacy lucide names / generic fallback) is `getDriverIcon`'s existing literal-string switch, now only reached once the URL check has been ruled out.

## What changed

Two commits:

1. **`refactor: move RegistryDriverIcon out of the connection-modal folder`** — moves the existing ``-with-fallback component from `src/components/modals/connection/` to `src/components/` (matching `ConnectionIconImage`'s sibling placement), so the general-purpose `driverUI.tsx` can reuse it without reaching into one modal's private folder. No behavior change, just updates the 2 existing importers.

2. **`fix: getDriverIcon never checked for a URL/data: manifest icon`** — adds the URL/`data:` check as the first branch in `getDriverIcon`, fixing all 7 affected call sites at once (they all route through `getDriverIcon`/`getConnectionIcon`). Also corrects `PluginManifest.icon`'s doc comment, which never documented URL/`data:` URI as valid values.

## Explicitly out of scope

`EngineCard.tsx`'s own `renderIcon` is left untouched beyond its import path — it has its own separate `Database` fallback for the URL-check branch (vs. `getDriverIcon`'s `Plug`), and unifying those two fallbacks is a small but real behavior change beyond what this issue asks for.

## Verification

- `pnpm tsc --noEmit` — clean
- `pnpm lint` — clean
- `pnpm vitest run` — 3734/3767 (33 pre-existing/unrelated failures — a `localStorage.clear()` jsdom gap in this local environment, confirmed present on a clean checkout, unrelated to this change)
- New `tests/utils/driverUI.test.tsx` (9 tests) covers: URL icon → ``, `data:` icon → ``, literal-string brand/lucide/fallback cases still work unchanged, and the full 3-tier priority order via `getConnectionIcon` (connection override beats manifest URL, manifest URL beats built-in fallback)
- Existing `tests/components/connection/EngineCard.test.tsx` (3 tests) still pass unchanged


aesslinger:
## Local verification notes

Ran a local review pass against `.rules/react.md`, `.rules/typescript.md`, `.rules/general.md` and independently verified every claim in the PR description against the actual code on this branch (not just skimmed the diff).

**Rules compliance** — all clean: no `react.md` #3 (Fast Refresh) concern (`driverUI.tsx` isn't a Context/component-export file), no `any` usage (the `as PluginManifest` casts in the new tests mirror an existing pattern already in `EngineCard.tsx`, not something new), English-only comments throughout.

**Claims independently reproduced, not taken on faith:**
- Re-grepped every `getDriverIcon`/`getConnectionIcon` caller fresh — confirmed exactly 7 external call sites plus `EngineCard.tsx`'s own separate call (the 8th), matching the PR description precisely. None missed, none orphaned after the `RegistryDriverIcon` move.
- Read `getConnectionIcon`'s current implementation directly — confirmed it's genuinely unchanged and already enforces the "connection override checked before falling back to `getDriverIcon`" behavior the PR description describes.
- Confirmed `EngineCard.tsx`'s `renderIcon` is untouched beyond its import path, and its `Database`-vs-`Plug` fallback divergence from `getDriverIcon` (called out as an intentional scope boundary) is real, not accidental.
- Test counts, `tsc`, and `pnpm lint` all reproduced exactly: 12/12 on the two targeted test files (9 new + 3 existing `EngineCard` tests, unchanged), full suite 3734/3767 with the same 33 pre-existing/unrelated failures cited in the description, `tsc --noEmit` clean, `pnpm lint` clean.

**One thing noted, not a blocker**: found 2 other icon-rendering sites in the codebase not covered by this PR's "8 places" framing — `PluginInstallConfirmModal.tsx`'s local `PluginIcon` and `PaletteResults.tsx`'s command-palette icons. Verified both are genuinely unaffected: `PluginIcon` already renders any string unconditionally as an `` (no literal-string bug to begin with), and `PaletteResults.tsx`'s icons are unrelated lucide-name icons for palette actions, not plugin manifest icons. No gap — just noting that "8 places" was scoped to `getDriverIcon`'s actual callers, not a literal count of every icon anywhere in the app.

**Manual test plan** (real data, no synthetic fixture needed — the real `tabularis-postgresql-plugin` manifest already declares a URL icon, exactly the scenario #632 describes):

| Step | Where | Before this fix | After this fix |
|---|---|---|---|
| Choose a database picker | `EngineCard` | ✅ branded SVG (already worked) | ✅ unchanged |
| Sidebar entry | `OpenConnectionItem` / `ConnectionGroupItem` | ❌ generic `Plug` | ✅ branded SVG |
| Connections page, card view | `ConnectionCard` | ❌ generic `Plug` | ✅ branded SVG |
| Connections page, list view | `ConnectionListItem` | ❌ generic `Plug` | ✅ branded SVG |
| Appearance settings preview | `AppearanceSection` | ❌ generic `Plug` | ✅ branded SVG |
| Visual explain modal | `VisualExplainModal` | ❌ generic `Plug` | ✅ branded SVG |

No known blockers.

debba:
Hi @aesslinger, while reviewing I pushed two small commits on top of your branch:

- 21a30b39 extracts a shared `isUrlIcon` helper in `driverUI.tsx` and makes the URL/`data:` scheme check case-insensitive, so manifests with e.g. `HTTPS://` or `DATA:` icons are handled the same way everywhere (NewConnectionModal, EngineCard, InstallGate).
- e00a1891 consolidates the driverUI-related tests under `tests/utils/driverUI.test.tsx` to match the repo's testing conventions.

Both test files pass locally (25/25). Let me know if you agree with the changes, happy to adjust or revert if you'd prefer a different approach.

aesslinger:
> Hi @aesslinger, while reviewing I pushed two small commits on top of your branch:
> 
> * [21a30b3](https://github.com/TabularisDB/tabularis/commit/21a30b394602a20a733d0d7f1cdb7af3b47e9191) extracts a shared `isUrlIcon` helper in `driverUI.tsx` and makes the URL/`data:` scheme check case-insensitive, so manifests with e.g. `HTTPS://` or `DATA:` icons are handled the same way everywhere (NewConnectionModal, EngineCard, InstallGate).
> * [e00a189](https://github.com/TabularisDB/tabularis/commit/e00a189111b3490a8a57476863b360cc6ae61438) consolidates the driverUI-related tests under `tests/utils/driverUI.test.tsx` to match the repo's testing conventions.
> 
> Both test files pass locally (25/25). Let me know if you agree with the changes, happy to adjust or revert if you'd prefer a different approach.

Looks good! Thanks!

PR #650, opened 2026-08-15
feat: follow system light/dark appearance with per-mode theme selection

iamthenuggetman:
Closes #649

## Summary

Users can now pick separate light and dark themes; in **Follow System** mode the app watches the OS appearance (`prefers-color-scheme`) and live-applies the theme for the current mode, including the native window chrome via Tauri `setTheme`. **Static** mode keeps a single fixed theme (previous behavior, still the default).

## Changes

- **Backend** (`src-tauri/src/config.rs`): three new optional `AppConfig` fields — `followSystemTheme`, `lightThemeId`, `darkThemeId` — persisted to `config.json` via the existing field-merge `save_config`. Zero migration: absent fields = static mode, existing users unaffected. Serde round-trip + defaults tests added.
- **Resolution** (`src/utils/themeManagement.ts"): new pure `resolveActiveThemeId(settings, systemIsDark)` helper; per-mode picks resolve through the previously dead `getSystemThemeId`.
- **Provider** (`src/contexts/ThemeProvider.tsx`):
  - hydrates the new settings from `get_config` and persists them via `updateSettings` (was a no-op);
  - `matchMedia` change listener now uses the user's per-mode picks (previously hardcoded `tabularis-dark`/`tabularis-light` behind a flag nothing could set) with mode-correct preset fallback if a picked theme no longer resolves (e.g. deleted custom theme);
  - toggling Follow System applies the current-mode theme immediately;
  - syncing native window chrome (`getCurrentWindow().setTheme`) on every theme application — logged and non-fatal on failure;
  - deleting a custom theme resets per-mode picks that referenced it.
- **UI** (`src/components/settings/AppearanceTab.tsx`): Static / Follow System `SettingButtonGroup`; in follow mode two pickers filtered by theme classification (`monacoTheme.base`, covers custom themes); static mode picker unchanged.
- **i18n**: 6 new keys (`themeMode`, `themeModeDesc`, `themeModeStatic`, `themeModeSystem`, `lightTheme`, `darkTheme`) in all 11 locales.
- Editor theme is untouched: "Same as App" resolves through `currentTheme` and follows system switches automatically; explicit editor overrides stay fixed.

## Testing

- `pnpm vitest run`: 3789 tests pass (new: resolution helper unit tests; provider tests for load hydration, persistence, immediate apply on toggle, bidirectional OS switching, unresolvable-pick fallback, static-mode immunity; AppearanceTab UI tests).
[...]

PR #651, opened 2026-08-15
feat(theme): support system light and dark themes

### LINKED ISSUE #649: [Feature]: Follow system light/dark appearance with per-mode theme selection
## Problem

Currently the selected theme stays fixed regardless of the OS light/dark appearance. Users who switch their system between light and dark mode must manually change the app theme each time.

## Requested behavior

1. Let the user select a **light theme** and a **dark theme** separately.
2. **Follow System** mode: the app watches the OS appearance (`prefers-color-scheme`) and live-applies the user's theme for the current mode — including the native window chrome.
3. **Static** mode: one fixed theme, ignoring the OS appearance (current behavior, remains the default).

## Implementation notes

- New optional `AppConfig` fields: `followSystemTheme`, `lightThemeId`, `darkThemeId` (config.json, zero migration — absent fields = static mode).
- Light/dark pickers filtered by theme classification (`monacoTheme.base`); custom themes included.
- Editor theme with "Same as App" follows automatically; explicit editor overrides stay fixed.

Design spec: `docs/superpowers/specs/2026-08-15-system-theme-sync-design.md`

Soundcreates:
## Summary

- add persisted light and dark theme mappings
- add optional Follow system appearance mode using `prefers-color-scheme`
- expose filtered light/dark theme pickers in Appearance settings
- preserve SQL editor `Same as App` synchronization

Closes #649

## Validation

- `pnpm typecheck`
- `pnpm lint`
- `pnpm build`
- `cargo check --manifest-path src-tauri/Cargo.toml`
- `git diff --check`

The repository-wide `cargo fmt --check` currently reports unrelated pre-existing formatting differences outside this change.


debba:
Thanks a lot for this, and sorry for the timing overlap: #650 was opened by the issue author a few minutes before yours and lands the same config fields and mode switch, plus the native window chrome sync via Tauri setTheme (the Linux/GNOME CSD case discussed in #649), all 11 locales and test coverage. I've merged that one and I'm closing this to avoid two parallel implementations of the same feature.

Two things from your PR I'd genuinely like to see as follow-ups if you're up for it:

1. hydrating settings.customThemes from the loaded custom themes in ThemeProvider (#650 doesn't do that and it looks like a real gap);
2. defaulting followSystemTheme to true for fresh installs (no theme in config.json), keeping Static for existing configs, as raised in the #649 thread.

Happy to review either quickly.

PR #657, opened 2026-08-17
Relayout editor during resize

### LINKED ISSUE #113: [Bug]: Display issue when resizing SQL editor
### Describe the bug

https://github.com/user-attachments/assets/62ad6268-00c4-4a9f-b5f5-a807b0f3322d

### To Reproduce

Drag to resize the SQL editor.

### OS Version

macOS 26.4 (25E246)

### Tabularis Version

Version 0.9.12 (0.9.12)

### Relevant Log Output

```shell

```

DhruvShah-Dev:
## Summary

- Explicitly relayout the active Monaco editor while the SQL editor resize handle is dragged.
- Relayout once more after releasing the handle so Monaco uses the final committed panel height.

## Why

Issue #113 reports visual glitches while resizing the SQL editor. The resize handler updates the editor panel height directly during drag and commits React state on mouseup. Monaco's automatic layout can lag behind those manual DOM height writes, leaving stale measurements during the drag.

Fixes #113.

## Validation

- pnpm.cmd typecheck
- pnpm.cmd lint


debba (approved):
Checked this out locally and read through the resize path. The mechanism checks out: the wrapper runs with `automaticLayout: true`, and Monaco's ResizeObserver is async, so it lags a frame or two behind the manual height writes the drag handler does inside its rAF. Calling `layout()` in the same rAF right after the height write gives the correct ordering (DOM write first, then measure), and the extra relayout on mouseup is a nice safety net for the committed height. This is the standard fix for that class of glitch.

A few things I looked at specifically:

- Split view is not a gap here: within a pane only the active tab's panel is visible, and the `[data-editor-panel]` query is scoped to that pane's root, so "relayout the active editor" is exactly "relayout the only visible editor".
- Query builder tabs and any tab without a registered Monaco instance degrade to a no-op through the optional chaining, so behavior there is unchanged.
- Capturing `activeEditor` once at drag start is fine, the overlay blocks any interaction that could switch tabs mid-drag.
- The branch is a fair bit behind main but merges clean, and the `startResize` area on main has not changed in the meantime.

`pnpm typecheck` and `pnpm lint` are both clean on the branch. No automated tests, but there are none for Editor.tsx today and this is rAF plus real mouse geometry, so I'd rather verify it by dragging the handle than mock all of that. I'll do exactly that with a long wrapped query before the next release.

Thanks for the fix!

PR #666, opened 2026-08-19
fix(postgres): support client certificate authentication (mTLS)

adisusilayasa:
## Description
Fixes PostgreSQL TLS connections when client certificates (`ssl_cert`) and private keys (`ssl_key`) are provided.

Previously, `build_postgres_tls_connector` in `src-tauri/src/pool_manager.rs` hardcoded `.with_no_client_auth()` across all SSL modes, causing PostgreSQL servers requiring client-side certificate authentication (e.g. Google Cloud SQL with mTLS enabled) to reject connections with `connection requires a valid client certificate`.

## Changes
- Added `load_client_auth_from_pem` to parse client certificates and private keys from PEM files using `rustls_pemfile`.
- Updated `build_postgres_tls_connector` to attach client credentials via `.with_client_auth_cert(...)` when `ssl_cert` and `ssl_key` are supplied.
- Included `ssl_cert` and `ssl_key` in `build_connection_key` for PostgreSQL connection pool keying.
- Added unit tests covering client certificate loading, connector configuration, and pool key changes.

### DIFF
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -6459,7 +6459,7 @@ dependencies = [
 
 [[package]]
 name = "tabularis"
-version = "0.19.0"
+version = "0.20.0"
 dependencies = [
  "aes-gcm",
  "argon2",
--- a/src-tauri/src/pool_manager.rs
+++ b/src-tauri/src/pool_manager.rs
@@ -7,7 +7,7 @@ use rustls::client::{verify_server_cert_signed_by_trust_anchor, WebPkiServerVeri
 use rustls::crypto::verify_tls12_signature;
 use rustls::crypto::verify_tls13_signature;
 use rustls::crypto::CryptoProvider;
-use rustls::pki_types::{CertificateDer, ServerName, UnixTime};
+use rustls::pki_types::{CertificateDer, PrivateKeyDer, ServerName, UnixTime};
 use rustls::server::ParsedCertificate;
 use rustls::DigitallySignedStruct;
 use rustls::{ClientConfig, Error as TlsError, RootCertStore};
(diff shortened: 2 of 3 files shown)


debba:
Hi @adisusilayasa 

Tested this locally against a Postgres 16 container with `clientcert=verify-full` in pg_hba.conf, which reproduces the Cloud SQL failure mode from the description. The fix works: without a client cert the server rejects with "connection requires a valid client certificate", with cert and key it connects and queries fine, and a cert signed by a different CA gets rejected as expected. All 66 pool_manager tests pass on the branch, including the new ones.

One edge case worth addressing before merge: `build_postgres_tls_connector` runs unconditionally, so with `ssl_mode=disable` the new code still reads and validates the cert/key files even though TLS is never negotiated (the pg config maps disable to `PgSslMode::Disable`). The connection modal hides the cert fields when SSL is disabled but doesn't clear them, so a connection that once had a client cert configured and was later switched to disable will now fail to connect if the PEM files were moved or deleted, or if only one of the two fields is set. Before this change those fields were simply ignored for postgres, so this is a small behavior regression. Skipping client auth loading when the mode is disable would restore the old behavior there:

```rust
let client_auth = if ssl_mode == "disable" {
    None
} else {
    match (user_cert, user_key) {
        // ...
    }
};
```

Small nit on `test_load_client_auth_from_pem_valid`: it writes fixed filenames into `env::temp_dir()`, so two test runs on the same machine can collide, and the cleanup lines never run if an assertion fails earlier. Using the `tempfile` crate would handle both.

Nothing else blocking from my side, the rest looks good.

adisusilayasa:
Thanks @debba! Updated:
- Skipped client certificate & key loading when `ssl_mode == "disable"` (and ignored cert/key in the connection pool key when SSL is disabled).
- Updated `test_load_client_auth_from_pem_valid` to use `tempfile::tempdir()` for test isolation and auto-cleanup.
- Added tests verifying disabled SSL mode ignores client cert/key parameters and connection pool keys.

debba:
@adisusilayasa Thank you for addressing the requested changes! Everything looks good now, so I’ll proceed with the merge.

You’re more than welcome to keep contributing to Tabularis. If you’d like, you can also join us on [Discord](https://discord.com/invite/K2hmhfHRSt) — we’d be happy to have you there!

PR #679, opened 2026-08-25
feat(autocomplete): rank columns of nearest table first (closes #507)

### LINKED ISSUE #507: [Feat]: SQL autocomplete — rank columns of the nearest table first
**Problem**

Follow-up to #505. When several tables are in scope, all their columns rank equally — every suggestion gets the same `0_<name>` sortText, so Monaco falls back to alphabetical order:

```sql
SELECT * FROM orders o JOIN customers c ON |
```

At that cursor you almost certainly want a `customers` column first (you just joined it), but you get an alphabetical mix of both tables. Same after `WHERE`, where the FROM table is the more likely intent.

Keyword ranking already solved this exact problem in #505 with a `2_0_` / `2_1_` sortText split (`getKeywordRelevance`); columns should get the same treatment.

**Proposed fix**

Split the column sortText into `0_0_<name>` for the "nearest" table and `0_1_<name>` for the rest:

- in an `ON` clause, the nearest table is the one from the JOIN being joined right now;
- after `WHERE`, a reasonable default is the FROM table;
- with a single table in scope nothing changes.

The missing piece is knowing *which* table reference was seen last in the current clause. The tokenizer in `sqlContext.ts` already walks past those references — it just doesn't keep them. Record the last-seen table reference in the clause frame and expose it from `analyzeSqlContext`, then let the provider match it against the alias map when assigning sortText.

This is deliberately a heuristic — no scoring model, just "the table you mentioned last probably matters most". Cheap to compute (no extra passes) and easy to revert per-clause if it turns out to be wrong somewhere.

**Tests**

Unit tests in `tests/utils/sqlContext.test.ts` for the last-seen reference exposure (FROM, JOIN chains, subqueries restoring the outer frame), plus a provider-level test in `tests/utils/autocomplete.test.ts` asserting the ordering: in the `ON` example above, `customers` columns must sort before `orders` columns.

**Scope**

Small — it can ride along with any other autocomplete PR rather than needing its own.

adisusilayasa:
### Problem

Follow-up to #505 / closes #507. When multiple tables are in scope, all their columns were given the same `0_<name>` sortText prefix. In an `ON` clause or after `WHERE`, columns from the table being joined or the primary FROM table were mixed in alphabetical order with all other in-scope tables.

### Solution

1. **Expose `lastTableRef` in `analyzeSqlContext`:**
   - In `ON` / `USING` clauses: captures the table/alias being joined right now.
   - In `WHERE` / `SELECT` / `SET` clauses: defaults to the primary `FROM` table.
   - Properly handles subqueries, restoring the outer scope's table reference on closing parenthesis.
2. **Rank Nearest Table Columns First:**
   - When multiple tables are in scope, columns from the nearest table receive `sortText: "0_0_<col>"`, while columns from other tables receive `sortText: "0_1_<col>"`.
   - Single-table queries remain unchanged with `sortText: "0_<col>"`.

### Tests

- Added unit tests in `tests/utils/sqlContext.test.ts` for `lastTableRef` exposure across single/multi-table FROMs, aliases, schema-qualified tables, JOIN chains, subqueries, UPDATE, and INSERT INTO.
- Added provider-level sorting assertion in `tests/utils/autocomplete.test.ts` verifying that nearest table columns sort before other tables in `ON` and `WHERE` clauses.

### DIFF
--- a/src/utils/autocomplete.ts
+++ b/src/utils/autocomplete.ts
@@ -359,8 +359,22 @@ export const registerSqlAutocomplete = (
 
         const seenColumns = new Set<string>();
 
+        const nearestRef = sqlContext.lastTableRef?.toLowerCase();
+
         matchingTables.forEach((table, idx) => {
           const columns = results[idx];
+          const isNearest = Boolean(
+            nearestRef &&
+              (table.name.toLowerCase() === nearestRef ||
+                Array.from(scopedAliases.entries()).some(
+                  ([alias, ref]) =>
+                    alias.toLowerCase() === nearestRef &&
+                    ref.name.toLowerCase() === table.name.toLowerCase() &&
+                    (!ref.schema || !table.schema || ref.schema.toLowerCase() === table.schema.toLowerCase()),
+                )),
+          );
+          const sortPrefix = matchingTables.length > 1 ? (isNearest ? "0_0_" : "0_1_") : "0_";
+
           columns.forEach(col => {
             if (!seenColumns.has(col.label)) {
               seenColumns.add(col.label);
@@ -379,7 +393,7 @@ export const registerSqlAutocomplete = (
                 kind: monaco.languages.CompletionItemKind.Field,
                 detail: `${col.detail} — ${table.name}${aliasHint}`,
                 ...buildIdentifierInsert(col.label, range),
-                sortText: `0_${col.label}`,
+                sortText: `${sortPrefix}${col.label}`,
               });
             }
           });
--- a/src/utils/sqlContext.ts
+++ b/src/utils/sqlContext.ts
@@ -271,6 +271,9 @@ const applyWord = (frame: Frame, upper: string, prev: WordToken | null): void =>
     case 'SELECT':
       frame.clause = 'select';
       frame.isStatement = true;
+      frame.fromTableRef = null;
+      frame.currentJoinTableRef = null;
+      frame.lastTableRef = null;
       frame.awaitingAlias = false;
       break;
     case 'FROM':
@@ -315,6 +318,9 @@ const applyWord = (frame: Frame, upper: string, prev: WordToken | null): void =>
       if (prev?.upper === 'FOR' || prev?.upper === 'KEY') break;
       frame.clause = 'update';
       frame.isStatement = true;
+      frame.fromTableRef = null;
+      frame.currentJoinTableRef = null;
+      frame.lastTableRef = null;
       frame.awaitingAlias = false;
       break;
     case 'SET':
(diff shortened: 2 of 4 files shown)


debba (approved):
Hi @adisusilayasa, thanks for the follow-up on #507. I pulled the branch locally, ran the suite (3868 tests green, typecheck and lint clean) and exercised the analyzer with about thirty extra inputs beyond the ones in the tests: JOIN chains, USING, GROUP BY / HAVING / ORDER BY, quoted identifiers in MySQL / Postgres / SQL Server style, schema-qualified names, CTEs, comments and string literals containing keywords, nested subqueries and statement separators. Everything resolved to the expected table. Nice work, the frame-based tracking fits the existing analyzer well.

Approving as is. A few things I would suggest as follow-ups (or in this PR if you prefer, none of them are blockers):

1. **Shared column names lose the nearest-table boost.** Columns are deduplicated by label in `scopedAliases` insertion order, so a column that exists in both tables is emitted from whichever table was parsed first. With `SELECT * FROM orders o JOIN customers c ON |` and both tables having an `id` column, `customer_name` gets `0_0_` but `id` gets `0_1_` and is attributed to `orders`. Since `id` is exactly what you type in an ON clause, this undercuts the feature in its headline case. Iterating the nearest table first before the dedupe loop fixes it, e.g. compute `isNearest` per table and then:

   ```ts
   const ordered = [...matchingTables.entries()].sort(
     ([, a], [, b]) => Number(isNearest(b)) - Number(isNearest(a)),
   );
   ordered.forEach(([idx, table]) => { const columns = results[idx]; ... });
   ```

2. **Comma-separated FROM lists return the last table, not the first.** `SELECT * FROM orders o, customers c WHERE |` resolves to `c`, because `,` does not reset `awaitingAlias`, so `customers` and `c` are consumed as aliases of `orders`. The result is a reasonable "last mentioned" heuristic, but the `SqlContextInfo` doc comment says "primary FROM table", and there is no test for the comma form. Either add a test and reword the comment, or reset `awaitingAlias` when a `,` is seen in the `from` clause.

3. **SELECT list.** The description says SELECT defaults to the FROM table, but the analyzer only sees text before the cursor, so in `SELECT | FROM orders o JOIN customers c ...` the ref is null and all columns get `0_1_`. Harmless (relative order is unchanged), just worth adjusting the description so it does not promise more than it does.

4. Minor: quoted identifiers now carry their text into `prev.upper` where they used to be `''`. A quoted `"in"` or `"as"` right before `(` would be read as an IN list or CTE body. Purely theoretical, mentioning it only for completeness.

Thanks again!

PR #685, opened 2026-08-28
Implement builtin PostgreSQL -> plugin migration flow

aesslinger:
## Summary

Implements the builtin PostgreSQL → plugin migration flow designed in [`.github/planning/postgres-plugin-force-install.md`](.github/planning/postgres-plugin-force-install.md). All 5 chunks of the design are implemented, unit-tested, and manually verified in a dev build:

- **On-demand plugin install** — the `postgresql` plugin installs automatically, at app startup, only when at least one saved connection is on the builtin `postgres` driver and the installed version is compatible (`force_install.rs`). This re-triggers on every launch: a user who uninstalls the plugin while a builtin-`postgres` connection still exists gets it silently reinstalled and reactivated on the next launch, with no opt-out. Accepted for this release — flagging it here per review feedback rather than leaving it undocumented.
- **Deprecation signaling** everywhere the builtin driver is surfaced: a "Deprecated" badge (with tooltip and tentative removal date) in the connection-type picker, Settings → Plugins, and each connection row
- **Migration entry points**: a dismissible banner (nudge or "couldn't be downloaded" variant, gated on plugin readiness + registry connectivity), a per-connection row button + context-menu action, and a bulk "Review connections" checklist modal
- **Capability-gap detection** (`findUnsupportedFeatures.ts`) — compares each connection against the installed plugin's declared capabilities (SSL, connection strings) and defaults an unsupported or connection-URI-based (keychain secret won't carry over) connection to **unchecked** in the bulk checklist, with the specific gap named inline
- **Reversible migration, with no silent failures** — post-migration connection test, one-click Undo, persisted migration history. Every failure mode is caught and surfaced as a toast rather than left to fail silently:
  - a **connection-level** failure (post-migration test fails) or **process-level** failure (plugin didn't start) both offer Undo + a pre-filled GitHub Issue Form
  - an **unexpected** failure (e.g. the connection was deleted or renamed concurrently, or the IPC call itself errors) resolves to a `"failed"` outcome instead of an unhandled rejection, and still offers Undo
  - **Undo itself** can fail (e.g. the same concurrent-delete race) — that's now surfaced as its own error toast instead of disappearing into an unhandled rejection
  - the bulk checklist's migrate loop is resilient per-row: one connection's failure no longer aborts the remaining batch or leaves the "Migrate N selected" button permanently disabled
- **Issue reporting** via structured GitHub Issue Forms (`migration-failure.yml`, `capability-gap.yml`, `bug_report.yml`) in the plugin repo, with a static fallback repo URL so reporting still works if the Tabularium registry API is unreachable
- **`migrationModeByDriver`** staged for a later, evidence-gated switch to forced migration — today every driver is `opt-in`

Everything is parameterized by `(builtinId, pluginId)` so the next deprecation (mysql, sqlite) is additive, not a redesign.

**Not in this PR:** Forced migration mode is designed but not enabled — `migrationModeByDriver` stays `opt-in` for all drivers pending real adoption data. The design doc (`.github/planning/postgres-plugin-force-install.md`) is scratch planning material, not permanent docs — it will be removed from this PR as a last step, just before this comes out of draft, rather than in a follow-up.

## Screenshots

Screenshot 2026-09-01 at 2 50 22 PM
Screenshot 2026-09-01 at 2 50 54 PM
Screenshot 2026-09-01 at 2 51 24 PM

## Test plan

Manual verification steps (all confirmed working in a local dev build against `connections.dev.json` test fixtures — a debug-only override, never the real `connections.json`):

1. **Force-install trigger** — with a saved connection on the builtin `postgres` driver and the `postgresql` plugin not yet installed, launch the app. Expect: the plugin installs automatically in the background (Settings → Plugins shows it installed and active), no user action required.
2. **Migration banner appears** — with at least one builtin-postgres connection and the plugin ready, open the Connections list. Expect: a dismissible banner reading "Try the new PostgreSQL plugin…" with a "Review connections" link and a dismiss (×) button.
3. **Banner dismissal and resurface** — dismiss the banner, then add a new builtin-postgres connection. Expect: the banner stays hidden for the connections that existed at dismissal time, but resurfaces once a new, not-yet-dismissed builtin connection exists.
4. **Deprecated badge everywhere** — check the connection-type picker (+ Add Connection), each builtin-postgres connection card/row, and Settings → Plugins' builtin PostgreSQL card. Expect: a "Deprecated" badge with a tooltip naming the replacement plugin and tentative removal date on all three surfaces; the plugin's own catalogue entry has no such badge and sorts ahead of the builtin.
5. **Per-connection migrate button** — on a builtin-postgres card, click the migrate (arrows) button, or right-click → "Switch to plugin." Expect: a confirm dialog ("Switch `<name>` to the PostgreSQL plugin? If it's currently connected, it'll be disconnected and reopened using the plugin"), then the driver flips to `postgresql` and a post-migration connection test runs, surfacing a success or failure toast with Undo (and Report an issue, on failure).
6. **URI pre-flight warning** — repeat step 5 on a connection with a connection-string/URI stored in the keychain. Expect: the confirm dialog is replaced with a variant warning that the stored string won't carry over and will need to be re-entered.
7. **Undo** — click Undo on the post-migration toast (or right-click → "Switch back to builtin" on an already-migrated card). Expect: the connection flips back to `postgres` and its badge/action reverts.
8. **Safety gate on non-migratable drivers** — create a MySQL (or other non-postgres) connection. Expect: no migrate button and no "Switch" context-menu action appear on it — the migration hook only fires for the wired `postgres:postgresql` pair.
9. **Bulk migration checklist** — with 2–3 builtin-postgres connections in different states (clean, one using a feature the plugin doesn't declare, one URI-based), click "Review connections" on the banner. Expect: the checklist modal opens (not the New Connection picker); the clean connection is checked by default; a connection using an unsupported capability is unchecked with the specific gap named inline and a "Report this gap" action; a URI-based connection is unchecked with an inline keychain warning. Click "Migrate N selected" — only the checked rows migrate, sequentially, with per-row status.
10. **Report an issue / Report this gap** — click either report action. Expect: a browser tab opens to the plugin repo's pre-filled GitHub Issue Form (`migration-failure.yml` or `capability-gap.yml`) with plugin/app version, OS, and the specific error or feature already filled in; re-opening the checklist shows "Reported" instead of the button for a gap already filed. This works even when the Tabularium registry API is unreachable (confirmed by disabling network access to the registry in this sandbox), via a static fallback repo URL for the first-party plugin.
11. **Failure surfaces a toast instead of failing silently** — remove a connection's on-disk record while it's still shown in the UI (simulating a concurrent delete/rename), then click "Switch to plugin" on it. Expect: an error toast reading `Couldn't switch <name> to the plugin: Connection not found`, with an Undo action — **confirmed live** against this exact scenario; before this PR's error-handling fix, the same action failed with no toast, no console error, and no other visible feedback.
12. **Automated checks** — `pnpm tsc --noEmit` (clean), `pnpm lint` (clean, one pre-existing unrelated warning in `ThemeProvider.tsx`), `pnpm vitest run` (246 files / 4050 tests — 43 failures in `useSidebarResize`/`SettingsProvider`/`ThemeProvider`, all reproducing identically with none of this PR's commits present, so a local environment issue rather than a regression from this branch), and `cargo test --lib` (1219 tests — 5 failures in `askpass`/`storage_location_tests`, same story: zero diff on those files from this PR's own commits). Every module this PR actually touches (`plugins::force_install`, `plugins::driver`, `pool_manager_tests`, and the touched `tests/hooks`/`tests/components`/`tests/pages` files) passes cleanly on its own.

## Review follow-ups (@debba, 2026-09-03)

All three blocking issues and the smaller items from review are addressed, each as its own commit:

- **Activation could silently disable other installed plugins** — `merge_active_driver` now seeds from `installer::list_installed()` when no preference has been saved yet, instead of starting from an empty list and narrowing the implicit "all installed" set to just the plugin being force-installed.
- **Frontend/backend activation race** — the backend emits `tabularis://plugin-activated` after persisting activation; `useDrivers` and `SettingsProvider` both listen and refresh, instead of the frontend staying on a stale snapshot until restart (or clobbering the write on the next unrelated setting save).
- **Migration checklist's default selection computed before data loads** — first moved from a `useState` initializer to a `useEffect` gated on the rising edge of `isOpen`; later superseded by a simpler fix (`43428ce3`): `Connections.tsx` now conditionally mounts `MigrationChecklistModal` (`{isMigrationChecklistOpen && <MigrationChecklistModal ... />}`) instead of always rendering it, so the original plain `useState` initializer runs exactly once per open and the rising-edge effect was removed.
- Migration/undo now actually reconnect a connection that was open before the flip (the confirm dialog already promised this).
- Added the missing `connections.deprecated*`/`common.dismiss`/`connectionCatalogue.deprecated*` translations across all 10 non-English locales, and dropped a leftover `[Switch to plugin]` placeholder from the English source.
- The outcome toast is now shown at the point `migrateConnection` resolves instead of via a `lastOutcome` effect — fixes a `.rules/react.md` violation and a real bug where the bulk checklist's migrations were also firing this toast on top of their own per-row status UI.
- `force_install.rs`'s tests moved to a sibling `force_install_tests.rs`, per `.rules/rust.md`.
- The banner's tentative removal date is now read from `PluginManifest.deprecated.removal_date` (the same field the badge/tooltip use) instead of being a second hardcoded date in the translation string.
- `Connections.tsx` now reads `builtinId`/`pluginId` back from `useBuiltinDriverMigration` instead of hardcoding `"postgres"`/`"postgresql"` at the report-issue and checklist-modal call sites.

## Status

Draft — ready for review, not yet ready to merge. Please review the manual test steps above against your own read of the design doc before this comes out of draft.


debba (changes requested):
Hi @aesslinger,

I finally sat down with this one properly. I checked out the branch, ran the tests (the six touched vitest files are green, 130 tests, and `cargo test --lib force_install::` passes 8/8), confirmed the three issue forms are actually in the plugin repo and that the registry serves beta.9 with `minRuntimeVersion` 0.20.0, and then I ran it against my own config and a real Postgres container. Really nice work overall, especially the "never reject" contract in the hook, that part reads very well.

That said, I hit three things that I think we need to sort out before this leaves draft, because they bite real users. One of them bit me directly.

**The activation write can disable every other plugin.** `load_plugins` treats `active_external_drivers: None` as "load everything installed" (there's a comment in `lib.rs` saying exactly that), and `PluginsTab` / `Settings.tsx` do the same on the frontend. But `merge_active_driver(None, "postgresql")` produces `Some(["postgresql"])`, so on the next launch everything else is skipped as disabled. My machine has `dynamodb` and `sqlserver` installed and `active_external_drivers: null` in config.json, so launching this branch would have quietly turned both of them off. I'd seed the list from `installer::list_installed()` when the current value is `None`, so the implicit "all active" set becomes explicit before you append. If you pass the installed ids into `merge_active_driver` it stays a pure helper and the tests keep working.

**The background write races with the frontend settings snapshot.** `updateSetting` sends the whole settings object to `save_config`, and the Rust merge treats any `Some(active_external_drivers)` as authoritative. So: app starts with `["mssql"]` in the frontend, the force-install finishes a few seconds later and writes `["mssql", "postgresql"]`, the user changes the theme, the frontend sends `["mssql"]` again and the plugin is deactivated on next launch. Related: nothing tells the frontend that the install finished, so on the very first launch the banner shows the "couldn't be downloaded" variant even though it worked, until you restart. An event from `ensure_plugin_installed` that `useDrivers` / `SettingsProvider` listen to would fix both. Or you could do the activation write from the frontend on that event, the same way `NewConnectionModal` already does it.

**The checklist's default selection is computed once and never again.** The `useState` initializer in `MigrationChecklistModal` runs when `Connections.tsx` first renders, which is before `loadConnections()` resolves and before `useDrivers` has the plugin manifest. The comment says the component remounts with the parent's conditional render, but `Connections.tsx` renders `<MigrationChecklistModal isOpen={...}>` with no conditional around it. On a cold start every row is unchecked and "Migrate 0 selected" is disabled. The existing tests pass because they render with `isOpen` and data already there. This fails on the branch:

```tsx
const { rerender } = render(<MigrationChecklistModal isOpen={false} connections={[]} manifest={undefined} {...props} />);
rerender(<MigrationChecklistModal isOpen connections={[cleanConn]} manifest={manifest} {...props} />);
expect(screen.getByRole("checkbox")).toBeChecked(); // Received element is not checked
```

Easiest fix is `{isMigrationChecklistOpen && <MigrationChecklistModal ... />}` so the initializer really does run per open, or derive the default set with `useMemo` and reset it when the modal opens.

A few smaller things, in rough order of importance:

- `undoMigration` doesn't disconnect an open connection before flipping the driver back, unlike `migrateConnection`. And neither direction actually reopens the connection, while the confirm text promises "it'll be disconnected and reopened". Either reconnect after a successful test or change the copy to "you'll need to reconnect".
- i18n: `connections.deprecated`, `connections.deprecatedTooltip` and `connections.deprecatedTooltipDate` are missing from all ten non-English locales. `ko`, `ru` and `tl` also miss `common.dismiss`, `connectionCatalogue.deprecated` and `connectionCatalogue.deprecatedTooltip`.
- There's a merge conflict with `main`, only in `src/types/plugins.ts`.
- The outcome-toast `useEffect` in `Connections.tsx` calls `showToast` and `clearOutcome()` synchronously, which is the pattern `.rules/react.md` #2 asks us to avoid. Since `migrateConnection` already resolves with the outcome, you could show the toast right where the promise resolves and drop the `lastOutcome` state entirely. That would also let the checklist decide whether it really wants N sticky toasts on top of the per-row status.
- `force_install.rs` has about a hundred lines of inline tests; per `.rules/rust.md` #4 they should live in a sibling `tests.rs`.
- The removal date lives in two places: `deprecation_for_builtin` has `2026-10-05` and `migration.banner.nudge` hardcodes "Oct 5, 2026". I'd interpolate `removal_date` from the manifest.
- `connections.deprecatedTooltip` ends with a literal `[Switch to plugin]` inside a `title` attribute, looks like a leftover placeholder.
- `handleReportIssue` in `Connections.tsx` hardcodes `"postgres"` / `"postgresql"` while the hook is parameterized.
- Your design doc says the reinstall-after-uninstall behaviour should be called out in the PR description, but it isn't. Someone who uninstalls the plugin but keeps a builtin connection gets it reinstalled and re-activated on every launch with no way to opt out. I'm fine with that as a decision, I just want it written down.
- Since the description already says the planning doc is scratch material, I'd remove it in this PR rather than in a follow-up.

One more observation from the real data: the published plugin declares `supports_ssl: true` and `connection_string: true`, so with the live plugin no capability gap can ever show up in the checklist. That path is only reachable from the unit tests today. Not a problem, just something to keep in mind when reading the "unchecked by default" behaviour.

Thanks again for the thoroughness on this, the error handling and the design doc made the review much easier than it could have been.

PR #712, opened 2026-09-04
fix(snap): repair snap sandbox integration 

### LINKED ISSUE #710: [Bug]: Snap package: SQL editor stuck at "Loading..." — bundled WebKitGTK 2.46 segfaults (deb with system WebKit 2.50 works)
### Describe the bug

On the **Snap** package, the SQL editor never initializes: every Console tab shows a centered **"Loading..."** forever, while everything else in the app (sidebar, schema browser, table data grid, EXPLAIN, export) works normally. Opening the WebKit inspector (right-click → Inspect Element) makes the whole window go black and hang.

`dmesg` shows the WebKitGTK bundled inside the snap segfaulting — once when the editor tries to initialize, and again when the inspector is opened:

```
kernel: tabularis[492495]: segfault at 8 ip 00007860371b9e15 sp 00007ffd0bb34f20 error 4 in libwebkit2gtk-4.1.so.0.21.10[786036916000+3ba3000]
kernel: tabularis[499017]: segfault at 8 ip 000078d98bf80e15 sp 00007ffe28658570 error 4 in libwebkit2gtk-4.1.so.0.21.10[78d98b6dd000+3ba3000]
```

The snap ships `libwebkit2gtk-4.1.so.0.21.10` (WebKitGTK 2.46.x, from the `gnome-46-2404` runtime). The **.deb** package, which uses the system WebKitGTK **2.50.4**, works perfectly on the same machine with the same connection and the same database — so this is specific to the WebKit build bundled in the snap.

Things I ruled out before finding the segfault (in case it helps triage):

- Not the data: an empty table (`audit_log`) shows the same "Loading...".
- Not the network/CDN: `@monaco-editor/react` is used without `loader.config`, so Monaco is fetched from `cdn.jsdelivr.net` at runtime — but from *inside the snap sandbox* (`snap run --shell tabularis`) the loader URL returns 200 in 0.6 s, and the GNOME proxy mode is `none`.
- Not CSP: `app.security.csp` is unset.
- Not NVIDIA DMA-BUF: launching with `WEBKIT_DISABLE_DMABUF_RENDERER=1 WEBKIT_DISABLE_COMPOSITING_MODE=1` changes nothing.

Two smaller snap-packaging issues seen along the way:

1. `password-manager-service` is declared as a plug but **not auto-connected**, so saving a connection fails with `Platform secure storage failure: DBus error: An AppArmor policy prevents this sender from sending this message to this recipient ... interface="org.freedesktop.Secret.Service" member="OpenSession"`. `sudo snap connect tabularis:password-manager-service` fixes it. Worth requesting auto-connect in the store, or documenting.
2. Journal noise at startup: `WebKitNetworkProcess: GDBus.Error:org.freedesktop.portal.Error.NotAllowed: This call is not available inside the sandbox` and `tauri_plugin_deep_link: Failed to run OS command xdg-mime: No such file or directory` — harmless, but suggests the snap is missing a couple of runtime bits.

Since Monaco is loaded from a CDN, a secondary suggestion: bundling it locally via `loader.config({ monaco })` would make the editor work offline and avoid one more moving part on Linux.

### To Reproduce

1. `sudo snap install tabularis` (0.22.0, rev 45)
2. `sudo snap connect tabularis:password-manager-service` (needed to save the connection at all)
3. Connect to any PostgreSQL (I used `pgvector/pgvector:pg16` in Docker, port 55432)
4. Open any table → Console
5. Editor area shows "Loading..." indefinitely; `dmesg | grep tabularis` shows the segfault
6. Right-click the editor → Inspect Element → window turns black and hangs (second segfault)
7. `sudo snap remove tabularis && sudo dpkg -i tabularis_0.22.0_amd64.deb` → editor works immediately

### OS Version

Ubuntu 22.04.5 LTS, kernel 6.8.0-138-generic, GNOME 42.9 on **X11**
Hybrid graphics: Intel iGPU + NVIDIA RTX 4060 Laptop, NVIDIA driver 580.173.02, `prime-select on-demand`
System `libwebkit2gtk-4.1-0` 2.50.4-0ubuntu0.22.04.1 (used by the .deb — works)
Snap runtime: `gnome-46-2404`, bundled `libwebkit2gtk-4.1.so.0.21.10` (crashes)

### Tabularis Version

0.22.0 — snap rev 45 (broken) vs `tabularis_0.22.0_amd64.deb` (works)

### Relevant Log Output

```
# journalctl -k
kernel: tabularis[492495]: segfault at 8 ip 00007860371b9e15 sp 00007ffd0bb34f20 error 4 in libwebkit2gtk-4.1.so.0.21.10[786036916000+3ba3000] likely on CPU 17
kernel: tabularis[499017]: segfault at 8 ip 000078d98bf80e15 sp 00007ffe28658570 error 4 in libwebkit2gtk-4.1.so.0.21.10[78d98b6dd000+3ba3000] likely on CPU 18

# journalctl --user (snap, before connecting password-manager-service)
[Keychain] Error setting password: Platform secure storage failure: DBus error: An AppArmor policy prevents this sender from sending this message to this recipient; type="method_call", sender=":1.3089" (uid=1000 pid=473438 comm="/snap/tabularis/45/usr/bin/tabularis " label="snap.tabularis.tabularis (enforce)") interface="org.freedesktop.Secret.Service" member="OpenSession" error name="(unset)" requested_reply="0" destination="org.freedesktop.secrets"

# journalctl --user (snap startup noise)
WebKitNetworkPr[499263]: GDBus.Error:org.freedesktop.portal.Error.NotAllowed: This call is not available inside the sandbox
tabularis_tabularis.desktop[499017]: [ERROR] tauri_plugin_deep_link::error - Failed to run OS command `xdg-mime`: No such file or directory
tabularis_tabularis.desktop[499017]: [WARN] tabularis_lib - Failed to register tabularis:// scheme: No such file or directory

# Backend was healthy the whole time (snap):
[INFO] tabularis_lib::commands - Retrieved 18 tables from airbag_srs
[INFO] tabularis_lib::commands - Retrieved 115 routines from public
[INFO] tabularis_lib::commands - Query executed successfully, returned 500 rows
[INFO] tabularis_lib::commands - Explain query completed successfully
```

wangyingsm:
Fixes #710

## What was actually wrong

While working on this I found that the premise of the issue ("bundled WebKitGTK 2.46 segfaults") is wrong, so the fix is different from what I first suggested there.

Unpacking the published snap (rev 45) shows it ships **WebKitGTK 2.52.6** (`libwebkit2gtk-4.1.so.0.21.10`, staged from noble-security by the `stage-packages` entry), *newer* than the 2.50.4 the working `.deb` uses on the same host. The gnome-46-2404 platform snap carries its own `so.0.21.7`, and the gnome extension bind-mounts the platform's `webkit2gtk-4.1/` directory (WebKitNetworkProcess/WebProcess, injected bundle) into the snap. So the app ran one WebKitGTK release's library with another release's helper processes.

The editor hang itself has a simpler cause. `@monaco-editor/react` was used without `loader.config`, so Monaco was fetched from `cdn.jsdelivr.net` at runtime. Inside the snap the gnome extension sets `GTK_USE_PORTAL=1`, which makes WebKit's network process resolve proxies through xdg-desktop-portal. The portal only allows that when snapd's `portal-info` reports `HasNetworkStatus=true`, i.e. when the `network-status` plug is connected. It was not, so every request made by the WebView itself failed with the `org.freedesktop.portal.Error.NotAllowed: This call is not available inside the sandbox` line from the issue — and Monaco never arrived. Everything else kept working because database traffic goes through the Rust side, not the WebView.

## Changes

**Frontend — Monaco is bundled, no CDN at runtime**
- `src/monaco.ts` binds the already-installed `monaco-editor` package to `@monaco-editor/react` via `loader.config({ monaco })` and installs a `MonacoEnvironment` whose workers are Vite `?worker` bundles (editor + JSON; SQL and plain text need no language service). `main.tsx` imports it first so nothing can call `loader.init()` before the loader is configured.
- `src/utils/monacoWorkers.ts` holds the label → worker selection (tested in `tests/utils/monacoWorkers.test.ts`).
- `src/utils/sqlHighlight.tsx` now initialises Monaco on first use instead of at import time, removing the module-evaluation-order dependency.
- Side effect: the editor works offline and behind restrictive proxies on every platform.

**Snap packaging (`snap/snapcraft.yaml`)**
- Add the `network-status` plug (auto-connected) so the portal stops refusing the WebKit network process.
- Stop staging `libwebkit2gtk-4.1-0`: the platform snap already provides WebKitGTK and the extension's layout already points the helper-process directory at it. This keeps library, helper processes and injected bundle from the same build, and WebKit security updates arrive through the platform snap without a Tabularis rebuild. (`libayatana-appindicator3-1` is kept.)
- Comment on `password-manager-service` explaining it is not auto-connected.

**Rust**
- New `sandbox` module (`sandbox.rs`, tests in `sandbox_tests.rs`) detecting Snap/Flatpak from `SNAP`/`FLATPAK_ID`.
- `lib.rs`: skip `deep_link.register("tabularis")` inside Snap/Flatpak. The exported `.desktop` entry already carries `MimeType=x-scheme-handler/tabularis`, and the sandbox has neither `xdg-mime` nor write access to the host's `mimeapps.list`, so the call could only fail. This removes the `Failed to run OS command xdg-mime` / `Failed to register tabularis:// scheme` startup errors.
- `keychain_utils.rs`: all keyring failures go through `keychain_error()`. A `PlatformFailure` inside a snap now ends with "Run `sudo snap connect tabularis:password-manager-service` and restart Tabularis." instead of leaving the user with the raw AppArmor/D-Bus text (tests in `keychain_utils_tests.rs`).

**Docs**
- README (Linux (Snap)): document the `snap connect` step.

## About `password-manager-service` auto-connection

This cannot be declared in `snapcraft.yaml`; only the Snap Store can grant it, on request from the publisher. @debba, could you open an auto-connect request on the snapcraft forum (category "store-requests", title like "Auto-connect request: tabularis password-manager-service", justification: credentials for database connections are stored in the system keychain via libsecret)? Until then the README note and the in-app hint cover it.

## Verification

Done here:
- `pnpm build` (tsc + vite): passes. `dist/` now contains `editor.worker-*.js` / `json.worker-*.js`; the only remaining `cdn.jsdelivr.net` strings are the loader's inert default config.
- `pnpm lint`: no new warnings. `pnpm vitest run`: 244 files / 3974 tests pass.
- `cargo test -- sandbox keychain_utils`: 9/9 pass. The new modules are clippy-clean; the ~130 pre-existing clippy warnings elsewhere are untouched.
- Built the `.deb` from this branch, extracted it, and launched `tabularis --explain plan.json` under `unshare -n` (no network interfaces at all) on Ubuntu 22.04 with system WebKitGTK 2.50.4: the Visual Explain "Raw Output" Monaco editor renders with JSON highlighting, so nothing is fetched from a CDN anymore.

Not done here (no snapcraft/LXD on this machine): building and running the snap. To verify on a snap-capable box:

```bash
pnpm tauri build --bundles deb
cp src-tauri/target/release/bundle/deb/tabularis_*_amd64.deb tabularis.deb
snapcraft pack
sudo snap install --dangerous tabularis_*_amd64.snap
sudo snap connect tabularis:password-manager-service
snap connections tabularis | grep network-status     # should show connected
tabularis  # open a table → Console: editor must render; journalctl --user should show no portal.Error.NotAllowed / xdg-mime lines
```


debba (approved):
Approved, and thanks a lot for digging into this one. Tracing the WebKit library/helper mismatch and the portal behaviour was really useful, and the fix itself is clean and well documented.

I've opened the auto-connect request for `password-manager-service` on the snapcraft forum, so hopefully the manual `snap connect` step goes away soon.

Merging now. Please keep the contributions coming, they are very much appreciated!

PR #718, opened 2026-09-05
feat(notebook): inline query plan in SQL cells, plus notebook UX fixes

harshavardhankonisa:
## Summary

This brings the notebook closer to `Jupyter Notebook` parity by promoting Visual EXPLAIN to a first-class, inline section of each SQL cell, and fixes a couple of notebook rough edges found along the way: silent result-export failures, two regex `lastIndex` bugs in the param resolvers, duplicate CTEs for repeated cell references, and cells remounting on reorder.

Everything here is scoped to the notebook feature. No backend changes.

### 1) Inline "Query Plan" section for SQL cells (new)

Previously, the only way to see a plan from a notebook cell was a small floating button in the AI overlay that popped open the full Visual EXPLAIN modal. That button is removed and the plan is now a proper cell section, rendered directly below the results, consistent with the existing Query / Results / Chart sections.

- **New `SqlCellExplain.tsx`** renders `VisualExplainView` inline (graph/tree view), with an `EXPLAIN ANALYZE` checkbox, a re-run button, and a popout to the full `VisualExplainModal`. It has its own resize handle (default 720px, 200px to 1200px) so the visualization has room to breathe.
- **Off by default, opt-in per cell.** A new `Network` icon in the cell toolbar (`NotebookCellHeader`) toggles the section. The button is highlighted (blue) while the plan is visible, and only appears for SQL cells whose driver supports EXPLAIN (`supportsExplain`, gated in `NotebookCellWrapper`).
- **Explains the resolved query, not the raw text.** `NotebookView` passes the cell's SQL through `resolveParams` and `resolveQueryVariables` before it reaches the plan section, so the plan reflects what would actually execute (`@param` values substituted, `{{cell_N}}` references expanded to CTEs). If a referenced cell has not been run successfully yet, the section shows `queryPlanUnresolved` listing the missing refs and no request is sent.
- **On-demand execution, never on every run.** When the section is first shown it runs `EXPLAIN` once (or `EXPLAIN ANALYZE` if the checkbox is ticked). We deliberately do not re-run it on each cell execution: `ANALYZE` on DML actually performs the write and doubles work on `SELECT`, so it stays behind an explicit action. If the query, schema or connection changes after a plan was fetched, the stale plan is replaced by `queryPlanOutdated` until Re-run is clicked.
- **Analyze is a per-source opt-in.** The choice is owned by the new `useExplainAnalyze` hook and bound to the query, schema and connection it was made for: editing any of them drops it, a data-modifying query is never analyzed unless the user opts in for that exact query, and an `AlertTriangle` warning shows while Analyze is on for DML. Ticking the checkbox never executes SQL; only Re-run does.
- **Persisted.** Visibility is stored per cell as `isQueryPlanVisible` on `NotebookCell` and in the `.tabularis-notebook` file format (`types/notebook.ts`, `utils/notebookFile.ts`), mirroring `isChartVisible`. Both the toolbar icon and the section's own collapse chevron flip the same flag.

### 2) Shared explain hooks (`useExplainPlan`, `useExplainAnalyze`)

The cell section, `VisualExplainModal` and `VisualExplainPage` each had their own copy of the `useState` + `invoke("explain_query_plan")` + `resolveExplainOutput` logic. That now lives once in `useExplainPlan`, which also guards against stale responses via `useLatestAsync` and normalises errors through `toErrorMessage`. Both existing surfaces were migrated onto it:

- `VisualExplainModal` gains an optional `viewState` prop so the cell popout can show an already-fetched plan with no second database round-trip (request controls are hidden in that mode). The initial request is deferred a microtask so StrictMode's setup/cleanup probe no longer fires it twice.
- `VisualExplainPage` gets source-version guards so a slow file picker, CLI handoff or deep-link request cannot overwrite a newer source. Two pre-existing bugs fixed on the way: a load error with no file/plan was hidden behind the "open a file" empty hint, and a rejected `openDialog` was an unhandled promise. Both now render the error.

### 3) Fix: notebook result export failed silently

`ResultToolbar` wrote CSV/JSON with no feedback and no error handling, so a failed write (permissions, cancelled mid-flight, disk error) just vanished. Both paths are collapsed into one `handleExport(format)` wrapped in try/catch, which logs the error and surfaces an alert: the new `resultExportSuccess` on success and the new `exportError` on failure.

### 4) Fix: regex `lastIndex` bugs in param / variable resolution

`hasParamReferences` and `hasCellReferences` called `.test()` on shared **global** (`/g`) regexes. `.test()` advances `lastIndex` on a global regex, so repeated calls on the same SQL alternated `true`/`false`. Both now test against a fresh, non-global copy built from `.source` (the same style already used by `extractParamReferences` / `extractCellReferences`).

Relatedly, `resolveQueryVariables` could emit duplicate CTEs (invalid SQL) when a cell was referenced more than once. It now emits a single CTE per referenced cell (tracked with a `seen` set) and replaces every placeholder occurrence in one pass with a replacer callback instead of one `.replace()` per ref. Per-result CTE serialization is cached in a `WeakMap` keyed by the `QueryResult` object, so re-rendering a plan-visible cell does not re-serialize unchanged results.

### 5) Fix: notebook cells remounted on reorder / insert / delete

`NotebookView` keyed each cell as `${cell.id}-${index}`, so every cell below an insert, delete or drag-reorder point got a new key and was unmounted and recreated (losing editor cursor/undo state, result scroll position and the plan panel). The key is now `cell.id`, which is already what the cell ref map uses. The `escapeBackslashes` option passed to `resolveQueryVariables` is now memoized and also applies to `mariadb`, not only `mysql`.

### 6) Minor

- `NotebookOutline` consolidated two imports from `notebookOutline` into one line.

### i18n

Added `resultExportSuccess`, `sectionQueryPlan`, `queryPlanOutdated`, `queryPlanUnresolved`, `explainPopout`, `toggleQueryPlan`, `hideQueryPlan` and `exportError` under `editor.notebook` in all 11 locales.

## Testing

- `pnpm typecheck` — clean.
- `pnpm lint` — 0 errors (only the pre-existing `ThemeProvider.tsx` warning, untouched).
- All 11 locale JSONs parse; the new keys are present in each (`tests/i18n/notebookQueryPlan.test.ts`).
- `pnpm vitest run` — 263 files, 4345 tests passing. New suites: `SqlCellExplain`, `SqlCell`, `NotebookCellWrapper`, `NotebookView`, `ResultToolbar`, `VisualExplainModal`, `VisualExplainPage`, `useExplainPlan`, `useExplainAnalyze`. Extended: `notebookParams`, `notebookVariables`, `notebookFile` (round-trip with `isQueryPlanVisible`, v1 import). A manual fixture lives at `tests/fixtures/notebook-query-plan.tabularis-notebook`.
image

### Manual

Import `tests/fixtures/notebook-query-plan.tabularis-notebook` against a connection whose driver supports EXPLAIN.

1. **Toggle + persistence** — click the `Network` icon on a SQL cell; the Query Plan section opens and the icon turns blue. Collapse via the section chevron; icon resets. Save, reload — state is restored. The icon is absent on a driver without EXPLAIN support.
2. **Resolved query** — run cell 1, open the plan on cell 2. The plan is for the CTE-expanded query with `@min_id` substituted and a single `cell_1` CTE despite `{{cell_1}}` and `{{cell_01}}` both being referenced.
3. **Unresolved refs** — open cell 2's plan before running cell 1: `queryPlanUnresolved` lists the refs, no request is sent. Cell 3 (`{{cell_99}}`) stays in that state.
4. **Outdated plan** — with a plan visible, change `min_id` to `2` or edit the SQL: the plan is replaced by `queryPlanOutdated` until Re-run.
5. **No re-run on execute** — press Run on the cell repeatedly; the plan panel does not refetch.
6. **DML safety** — `UPDATE t SET x = x WHERE 1 = 0`: Analyze is off by default, plain EXPLAIN runs. Ticking Analyze shows the warning and executes nothing until Re-run. Editing the query resets Analyze.
7. **Popout** — after a successful plan, open the modal: same plan, graph/tree switch works, no Analyze/Re-run controls, no extra request.
8. **Reorder keeps state** — type in a cell and open its plan, then drag/insert/delete cells above it: editor state and plan panel are intact.
9. **Result export** — CSV and JSON succeed with an alert; cancelling the dialog shows nothing; a read-only target shows `exportError`.

https://github.com/user-attachments/assets/908bf4d3-41b4-4ce4-9353-aa1c2b9b2d7f


debba (changes requested):
Thanks for this one, I like where it takes the notebook. Having the plan as a proper section next to Results and Chart feels much more natural than the little button in the AI overlay, and pulling the explain state into `useExplainPlan` was overdue. I checked out the branch and went through it properly, here is what I found.

## What I verified

The three fixes are real. I ran the main and branch versions of the helpers side by side:

- `hasParamReferences` on main, same SQL four times: `true, false, true, false`. On the branch: `true` every time. Same story for `hasCellReferences`.
- A query referencing `{{cell_2}}` twice on main produces two `cell_2` CTEs. On the branch it produces one and both placeholders get replaced.
- I wrote a notebook file with `isQueryPlanVisible` on two cells and ran it through validate, deserialize and serialize again. The flags survive and the cells without the flag stay clean.

Typecheck is clean, eslint on the changed files is clean, the 101 tests you listed pass, the branch merges into main without conflicts, and the new keys are in all 11 locales.

I also tested it against a local Postgres with a 20k row table. The inline section, the Analyze toggle, the DML warning, the popout and the persistence all behave as described.

## Things I would like to see before merging

**Regression tests.** The description says the 101 tests cover the `lastIndex` fix, the CTE dedup and the file round trip, but no test file changed in this PR and none of the existing tests exercise those paths (nothing calls the helpers twice, nothing references a cell twice, nothing mentions `isQueryPlanVisible`). Four small tests would close this: repeated `hasParamReferences`, repeated `hasCellReferences`, duplicate cell reference giving a single CTE, and a round trip with the new flag. Our testing rules ask for edge cases on utils anyway, and these are exactly the cases that bit you.

**Analyze on a query that turns into DML.** `analyze` is initialised from `isDml` only when the section mounts. If I open the plan on a SELECT and then rewrite the cell into an UPDATE, the checkbox stays ticked and the next re-run executes the write. I confirmed on Postgres that `EXPLAIN ANALYZE UPDATE` really updates the rows. The modal has the same pattern, but a modal lives for a few seconds while this section stays mounted for the life of the notebook, so the window is much bigger. I would force the checkbox off whenever the query becomes data modifying, and maybe only show the warning path when the user explicitly turns it back on.

## Smaller notes

- The effect in `SqlCellExplain` calls `runExplain`, which sets three states synchronously before the await. Our React rules frown on that, although the modal already did the same, so I will not hold the PR on it.
- The default height in the code is 720, the description says 460. Not important, just noting it since I went looking.
- The export feedback is an alert modal, not a toast. That is consistent with how notebook export already reports, so fine, but the description is a bit off.
- The toolbar tooltip always says "Show query plan" even while it is showing. A "Hide" variant would be nice.
- Once the query is edited the old plan stays with no hint that it is stale. The on-demand choice is right, a small "outdated" marker would make it obvious though.
- `hasParamReferences` and `hasCellReferences` have no callers anywhere in `src` at the moment, so the `lastIndex` bug never surfaced in the app. Still worth fixing, just so you know the blast radius is zero.

Overall this is good work and I am happy to approve once the tests are in. The DML thing I would really like fixed too, but I am fine if you prefer to do it in a follow up as long as it does not get lost.

PR #719, opened 2026-09-05
fix(test): bind storage globals to the jsdom environment

### LINKED ISSUE #713: Flaky localStorage.clear() TypeError in SettingsProvider/ThemeProvider/useSidebarResize tests
## Summary

Three test files intermittently fail their entire suite with the same setup-time crash:

```
TypeError: Cannot read properties of undefined (reading 'clear')
 ❯ tests/contexts/SettingsProvider.test.tsx:35:18
     localStorage.clear();
```

Affected files (all fail identically, at their own `beforeEach`'s `localStorage.clear()` call):

- `tests/contexts/SettingsProvider.test.tsx`
- `tests/contexts/ThemeProvider.test.tsx`
- `tests/hooks/useSidebarResize.test.ts`

`localStorage` itself is `undefined` at that point — not "exists but throws," the global binding hasn't been installed yet.

## Why this looks like a harness race, not a code bug

- The failure is nondeterministic: running a file alone vs. alongside other files changes whether it reproduces, and the specific set of "other files" doesn't matter in any way that points to test pollution — it looks like which Vitest worker/thread a file lands on, and when in that worker's lifecycle it runs, decides the outcome.
- When it fails, it fails **every test in the file** at the shared `beforeEach`, before any test body executes — ruling out anything test-specific.
- A companion warning shows up whenever the crash happens:
  ```
  ExperimentalWarning: localStorage is not available because --localstorage-file was not provided.
  ```
  That's Node's own native `localStorage` global (unrelated to jsdom's polyfill) — suggests two different `localStorage` providers (Node's built-in implementation and jsdom's own polyfill) are racing to attach to `globalThis`, and whichever loses that race in a given worker leaves the binding `undefined` for that run.
- Confirmed pre-existing: reproduced identically on `main` before any of the changes in `postgres-plugin-force-install`, and on multiple commits since — the failure rate/behavior doesn't correlate with any specific code change.

## Impact

Currently just test flakiness (no product bug), but it's a risk multiplier: anyone touching `SettingsProvider`, `ThemeProvider`, or `useSidebarResize` will see red locally or in CI on an unrelated change and either (a) waste time chasing a phantom regression, or (b) start ignoring red in these files out of habit — which is how a real regression eventually slips through unnoticed.

## Suggested next steps (not investigated in depth — flagging, not diagnosing)

- Check whether `vitest.config.ts`'s `environment: 'jsdom'` setup interacts with Node's experimental native `localStorage` (stable since Node 22) — possibly a `--no-experimental-global-webstorage` flag, an explicit jsdom `localStorage` shim in `tests/setup.ts`, or pinning/upgrading whichever of `vitest`/`jsdom` owns this initialization would close the race.
- Try `pool: 'threads'` vs `'forks'` in `vitest.config.ts` (if not already pinned) — the race is consistent with per-worker jsdom bootstrap timing, and pool strategy affects that.
- Once a fix lands, run the affected 3 files in isolation and together, several times, to confirm the race is actually closed rather than just less likely.

## Reproduction

```bash
pnpm vitest run tests/contexts/SettingsProvider.test.tsx
# sometimes passes, sometimes fails — rerun a few times to see both outcomes
```

be-student:
## Summary
Fixes #713.

Node's native Web Storage globals can keep Vitest from installing jsdom's storage. On the unchanged checkout under Node 26.0.0, all 40 tests in the three reported files fail at `localStorage.clear()`.

Bind `localStorage` and `sessionStorage` to the current worker's actual `jsdom.window` in shared setup, before test modules load. This preserves real browser Storage behavior and avoids per-file mocks or a shared Node storage file. Four regression cases verify both APIs at module evaluation, separate stores, and restoration after global stubs. The testing rules document the setup invariant.

## Validation
- Before: 40/40 reported tests failed; all four new regression cases failed.
- After, Node 26.0.0: full frontend suite passed, **244 files / 3,973 tests**.
- Pinned Node 24.18.0: all **44 targeted tests** passed.
- `pnpm run typecheck` and `pnpm run build` passed.
- `pnpm run lint` passed with one existing `ThemeProvider.tsx:342` exhaustive-deps warning in an unchanged file.
- `git diff --cached --check` passed. GitNexus impact/detect-changes ran; top-level setup has no resolved call-graph edges, so validation covered the complete frontend suite instead of assuming no impact.

The Rust backend and native application runtime are unchanged and were not tested for this harness-only change.

## AI disclosure
Implemented and validated using OpenAI Codex, submitted by be-student. No human review has been performed.

### DIFF
--- a/.rules/testing.md
+++ b/.rules/testing.md
@@ -161,3 +161,12 @@ Tests are configured in `vitest.config.ts`:
 - Test files are discovered in `tests/` directory
 - Setup file: `./tests/setup.ts`
 - Environment: `jsdom` for DOM-related tests
+
+### Browser storage in the test environment
+
+`tests/setup.ts` binds `localStorage` and `sessionStorage` to the current
+Vitest jsdom window before test modules load. Keep this shared setup: Node's
+native storage globals can otherwise shadow jsdom, yielding unavailable or
+non-browser storage. Do not work around it with per-test storage shims or a
+shared `--localstorage-file`. `tests/storageEnvironment.test.ts` verifies the
+real DOM storage identity, separate stores, and restoration after global stubs.
--- a/tests/setup.ts
+++ b/tests/setup.ts
@@ -1,7 +1,18 @@
+/// <reference types="vitest/jsdom" />
 import "@testing-library/jest-dom/vitest";
 import { cleanup } from "@testing-library/react";
 import { afterEach } from "vitest";
 import { vi } from "vitest";
+
+// Node's native storage globals can prevent Vitest from installing jsdom's.
+// Bind both APIs before test modules load, using this worker's real DOM storage.
+for (const name of ["localStorage", "sessionStorage"] as const) {
+  Object.defineProperty(globalThis, name, {
+    configurable: true,
+    writable: true,
+    value: jsdom.window[name],
+  });
+}
 
 // Cleanup after each test
 afterEach(() => {
--- a/tests/storageEnvironment.test.ts
+++ b/tests/storageEnvironment.test.ts
@@ -1,0 +1,45 @@
+/// <reference types="vitest/jsdom" />
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+// Capture during module evaluation, before any test hooks can repair the globals.
+const initialStorage = {
+  localStorage: globalThis.localStorage,
+  sessionStorage: globalThis.sessionStorage,
+};
+
+describe("jsdom storage environment", () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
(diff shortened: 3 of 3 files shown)


debba (approved):
Verified locally: reproduced the 40 failures with the old setup on Node 26.8.1 and confirmed the root cause in Vitest's jsdom environment (keys already present on the Node global are skipped unless whitelisted, and the storage globals are not). With this branch the 44 targeted tests and the full suite (244 files / 3973 tests) pass on Node 26, and on Node 22/24 with and without `--experimental-webstorage`. Approving. Thanks for the clean fix and the regression file.

PR #720, opened 2026-09-06
fix(dump): escape JSON columns and backslashes per dialect

### LINKED ISSUE #717: [Bug]: MySQL Dump Database escapes JSON columns inconsistently — import fails or silently corrupts non-ASCII JSON
### Describe the bug

`Dump Database` on a MySQL connection writes JSON column values with a different escaping rule than text column values, so the resulting `.sql` file cannot be imported correctly under any `sql_mode` — and in the common case it imports **without error while silently corrupting data**.

Root cause is in `escape_sql_value()` (`src-tauri/src/dump_commands.rs`, lines 288–301 at tag `v0.22.0`), combined with the way the MySQL driver reads JSON columns (`src-tauri/src/drivers/mysql/extract/json.rs`, which parses JSON columns into `serde_json::Value` before they reach the dumper):

```rust
serde_json::Value::String(s) => format!("'{}'", s.replace("'", "''").replace("\\", "\\\\")), // Basic escaping
_ => format!("'{}'", val.to_string().replace("'", "''")), // Fallback
```

Three distinct symptoms follow from this:

1. **JSON objects/arrays** (`_ =>` fallback): backslashes are *not* doubled. MySQL's default `sql_mode` treats backslashes in string literals as escape sequences, so:
   - `\"` inside a JSON string becomes `"` → the value is no longer valid JSON → `ERROR 3140 Invalid JSON text`.
   - `\uXXXX` (what PHP/Laravel `json_encode` emits for every non-ASCII character by default) becomes `uXXXX` → **still valid JSON, import succeeds, all non-ASCII text is silently destroyed**. In my dump a single JSON column contained 59,873 `\uXXXX` sequences; none survive a round trip.
   - `\\` becomes `\` → invalid JSON.
2. **Text columns** (`Value::String` branch): backslashes *are* doubled, which is correct for MySQL's default mode. Because rule 1 and rule 2 coexist in the same file, `NO_BACKSLASH_ESCAPES` fixes the JSON columns but then double-escapes every text column instead.
3. **JSON string scalars** (a JSON column whose value is a JSON string, e.g. `"\"[151, 152, 153]\""`): `json.rs` returns `Value::String`, so the dumper writes the *string content* without the surrounding JSON quotes and with backslashes doubled. The result is not valid JSON under either `sql_mode`.

Related fidelity issues in the same function: JSON `true`/`false` are written as `1`/`0` (type changes on re-import), and JSON `null` is indistinguishable from SQL `NULL`.

The importer (`execute_statements_streaming`) does not set `sql_mode` or a client charset either, so there is no way to compensate on the import side.

### To Reproduce

On any MySQL 8.x database:

```sql
CREATE TABLE json_escape_repro (
  id INT PRIMARY KEY,
  t  TEXT,
  j  JSON
);
INSERT INTO json_escape_repro VALUES
  (1, 'App\\Jobs\\Foo',                       '{"class":"App\\\\Jobs\\\\Foo"}'),
  (2, 'x',                                    '{"html":"<a target=\\"_blank\\">x</a>"}'),
  (3, 'x',                                    '{"name":"\\u4e2d\\u6587"}'),
  (4, 'x',                                    '"\\"[151, 152, 153]\\""');
```

1. Right-click the database → **Dump Database** (Include Structure + Include Data).
2. Open the `.sql` file. Observe that column `t` in row 1 is written as `'App\\\\Jobs\\\\Foo'` (backslashes doubled) while column `j` in the same row is written as `'{"class":"App\\Jobs\\Foo"}'` (not doubled). Row 4's `j` is written as `'"\\"[151, 152, 153]\\""'`.
3. Import the file into an empty database (either via **Import Database** or `mysql < dump.sql`):
   - Row 2 fails with `ERROR 3140 (22032): Invalid JSON text`.
   - If you delete row 2 from the file, the import **succeeds**, but `SELECT j FROM json_escape_repro WHERE id = 3` returns `{"name": "u4e2du6587"}` — the Chinese text is gone with no error.
   - Row 4 fails under `NO_BACKSLASH_ESCAPES`, and under the default mode is re-imported one JSON nesting level shallower than the original.

Expected: the dump round-trips byte-for-byte for JSON columns, the same way `mysqldump` does.

### Suggested fix

In `escape_sql_value`:

- For the `_` fallback (objects/arrays), apply the same escaping as the `String` branch: `val.to_string().replace("\\", "\\\\").replace("'", "''")`.
- For JSON string scalars coming from a JSON column, emit the JSON literal (`val.to_string()`, i.e. with the surrounding quotes) and then escape it, rather than the unquoted content. This needs the caller to know the column is JSON-typed, since `Value::String` is also used for ordinary text columns.
- When the column is JSON-typed, write JSON booleans as `true`/`false` instead of `1`/`0`, so the value type survives the round trip (keep `1`/`0` for non-JSON columns).

Alternatively, for MySQL, read JSON columns as raw text in the dump path (skip the `serde_json::from_str` parse in `json.rs`) so every value goes through the `String` branch with one consistent rule.

On the import side, it would also help to `SET NAMES utf8mb4` and to wrap the import in a transaction (or at least document that it is not atomic).

### OS Version

macOS 26.x (Darwin 25.6.0), Apple Silicon

### Tabularis Version

0.22.0 (also present on `main` at the time of writing)

### Relevant Log Output

```
ERROR 3140 (22032) at line 9651: Invalid JSON text: "Missing a comma or '}' after an object member." at position 113 in value for column 'order_logs.data'.
ERROR 3140 (22032) at line 1090: Invalid JSON text: "The document root must not be followed by other values." at position 4 in value for column 'moto_rental_order.moto_rental_schedule_ids'.
```

(Errors from importing a Tabularis-generated dump of a Laravel application's MySQL 8.4 database with `mysql` CLI; the first is symptom 1, the second is symptom 3.)

soliloquist:
Fixes #717

## What was wrong

`escape_sql_value()` in `dump_commands.rs` applied two different escaping rules inside the same dump file:

- `serde_json::Value::String` (text columns) doubled backslashes — correct for MySQL, but wrong for PostgreSQL (`standard_conforming_strings = on`) and SQLite, where a backslash is an ordinary character and doubling it re-imports as two backslashes.
- Everything else (the `_` fallback) only escaped single quotes. JSON-typed columns are extracted as parsed `serde_json::Value`s, so every JSON object/array went through this branch with its backslashes intact. On MySQL that means `\"` becomes `"` (invalid JSON → `ERROR 3140`), and `\uXXXX` becomes `uXXXX` — **valid JSON, import succeeds, all non-ASCII text silently destroyed**.
- A JSON column whose value is a JSON *string* arrived as `Value::String`, so the dumper wrote the string's content without its surrounding quotes (and with doubled backslashes), which is not valid JSON under any `sql_mode`.

## What this PR does

- Adds `escape_sql_string(driver, s)`: one dialect-aware rule for quoting string literals. MySQL doubles backslashes, escapes `'` as `''`, and writes NUL as `\0`; PostgreSQL and SQLite only escape `'`.
- `escape_sql_value(driver, val)` now routes strings *and* nested values through that helper, so a file never mixes two rules.
- Adds `escape_json_column_value(driver, val)` and uses it for columns whose type is `JSON` (MySQL, via `type_info().name()`) or `json`/`jsonb` (PostgreSQL, via `Column::type_()`). Values are written as JSON literals (`serde_json::Value::to_string()`), so JSON strings keep their quotes, booleans stay `true`/`false`, and non-ASCII is emitted as UTF-8 rather than `\uXXXX`. The result is then escaped like any other string literal for the dialect.
- Unit tests cover backslash handling per dialect, NUL, nested values, JSON string scalars, JSON booleans/numbers/null, and UTF-8 content, for both MySQL and PostgreSQL.

Behaviour change for PostgreSQL/SQLite text columns: backslashes are no longer doubled. The previous output re-imported `a\b` as `a\\b`; the new output round-trips.

Not changed in this PR (happy to follow up separately): the importer still runs statements outside a transaction and does not set `SET NAMES utf8mb4` on MySQL, and SQL `NULL` vs JSON `null` in a JSON column are still written identically as `NULL` because they are indistinguishable once extracted.

## Verification

Reproduction from #717 (MySQL 8.4):

```sql
CREATE TABLE json_escape_repro (id INT PRIMARY KEY, t TEXT, j JSON);
INSERT INTO json_escape_repro VALUES
  (1, 'App\\Jobs\\Foo', '{"class":"App\\\\Jobs\\\\Foo"}'),
  (2, 'x',              '{"html":"<a target=\\"_blank\\">x</a>"}'),
  (3, 'x',              '{"name":"\\u4e2d\\u6587"}'),
  (4, 'x',              '"\\"[151, 152, 153]\\""');
```

Before (v0.22.0): the dump of this table fails to import on row 2 (`ERROR 3140`), and with row 2 removed imports row 3 as `{"name": "u4e2du6587"}`.

After: I took the literals the new functions emit for these rows (the exact strings asserted in the unit tests) and re-imported them into a second table on MySQL 8.0; `a.j = b.j` and `JSON_TYPE()` match the originals for all rows, including the JSON-string and boolean cases. I have not yet exercised the full Tauri dump → import UI flow with this build, so a check on a real MySQL/PostgreSQL connection from the app would be welcome.

`cd src-tauri && cargo test --lib dump_commands` passes (existing test updated for the new signature, five new tests added).

### DIFF
--- a/src-tauri/src/dump_commands.rs
+++ b/src-tauri/src/dump_commands.rs
@@ -171,6 +171,7 @@ async fn export_table_data(
         "mysql" => {
             use crate::drivers::mysql::extract::extract_value;
             use crate::pool_manager::get_mysql_pool; // Returns String (JSON value or "NULL")
+            use sqlx::{Column, TypeInfo};
 
             let pool = get_mysql_pool(params).await?;
             let mut rows = sqlx::query(&query).fetch(&pool);
@@ -180,7 +181,14 @@ async fn export_table_data(
                 let mut values = Vec::new();
                 for i in 0..row.columns().len() {
                     let val = extract_value(&row, i, None);
-                    values.push(escape_sql_value(val));
+                    // JSON columns are extracted as parsed values and must be
+                    // written back as JSON literals, not as their string content.
+                    let is_json_column = row.column(i).type_info().name() == "JSON";
+                    values.push(if is_json_column {
+                        escape_json_column_value(driver, val)
+                    } else {
+                        escape_sql_value(driver, val)
+                    });
                 }
                 batch.push(format!("({})", values.join(", ")));
 
@@ -221,7 +229,14 @@ async fn export_table_data(
                 let mut values = Vec::new();
                 for i in 0..row.columns().len() {
                     let val = extract_value(&row, i, None);
-                    values.push(escape_sql_value(val));
+                    let column_type = row.columns()[i].type_();
+                    let is_json_column = *column_type == tokio_postgres::types::Type::JSON
+                        || *column_type == tokio_postgres::types::Type::JSONB;
+                    values.push(if is_json_column {
+                        escape_json_column_value(driver, val)
+                    } else {
+                        escape_sql_value(driver, val)
+                    });
                 }
                 batch.push(format!("({})", values.join(", ")));
 
@@ -256,7 +271,7 @@ async fn export_table_data(
                 let mut values = Vec::new();
                 for i in 0..row.columns().len() {
                     let val = extract_value(&row, i, None);
-                    values.push(escape_sql_value(val));
+                    values.push(escape_sql_value(driver, val));
                 }
                 batch.push(format!("({})", values.join(", ")));
 
@@ -285,7 +300,29 @@ async fn export_table_data(
     Ok(())
 }
 
-fn escape_sql_value(val: serde_json::Value) -> String {
+/// Quotes `s` as a string literal for the target dialect.
+///
+/// MySQL treats backslashes inside string literals as escape sequences
+/// (unless `NO_BACKSLASH_ESCAPES` is set), so a literal backslash must be
+/// doubled and a NUL byte written as `\0` — otherwise a `\u4e2d` escape
+/// inside JSON silently re-imports as `u4e2d` and `\"` breaks the JSON
+/// altogether. PostgreSQL (with the default
+/// `standard_conforming_strings = on`) and SQLite take backslashes
+/// literally, so doubling them there would corrupt the data instead; only
+/// the single quote needs escaping.
+fn escape_sql_string(driver: &str, s: &str) -> String {
+    match driver {
+        "mysql" => format!(
+            "'{}'",
+            s.replace('\\', "\\\\")
+                .replace('\'', "''")
+                .replace('\0', "\\0")
+        ),
+        _ => format!("'{}'", s.replace('\'', "''")),
+    }
+}
+
+fn escape_sql_value(driver: &str, val: serde_json::Value) -> String {
     match val {
         serde_json::Value::Null => "NULL".to_string(),
         serde_json::Value::Number(n) => n.to_string(),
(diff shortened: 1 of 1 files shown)


debba (approved):
Thanks for this, the fix looks solid to me. I verified it locally against MySQL 8.4 and PostgreSQL 16 with the rows from #717 plus a few extra edge cases (quotes, NUL, newlines, JSON string scalars, booleans, raw UTF-8): everything round-trips through dump and import on both dialects, and the unit tests pass on the branch.

I would be glad if you want to keep contributing. If you like, feel free to join our Discord server as well: https://discord.com/invite/K2hmhfHRSt

PR #723, opened 2026-09-06
fix(theme): follow native system appearance

### LINKED ISSUE #716: [Bug]: "Follow System" theme mode applies Light even when the OS is in Dark mode
### Describe the bug

With **Follow System** enabled in Settings → Appearance, the app applies the **Light** theme even though the OS is set to Dark. This is worse than the behavior reported in #649: instead of just staying on a fixed theme, Follow System now actively forces the wrong one.

I commented on #649 (opened by @iamthenuggetman, closed by their own #650) confirming the native window chrome (CSD) issue on Fedora/GNOME. That part — `getCurrentWindow().setTheme(...)` syncing the titlebar — works. But the underlying "is the OS in dark mode" detection is still wrong.

Confirmed directly: with the OS in Dark, opening the WebKit inspector (right-click → Inspect Element) and running in the console:

```js
window.matchMedia('(prefers-color-scheme: dark)').matches
```

Returns `false`. The app's Follow System logic relies entirely on this media query, so it picks the Light theme regardless of the real OS setting.

**Screenshot attached:** Tabularis (Follow System on) renders Light, while the native GTK Files app next to it renders Dark — confirming the OS itself is in Dark mode.

Image

### To Reproduce

1. Set the OS appearance to Dark (GNOME, Fedora 44, Wayland)
2. Run `tabularis --debug` (this opens the WebKit inspector)
3. In Settings → Appearance, enable **Follow System**
4. Observe: the app switches to a **Light** theme instead of Dark
5. In the inspector console, run `window.matchMedia('(prefers-color-scheme: dark)').matches` → returns `false`

### OS Version

Fedora Linux 44 (GNOME, Wayland), kernel 7.1.13-200.fc44.x86_64

### Tabularis Version

v0.22.0

### Relevant Log Output

```shell
No backend errors are logged for this; the theme-detection logic lives entirely in the frontend (webview) and fails silently. `tabularis --debug` was used to open the WebKit inspector and confirm the `matchMedia` value above.
```

be-student:
## Problem

On Linux WebKit, `prefers-color-scheme` can report light even when the desktop is using a dark theme. As a result, **Follow System** could select and retain the wrong Tabularis theme on Fedora/GNOME.

## Changes

- Resolve the current system appearance through Tauri's native `window.theme()` API.
- Subscribe to native `onThemeChanged()` events while Follow System is enabled.
- Pass `null` to `window.setTheme()` so the native window continues following the OS.
- Keep `prefers-color-scheme` as the browser-preview fallback when the native API is unavailable.
- Delay initial theme application until configuration loading completes, preventing the default theme from overriding the detected native theme.
- Add regression coverage for a native-dark/WebKit-light disagreement and for the browser fallback.

## Validation

- Regression test failed before the implementation and passes afterward.
- `NODE_OPTIONS=--no-experimental-webstorage pnpm test`: 243 files, 3,971 tests passed.
- `pnpm typecheck`: passed.
- `pnpm lint`: passed.
- `pnpm build`: passed.
- `git diff --check`: passed.
- GitNexus `detect-changes`: medium impact, limited to the existing theme-setting flows.

Closes #716

### DIFF
--- a/.rules/react.md
+++ b/.rules/react.md
@@ -3,3 +3,4 @@
 2. **Sync State in Effects:** NEVER call `setState` synchronously inside `useEffect`. This triggers unnecessary cascading renders. Use `useMemo` for derived state or initialize state directly if possible.
 3. **Fast Refresh Compatibility:** Files exporting React components (especially Contexts) must NOT export other constants or helper functions. Move helpers to separate utility files.
 4. **Library Safety:** Be aware of incompatible libraries with the React Compiler (e.g., `useReactTable`). Do not wrap their return values in `useMemo` if the library manages its own internal memoization or returns unstable function references.
+5. **Native System Theme:** In the Tauri runtime, resolve Follow System through `getCurrentWindow().theme()` and `onThemeChanged()`, and call `setTheme(null)` so the native window keeps following the OS. Use `prefers-color-scheme` only as the browser-preview fallback because WebKit may report the wrong Linux desktop theme.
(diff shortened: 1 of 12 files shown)


debba (changes requested):
Requesting changes only for the three small cleanups listed in my comment above (redundant `#[path]` attribute, dead media-query helpers in `themeManagement.ts`, D-Bus connection reuse). Everything else is verified and ready to merge.

debba (approved):
Verified live on GNOME/Wayland, all tests green.

PR #737, opened 2026-09-10
feat: configurable HTTP/SOCKS5 proxy with pluggable traffic scopes

coloraven:
## Summary

Adds configurable HTTP/SOCKS5 proxy support with a pluggable traffic-scope framework.

### Global settings
A new **Network** settings tab configures a shared proxy (protocol, host, port, optional username/password stored in the OS keychain) plus opt-in checkboxes for traffic scopes (extensible for future categories):

1. **App network requests** — plugin downloads, registry/JWKS fetches, update checks, WebDAV backup uploads
2. **Database connections** — direct TCP to databases
3. **AI / LLM endpoints** — provider HTTP calls
4. **SSH tunnels** — outbound connection to the SSH bastion

### Per-target overrides
- **Database connection** — override when creating/editing a connection (Advanced)
- **AI provider** — override per selected provider in the AI settings tab

### Precedence
Global settings have the **lowest** priority. A per-connection or per-provider setting can **inherit**, set a **custom** proxy, or **disable** proxying for that target.

### Supported proxy types
- HTTP (CONNECT)
- SOCKS5
- Optional username/password authentication

### Wiring notes
- `app_http`: updater, plugin installer/compat/integrity, WebDAV, legacy `raw.githubusercontent.com` registry.json, and Tabularium SDK (reqwest 0.12 client injected with the same proxied `app_http` client)
- `ai`: all `ai.rs` HTTP via `ai_http_client(provider)`
- `ssh_tunnel`: local TCP forward → SSH (`new_with_tcp_override`)
- `database`: TCP forward in `resolve_connection_params`; localhost after SSH/K8s is not re-proxied

## Test plan
- [ ] Enable global proxy + `app_http` scope; confirm update check / plugin download works through the proxy
- [ ] Enable `ai` scope; confirm LLM calls work; set a provider override to custom/disabled and confirm it wins over global
- [ ] Create a DB connection with custom proxy (no SSH); confirm connect works through the proxy
- [ ] Create a DB connection with SSH + proxy; confirm proxy applies to the bastion hop and the local tunnel leg is not double-proxied
- [ ] Disable proxy at connection level while global DB scope is on; confirm direct connect bypasses the proxy
- [ ] Confirm proxy passwords are stored in the keychain and never written to `config.json` / `connections.json`

### Global Setting

image

### Model endpoint
image

### database connection
image

### SSH connection

Reusable SSH connection profiles (Settings → SSH Connections) do **not** have their own proxy override UI in this PR. SSH traffic can still use a proxy via:
1. the global Network proxy with the **SSH tunnels** scope enabled, or
2. the **database connection** Advanced proxy override (when that connection uses SSH).
A dedicated per–SSH-profile proxy override can be added in a follow-up if needed.

### Also included
- Empty connection port fields now map to `undefined` instead of being saved as port `0`.


debba (changes requested):
Hi @coloraven, thanks a lot for this one. Proxy support has been requested for a while and the way you structured it (global scopes plus per-target overrides, passwords only in the keychain) is exactly how I would have wanted it. I went through the whole diff, ran everything locally and also drove the new TCP forwarder and the reqwest wiring against real HTTP CONNECT and SOCKS5 proxies, with and without authentication. It all works as described: precedence is right, bad credentials are rejected, HTTPS goes through both proxy types, and nothing proxy-related ends up in `config.json` or `connections.json`. Nice work.

There is one thing I need fixed before merging, and a handful of notes you can take or leave.

### Blocker

`cargo test` does not compile on this branch. The test helper `test_connection_params()` in `src-tauri/src/plugins/driver.rs` (around line 1358) builds a `ConnectionParams` literal without the new `proxy` field, so the test build fails with E0063. Adding `proxy: None,` (or `..Default::default()`) fixes it. CI will be red until then.

### Notes, non-blocking

1. **Forward cache key ignores protocol and credentials.** `build_forward_key` in `proxy/tcp_forward.rs` uses proxy host:port plus target only. If I change the password, username or protocol while keeping the same proxy host, the old forward keeps being reused until the app restarts. I reproduced this: a forward created with a wrong password still fails after fixing the password. Including protocol and username in the key, or dropping forwards on config save, would solve it.

2. **Idle connections through the forward are cut after 5 minutes.** `IO_TIMEOUT` puts a 300 s read timeout on both legs, so an idle pooled DB connection hits the timeout, `io::copy` errors and both sockets get shut down. The SSH tunnel does not do this. I would drop the read timeout once the proxy handshake is done and keep the timeout only for the CONNECT phase.

3. **SOCKS5 name resolution is inconsistent between the two paths.** reqwest gets a `socks5://` URL, which resolves hostnames locally (my proxy log shows `CONNECT 140.82.121.5:443`, ATYP 1), while the TCP forwarder sends the domain name to the proxy (ATYP 3). In a network where DNS only works through the proxy, DB connections would work and update checks would not. `socks5h://` for the reqwest side would make them behave the same.

4. **Locales.** Only `en.json` and `zh.json` got the new keys. We normally add new keys to all 11 locale files (English text is fine as a placeholder, we translate afterwards). Nothing breaks since i18next falls back to English, but it keeps the files in sync.

5. **Rust conventions from `.rules/rust.md`.** Tests are inline in `resolve.rs`; we keep non-trivial tests in a sibling `tests.rs`. The forwarder's pure helpers (CONNECT response parsing, SOCKS5 reply parsing, `base64_encode`) have no unit tests, and `base64_encode` could just use the `base64` crate which is already a dependency.

6. **Small things.** The port field change in `NewConnectionModal` (empty field now maps to `undefined` instead of being saved as port 0) is a good fix but is unrelated to the PR and worth a line in the description. `DatabaseContext.ts` has a mis-indented `params: {` line.

Once the blocker is in I am happy to merge; the rest can be follow-ups if you prefer. Thanks again, this is a solid feature.

debba (approved):
Hi @coloraven, thank you for the contribution and for following up on the review feedback! The blocking test compilation issue is fixed and CI is green. This is a valuable addition to Tabularis. I'm going to merge this now. Thanks again!

PR #754, opened 2026-09-13
fix(linux): repair desktop entry — %u for deep links, Categories, and capitalized Name

### LINKED ISSUE #671: fix(deb): broken desktop entry (missing %u for deep links, empty Categories, lowercase Name)
The desktop entry shipped inside the `.deb` (`usr/share/applications/tabularis.desktop`) has several problems, surfaced during the review of #670 (which fixed the snap side only):

```ini
Categories=
Comment=tabularis - A database manager for developers
Exec=tabularis
Icon=tabularis
Name=tabularis
```

## 1. Deep links broken on cold start (actual bug)

The entry declares `MimeType=x-scheme-handler/tabularis` but `Exec=` has no `%u` field code, so when a `tabularis://` link is opened while the app is not running, the app is launched without the URL and the deep-link handler never receives it.

Possible routes:
- check whether a newer Tauri version adds `%u` automatically when the deep-link plugin is configured;
- otherwise provide a custom `desktopTemplate` under `bundle > linux > deb` in `src-tauri/tauri.conf.json`, which would also allow fixing the points below in one go.

## 2. Empty `Categories=`

`bundle.category` is unset in `src-tauri/tauri.conf.json`, so the app files under no menu category. Setting e.g. `"category": "DeveloperTool"` (mapped to `Development` on Linux) fixes .deb and AppImage, and also sets the App Store category on macOS.

## 3. Lowercase `Name=tabularis`

The menu entry shows a lowercase name because `Name=` follows `productName`. `productName` itself stays lowercase (binary name, snapcraft paths, WM class and updater artifacts all depend on it), so the fix is to set `Name=Tabularis` in the custom `desktopTemplate` from point 1, the same way #670 did for the snap entry.

Point 1 is a user-facing bug today; 2 and 3 are metadata polish. Contributions welcome on any of these.

anandghegde:
Closes #671.

## The deep-link bug

The bundled desktop entry declares `MimeType=x-scheme-handler/tabularis` (derived from the `deep-link` plugin scheme in `tauri.conf.json`) but `Exec=tabularis` has no field code. Per the [freedesktop Desktop Entry spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s07.html), a handler only receives the URL if `Exec=` carries `%u`/`%U`. So opening a `tabularis://` link while the app is closed launches the app with no URL, and the deep-link handler never fires.

The issue asked whether a newer Tauri adds `%u` automatically when the deep-link plugin is configured. **It does not** — I checked the bundler template on `dev`, and it is still:

```
Exec={{exec}}
StartupWMClass={{exec}}
```

([`crates/tauri-bundler/src/bundle/linux/freedesktop/main.desktop`](https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-bundler/src/bundle/linux/freedesktop/main.desktop)). There is no `%u` and no branch on `mime_type`, so a custom `desktopTemplate` is the only route. That also lets points 2 and 3 be fixed in one go, exactly as the issue suggested.

## Changes

**New `src-tauri/linux/tabularis.desktop`** — Tauri's stock template plus three fixes:

| | Before | After |
|---|---|---|
| Deep links | `Exec=tabularis` | `Exec=tabularis %u` |
| Menu category | `Categories=` (empty) | `Categories=Development;` |
| Menu name | `Name=tabularis` | `Name=Tabularis` |

**`src-tauri/tauri.conf.json`** — sets `bundle.category: "DeveloperTool"` and points `deb` and `rpm` at the template.

### Notes on the approach

- **`productName` is untouched.** It stays lowercase, so the binary name, snapcraft paths, updater artifacts and WM class are all unaffected. `Name=` is pinned literally in the template instead — the same approach #670 took for the snap's `title:`.
- **`StartupWMClass` deliberately keeps bare `{{exec}}`**, with no `%u`. It must continue to match the actual WM class or window-to-launcher grouping breaks.
- **`DeveloperTool`** maps to `Development;` on freedesktop and `public.app-category.developer-tools` on macOS ([`category.rs`](https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-bundler/src/bundle/category.rs)), so this also fills in the previously-unset App Store category.
- **Scope note — rpm included.** The issue is titled `fix(deb)`, but releases ship an rpm too and it has the identical bug, so I wired the same template into `bundle.linux.rpm`. It is one line referencing the same file; happy to drop it if you'd rather keep this deb-only.
- **AppImage needs no entry of its own** — its bundler calls `debian::generate_data`, so it picks up the deb template automatically. That matches the issue's note that setting the category fixes both .deb and AppImage.

## Verification

Checks run on this change:

1. **Rendered the template through Handlebars** with the exact context `tauri-bundler` supplies (`categories`, `comment`, `exec`, `icon`, `name`, `mime_type`), producing:

```ini
[Desktop Entry]
Categories=Development;
Comment=tabularis - A database manager for developers
Exec=tabularis %u
StartupWMClass=tabularis
Icon=tabularis
Name=Tabularis
Terminal=false
Type=Application
MimeType=x-scheme-handler/tabularis
```

2. **Parsed that output as a freedesktop key/value file** and asserted `Type`, `Exec` ends in ` %u`, `Categories` is `;`-terminated, `Name` is capitalized, `StartupWMClass` stays lowercase, and `MimeType` is intact.
3. **Validated `tauri.conf.json` against the official schema for the pinned `tauri 2.10.2`** — passes. `BundleConfig`, `DebConfig` and `RpmConfig` all set `additionalProperties: false`, so this confirms `category` and `desktopTemplate` are genuinely recognized keys rather than silently-ignored ones.

If you add the `preview` label, I'm glad to help confirm on the resulting build that `tabularis://` opens correctly from a closed state.


debba:
Thanks a lot @anandghegde for this fix! Great write-up too: the check against the bundler template on `dev` and the Handlebars render saved us a lot of time on the review.

I reproduced both the bug and the fix locally by launching the app through a desktop entry with and without `%u` via `gio launch`: without it argv is empty and the handler never fires, with your template the `tabularis://install/...` URL arrives and the plugin install event is emitted. Merging now.

You are very welcome to keep contributing, and if you feel like it, join us on our Discord server: https://discord.com/invite/K2hmhfHRSt

PR #755, opened 2026-09-13
fix: sanitize quoted local file paths for DB connections

coloraven:
## Summary
- Normalize pasted local filesystem paths before open/exists checks (strip wrapping quotes, curly quotes, `file://`, BOM/ZWSP).
- Shared helpers on both Rust (`fs_path`) and TypeScript (`fsPath`), wired through New Connection and connection-string parsing for SQLite and other file/folder drivers.

## Test plan
- [ ] Paste a quoted Windows path (e.g. `"C:\data\db.sqlite"`) into New Connection and confirm it connects
- [ ] Paste a `file:///...` URL and confirm it resolves to a usable path
- [ ] Confirm unquoted normal paths are unchanged
- [ ] Run `tests/utils/fsPath.test.ts` / related Rust `fs_path` unit coverage in CI


debba (changes requested):
Thanks for this, the quote/BOM peeling is a nice usability fix. A few things to address before merge:

**Blocking**

1. **Merge conflicts** – the branch conflicts with `main` in `src-tauri/src/commands.rs` (see my earlier comment, it depends on #737 which is now merged). Please rebase.

2. **`file://` handling is incomplete** in both `strip_file_uri_prefix` (`src-tauri/src/fs_path.rs`) and `stripFileUriPrefix` (`src/utils/fsPath.ts`):
   - `file://localhost/home/u/a.db` becomes `/localhost/home/u/a.db`: the authority is folded into the path, even though the doc comment lists this form as supported. Same for UNC-style `file://server/share/a.db`.
   - Percent-encoding is never decoded: `file:///home/u/my%20db.db` stays `/home/u/my%20db.db` and the `exists()` check fails.
   
   Rather than hand-rolling this, please use a real URL parser: the `url` crate is already a dependency (`Url::parse(...).to_file_path()`), and on the TS side `new URL(value)` + `decodeURIComponent(pathname)` (plus the drive-letter fix-up) covers the same cases. Please add tests for `file://localhost/...` and a percent-encoded path in both languages.

3. **Sanitization applied to every driver on the backend.** `resolve_connection_params` now runs `unwrap_database_selection_paths` on `database` for all connections, including MySQL/PostgreSQL/etc. where `database` is a schema name, not a path. The frontend correctly gates this on `file_based` / `folder_based` capabilities; the Rust side should do the same instead of rewriting database names for network drivers. It would also let you apply it once at the end of the function instead of in four separate return branches.

**Non-blocking**

4. `fs_path.rs` keeps its tests inline in `mod tests`. Per `.rules/rust.md` (rule 4), non-trivial tests go in a sibling `fs_path_tests.rs` loaded with `#[cfg(test)] mod fs_path_tests;`, like the other modules do.

5. `unwrap_quoted_path` / `unwrapQuotedPath` are introduced as "deprecated aliases" in brand-new files and have no callers. Please drop them (and the related test) rather than shipping dead code.

6. The `isLocalPathDriver` computation is duplicated in `NewConnectionModal.tsx` (test and save handlers). A single `useMemo` on `activeDriver` would do.

7. All items in the test plan are unchecked. Please confirm you've actually run the manual checks, in particular the quoted Windows path case that motivated the PR.

PR #756, opened 2026-09-13
fix: cache plugin registry across sessions; force-refresh only on demand

coloraven:
## Summary
- Prefer cached plugin registry catalogue instead of a full network pull every time New Connection / Connections mounts the catalogue (often multiple times).
- In-process cache plus on-disk `plugin_registry_cache.json` (12h TTL); re-merge install status from the local plugins dir on soft reads.
- Settings Refresh, post-install, and `plugin-activated` still force a live pull.

## Test plan
- [ ] Open Connections / New Connection twice within a session — second open should not wait on a full registry network round-trip
- [ ] Quit and relaunch within 12h — catalogue should appear from disk cache without a network pull
- [ ] Settings → Plugins → Refresh forces a network refresh
- [ ] Install a plugin, then confirm catalogue install status updates
- [ ] Frontend unit tests: `src/utils/pluginRegistryFetch.test.ts`


debba (changes requested):
@coloraven thanks for this, the catalogue really was being pulled way too often and the overall shape here (memory + disk, force only on demand) is the right one. I checked it out locally and went through the whole flow, install/uninstall/force-install all invalidate correctly and the `force` flag makes it down to the Rust command as described. Two things I need fixed before merging though, plus a few smaller notes.

### Blockers

**1. The Rust tests are flaky.**
The four tests in `registry_cache.rs` share the global `REGISTRY` / `INSTALLED` / `TEST_CACHE_DIR` statics and cargo runs them in parallel. One test resets `TEST_CACHE_DIR` while another is still reading from disk, so the read misses. I ran it 15 times and `disk_survives_memory_invalidate` failed 6 of them:

```
cd src-tauri && for i in $(seq 1 15); do cargo test --lib plugins::registry_cache 2>&1 | grep 'test result'; done
```

There's a nastier side effect too: when `TEST_CACHE_DIR` is `None`, `disk_cache_path()` falls back to the real `get_app_config_dir()`, so a test can end up writing `plugin_registry_cache.json` into the developer's actual config dir. A `static TEST_LOCK: Mutex<()>` held in `with_temp_cache_dir` (and in `disk_helpers_round_trip`) fixes both, or `serial_test` if you prefer.

**2. Stale frontend cache after a deep-link install.**
`useDeepLinkInstall` emits `tabularis://plugin-installed`, but the only listener is `PluginsTab`. The module-level `memoryCache` in `pluginRegistryFetch.ts` has a 12h TTL and nothing invalidates it in that flow. So: install a plugin via `tabularis://install/<slug>` with Settings closed, then open New Connection. It does a soft fetch, hits the frontend cache and shows the plugin as not installed for the next 12 hours. Before this PR every mount re-scanned, so this is a regression. Easiest fix is calling `invalidatePluginRegistryCache()` in `confirm()` of `useDeepLinkInstall`, or have `useConnectionCatalogue` also listen to `plugin-installed`.

### Non-blocking

- `.rules/testing.md` puts tests under `tests/utils/`, not `src/utils/`. There's precedent in the repo for the latter so I won't insist, just flagging it.
- The `plugins` module already keeps tests in sibling files (`tests.rs`, `force_install_tests.rs`), inline tests in `registry_cache.rs` break that pattern (`.rules/rust.md` #4).
- `useConnectionCatalogue.ts` got reformatted from single to double quotes across the whole file. Nothing enforces that and it makes the diff noisy.
- Small race in `fetchPluginRegistry`: a force call issued while a soft fetch is in flight starts a second invoke, and whichever resolves last wins `memoryCache`. On a slow network the soft one can land after the force and put stale data back. A sequence counter on the request would close it.
- After install the frontend forces a full network pull, but the backend already reconciles install status from the disk cache on soft reads. A soft fetch after `invalidatePluginRegistryCache()` would avoid the round-trip. Design choice, your call.
- The frontend cache doesn't know about `registry_key`, so if `tabulariumRegistryUrl` changes the backend misses but the frontend keeps serving the old catalogue. Low impact since that's only editable in the config file today.

Everything else checked out: vitest 4/4, eslint and tsc clean, fmt clean on the touched files. Once the two blockers are sorted I'll test the deep-link path again and merge.

debba:
@coloraven quick follow-up: main has moved quite a bit since my review and the PR now conflicts in 4 files.

The startup work in d16d5bb2 already added frontend deduplication (`PluginRegistryProvider` with a request counter, keyed on the registry URL and refreshed on `plugin-installed` / `plugin-activated`, plus a shared resource in `useConnectionCatalogue`). So `pluginRegistryFetch.ts` and the hook changes are no longer needed, and dropping them also takes care of the deep-link blocker and the race note.

The Rust side is still very welcome, since every `fetch_plugin_registry` call on main still hits the network. When you rebase:

1. Keep `registry_cache.rs`, but fix the parallel test issue (a shared `TEST_LOCK` or `serial_test`) and move the tests to a sibling `registry_cache_tests.rs`.
2. `fetch_plugin_registry` now resolves `installed_version` for themes from the theme catalog. `reconcile_install_status` needs the same logic, otherwise installed themes show as not installed on every cache hit. A shared helper used by both paths would be ideal.
3. Only the Settings Refresh button should pass `force: true`. The rest can stay soft, since the backend reconciles install status on read.
4. Consider writing the disk cache to a temp file and renaming it, as Kilo suggested.

Happy to retest once it's rebased.

PR #766, opened 2026-09-15
feat(connections): connect through AWS SSM Session Manager port forwarding

### LINKED ISSUE #765: [Feat]: Connect through AWS SSM Session Manager port forwarding
### Prerequisites

### 🚀 Feature Proposal

Add AWS Systems Manager Session Manager port forwarding as a connection method, alongside the existing SSH tunnel and Kubernetes port-forward methods.

A new "AWS SSM" tab on the connection modal would take a managed node id and optionally an AWS profile and region. On connect, Tabularis opens an `aws ssm start-session` port-forwarding session and points the driver at the resulting local port, exactly as the SSH and Kubernetes methods already do.

Both SSM documents are needed:

- `AWS-StartPortForwardingSession` when the target host is loopback, meaning a port on the managed node itself.
- `AWS-StartPortForwardingSessionToRemoteHost` for any other host, reaching an RDS instance or ElastiCache node through the managed node.

Supporting only the remote-host document would lock out users whose IAM policy grants only the plain one, which is a common restriction.

### Motivation

Plenty of production databases sit in a private subnet with no bastion host to SSH into. The AWS-native way in is Session Manager: it needs no inbound port, no SSH key distribution and no public IP, and every session is logged in CloudTrail. Teams on that setup currently have to keep `aws ssm start-session` running in a terminal and point Tabularis at the local port by hand, remembering to restart it whenever AWS times the session out.

This is the same shape of problem the existing SSH and Kubernetes connection methods already solve, so it fits the established pattern rather than adding a new concept.

### Example

Connecting to an RDS PostgreSQL instance through a bastion EC2 instance:

```
General tab
  Host: db.abc123.eu-west-1.rds.amazonaws.com
  Port: 5432

AWS SSM tab
  Use AWS SSM:  on
  Managed node: i-0123456789abcdef0
  AWS profile:  prod
  AWS region:   eu-west-1
```

Tabularis resolves this to:

```
aws ssm start-session \
  --target i-0123456789abcdef0 \
  --document-name AWS-StartPortForwardingSessionToRemoteHost \
  --parameters '{"host":["db.abc123.eu-west-1.rds.amazonaws.com"],"portNumber":["5432"],"localPortNumber":["<ephemeral>"]}' \
  --profile prod --region eu-west-1
```

and connects the driver to the resulting `127.0.0.1:<ephemeral>` port.

Credentials would be delegated entirely to the AWS CLI, so SSO, assumed roles and credential processes work as configured and nothing AWS-specific needs to be stored by Tabularis.

I have an implementation of this ready and will open a PR referencing this issue.

Davydhh:
Closes #765

Adds AWS Systems Manager Session Manager port forwarding as a connection method, alongside the existing SSH tunnel and Kubernetes port-forward.

## How it works

A new **AWS SSM** tab on the connection modal takes a managed node id, plus an optional AWS profile and region. On connect, `resolve_connection_params` opens an `aws ssm start-session` port-forwarding session and rewrites host/port to the resulting local port, the same way the SSH and Kubernetes methods do. The forward target is the connection's own host/port from the General tab, so there are only four new fields on `ConnectionParams`.

The SSM document is derived from that host rather than asked for:

| Target host | Document |
|---|---|
| loopback (blank, `localhost`, `127.0.0.1`, `::1`) | `AWS-StartPortForwardingSession` |
| anything else | `AWS-StartPortForwardingSessionToRemoteHost` |

Both are supported deliberately: some IAM policies grant only the plain document, and a remote-host-only implementation would lock those users out.

Credentials are delegated entirely to the AWS CLI, so SSO, assumed roles and credential processes work as configured. Nothing SSM-specific is written to the keychain and no credential material passes through the app.

## Notes for review

Three decisions are worth a look:

**Readiness comes from the spawned process, never from a TCP probe.** An earlier draft probed the local port to decide the tunnel was up. That is unsafe: the probe also succeeds against a foreign process that won the race for that port between our bind and the plugin's, and the user would then silently query the wrong database. Readiness is now taken only from the plugin reporting its own open port on stdout. The trade is that a wording change in the plugin costs a loud timeout, which is the safe direction to be wrong.

**Teardown signals the whole process group.** `aws ssm start-session` spawns `session-manager-plugin` as a child, and that child is what holds the forwarded port. Killing only the CLI leaks the plugin and the port, so the group gets a SIGTERM on Unix and `taskkill /T` on Windows. SIGTERM rather than SIGKILL lets the CLI close the session server-side instead of leaving it to time out. This adds a `libc` dependency under `cfg(unix)`.

**Cached sessions are checked for liveness before reuse.** AWS ends sessions on idle timeout or at `MaxSessionDuration`, so a key match alone is not enough; without the check, every connect after a timeout would hand back a dead local port until the app restarted.

AWS CLI failures are classified into actionable messages (plugin missing, expired SSO, absent credentials, `TargetNotConnected`, IAM denial, port conflict, wrong region) with the raw AWS output kept alongside. The second commit fixes the frontend classifier, which was re-bucketing an IAM "Access denied" as a database login failure and a missing `aws` binary as an unreachable database server.

## Out of scope

Static access-key fields, a `session-manager-plugin` path override, an explicit local-port override, EC2/RDS target discovery, and Secrets Manager passwords. All are additive if wanted.

## Testing

- 32 new Rust tests covering argument construction, document derivation, readiness matching, failure classification, tunnel-key collision and session liveness.
- 6 new frontend tests covering document resolution and error classification.
- Full suites pass: 1316 Rust (the 5 failures on `main` for askpass socket paths and `storage_location` are pre-existing and unrelated), 4551 frontend across 286 files. `tsc --noEmit` and eslint clean.
- Manually verified against an EC2 managed node forwarding to an RDS PostgreSQL instance, and against a port on the node itself.

i18n keys are added to all 11 locales.


debba:
Hi @Davydhh , 
thanks for this PR, can you fix merge conflicts?

Davydhh:
@debba Sure, done

PR #775, opened 2026-09-16
feat: add Windows Authentication checkbox for integrated-auth drivers

egertaia:
## Summary
- Add `supports_integrated_auth` to `DriverCapabilities` (Rust + TS) and `integrated_auth` to `ConnectionParams` (Rust + both TS `ConnectionParams` shapes — the shared one in `src/utils/connections.ts` and the modal's own local one), forwarded verbatim to plugins over the existing JSON-RPC `params` payload.
- Add a "Use Windows Authentication" checkbox to `NewConnectionModal`, shown only when the active driver declares `supports_integrated_auth`. Checking it hides and clears the username/password fields instead of asking the user to hand-write a connection string for something the UI could just ask for directly.
- Wire up the three built-in drivers (mysql, postgres, sqlite) with the new capability field set to `false`, since `DriverCapabilities` is a plain struct literal at each of those call sites.

Companion plugin-side PR (SQL Server is the first driver to opt in): TabularisDB/tabularis-sqlserver-plugin#25

## Test plan
- [x] `npm run typecheck` — clean
- [x] `npx vitest run tests/components/modals/NewConnectionModal.test.tsx` — 32/32 passing, including 3 new tests for the checkbox (shows only for `supports_integrated_auth`, hides/restores username+password on toggle)
- [x] Manually verified end-to-end via a Windows dev build (`pnpm tauri dev`) against a real domain-joined SQL Server: checked the box, connected via SSPI with no username/password, same as SSMS


debba:
Thanks a lot for this, and for testing it against a real domain-joined SQL Server.

I made some changes to keep the core footprint as small as possible: the host side is now a generic hook on the `connection-modal.extra_fields` slot (`setCredentialFieldsHidden`), while the "Use Windows Authentication" checkbox and the `integrated_auth` flag move into the SQL Server plugin (UI extension plus the opaque `extra` map, so no new capability and no new `ConnectionParams` field in core). See #780 for the details.

Tomorrow I will look at how to integrate that with the work you did here and in TabularisDB/tabularis-sqlserver-plugin#25, so nothing of your testing gets lost. I will keep you posted.

debba:
Closing this in favour of #780, which supersedes it. Thanks again @egertaia for the work here and for the end-to-end test against a real domain-joined SQL Server.

To recap why #780 replaces this PR rather than building on top of it: the goal is the same (a "Use Windows Authentication" checkbox that hides the login inputs), but the core change is now driver-agnostic. Instead of a `supports_integrated_auth` capability and a first-class `integrated_auth` field in `ConnectionParams`, the host only exposes `credentialFieldsHidden` / `setCredentialFieldsHidden` on the existing `connection-modal.extra_fields` slot. The checkbox itself, and the `integrated_auth` flag, live in the SQL Server plugin as a UI extension and travel in the opaque `extra` map that the host already persists and forwards. That keeps core free of SQL Server specific knowledge and lets other drivers (Kerberos, IAM tokens) reuse the same hook.

The modal behaviour in #780 (hide and clear username/password, restore them on toggle, reset on driver change) and its tests follow what you designed and verified here, so you are listed as co-author on the #780 commit. The plugin side builds on TabularisDB/tabularis-sqlserver-plugin#25 and will be opened there as a follow-up, again with your work credited.

If you have a chance, it would be great if you could test the #780 branch together with the updated plugin on your domain-joined Windows machine, since that is the one scenario I cannot reproduce locally. Once it is confirmed it will land in the next nightly / release.

Issue #771, opened 2026-09-16
Feature proposal: share connections and their credentials with a team, through an encrypted vault on a shared folder


drbelt27:
## The problem

My team of a handful of developers shares a network folder (SharePoint/OneDrive, but any SMB share would do). We work on the same ~40 database servers, almost all behind per-server SSH tunnels. Today every one of us retypes the same hosts, users, passwords and tunnel credentials by hand, and when a password rotates we find out one by one, when a connection stops working.

The existing export/import gets a colleague started once, but it is a one-shot copy: after that the two machines drift apart, and there is no way to keep a set of connections in sync across a team.

I have implemented this for my own team and it is working, but before I send a pull request I would rather know whether you want the feature in core at all, and in what shape. The code touches `connections.json`, the keychain path and `find_connection_by_id`, so it is not something I want to drop on you as a surprise.

## What I built

An opt-in vault file on a folder the team already shares.

**Encryption.** The payload is AES-256-GCM under a key derived with Argon2id from a master password the team agrees on. It reuses `export_crypto`, which already does exactly this for encrypted exports â€” I only exposed `derive_key` plus a `seal`/`open` pair for callers that keep their own envelope. The header stays in plain text (format, revision, KDF parameters, a verifier blob) so a wrong password is reported as a wrong password instead of surfacing as a corrupt file. The salt is generated once and lives in the header, so every member derives the same key.

**The master password is never persisted.** Not in the config, not in the keychain. It is asked once per launch, the 32-byte key is derived once and kept in memory for the session. That is the point of the feature, and it is also why shared connections are unusable while the vault is locked: `find_connection_by_id` asks the vault for a shared connection's secrets and fails if it is locked, rather than falling back to anything else.

**How it fits the existing model.** Shared connections are materialized into the local `connections.json` *without* their secrets, flagged `shared: true`. Every existing screen â€” the list, groups, tags â€” keeps working untouched. `save_connections_file` strips the secrets of a shared connection centrally, so no code path can leak one into the file.

**Concurrency.** The shared file is never held open. Each operation is a read, an in-memory merge, and at most one write, all inside a lock file that is released immediately (stale locks older than 30s are broken). The merge is a proper three-way merge against the payload of the last successful sync, cached locally and sealed with the same key â€” without that base you cannot tell "my colleague added this" from "I deleted this". Deletions propagate as tombstones that expire after 90 days. When the same record changed on both sides the local version wins and the resolution is logged.

[...]

PR #776, opened 2026-09-16
refactor(export-crypto): expose the key-and-seal primitives

drbelt27:
First of the split proposed in #771. Small and self-contained: no behaviour change, no new feature, just the shape `export_crypto` needs so a caller can keep a long-lived envelope.

## Why

`encrypt` and `decrypt` derive a key from a password on every call. That is right for a one-shot export file, where the envelope carries its own KDF parameters and is opened once.

A caller that owns a long-lived file wants the opposite: derive the key once, keep it for the session, then seal and open repeatedly under it, with the KDF parameters stored in its own header. Doing that today means duplicating the Argon2id and AES-GCM setup, which is exactly the kind of duplication you do not want two copies of.

## What changes

- `derive_key` and the Argon2id defaults (`ARGON2_M_COST`, `ARGON2_T_COST`, `ARGON2_P_COST`, `SALT_LEN`) become public.
- New `random_salt()`, for a caller that generates a salt once and stores it.
- New `Sealed { nonce, ciphertext }` with `seal_with_key` / `open_with_key`.

Every seal draws a fresh nonce from `OsRng`. That matters more here than it does for `encrypt`: reusing a nonce under a long-lived AES-GCM key is fatal, and a long-lived key sealing repeatedly is precisely this API's use case. There is a test asserting two seals of the same plaintext differ in both nonce and ciphertext.

`encrypt` and `decrypt` are untouched, and the existing export path behaves exactly as before.

## Tests

Eight new tests next to the existing ones:

- the same password and salt derive the same key, a different password or salt does not;
- `random_salt` is not a constant and has the expected length;
- seal/open round-trips;
- every seal uses a fresh nonce;
- another key cannot open a sealed blob;
- a tampered ciphertext is rejected (GCM authenticates, so it must fail rather than decrypt to junk);
- a malformed or empty nonce is rejected.

They use cheap Argon2 parameters so they stay fast; the production parameters are still exercised through `encrypt`/`decrypt`.


drbelt27:
Fixed in e0abf86.

You are right that making `derive_key` public reopens the guard `decrypt` provides: the parameters always come from a file somebody else may have written, and a caller keeping its own envelope reads them from its own header without ever passing through `decrypt`. The team vault I am building on this is exactly that caller, and its header sits on a folder several people can write to — an unbounded `m_cost` there would be a denial of service on every unlock attempt, including the failed ones.

The `ARGON2_MAX_*` check now lives inside `derive_key`, so every caller gets it rather than each having to remember. `decrypt` keeps its own check, which produces a clearer message and short-circuits before touching the payload.

Two tests added: the limits are refused on each of the three parameters, and the production parameters still pass.

drbelt27:
Rebased onto current `main` (0.25.0) — applied clean, no conflicts.

Flagging this one specifically because it is the cheapest thing in the stack to say yes to:

- no behaviour change: `encrypt`/`decrypt` are byte-for-byte what they were, this only makes `derive_key`, `seal_with_key` and `open_with_key` reachable by callers that keep a long-lived envelope;
- it does not depend on the team-share feature landing, or on how the OneDrive question in #771 resolves. If you decided tomorrow that the vault should be a plugin rather than core, this PR would still be the right shape;
- the KDF bounds Kilo asked for are in, and they now live inside `derive_key` rather than only in `decrypt` — which is the actual fix, since a caller reading parameters out of a file somebody else wrote is exactly the exposed path.

Local on the rebased branch: `cargo check --all-targets` clean, export_crypto suite green.

One thing worth knowing if you are looking at the checks: the red cross is Kilo reporting *"Insufficient credits to run review — Review could not start because the account has insufficient credits."* It is a billing notice on your side, not a finding against this code. The repo's own CI has never run on this PR at all — it is sitting in `action_required`, waiting for a maintainer to approve workflows for a first-time contributor.

PR #777, opened 2026-09-16
feat(ssh): resolve environment variables in SSH key paths

drbelt27:
Independent of the team-share series in #771 — it stands on its own and touches nothing those PRs touch.

## The problem

An SSH key path typed as `%USERPROFILE%\ssh_keys\id_ed25519`, `$HOME/.ssh/id_rsa` or `~/.ssh/id_rsa` is taken literally, so the tunnel does not find the file. People work around it with hardlinks and junctions; one of my colleagues had `mklink /J C:\ssh_keys "%USERPROFILE%\ssh_keys"` written in a connection comment so the team could share one absolute path.

## What changes

Expansion happens when the tunnel is opened, never when the path is saved, so the stored profile keeps the variable in it and stays portable.

Accepted everywhere, on every platform: `%NAME%`, `${NAME}`, `$NAME`, and a leading `~`. `HOME` and `USERPROFILE` both resolve to the home directory on all three platforms, so a path written on Windows still resolves on macOS and Linux — which matters to anyone syncing their config folder between machines, and would matter more if shared connection profiles ever land.

Two deliberate choices:

- A name that resolves to nothing is left **exactly as typed**, not replaced with an empty string. The user then sees the unresolved variable in the error instead of a silently wrong path.
- A leading `~` only counts when it is the whole first segment, so `~other/.ssh/id_rsa` (another account's home, which cannot be resolved) and a file literally named `~note.txt` are left alone.

Applied in `SshTunnel::new_with_tcp_override` rather than in `new`, so the proxy-fronted path introduced recently gets it too, and in the test entry point, so "Test connection" reads the same file as the real connection.

## Tests

13 tests on `expand_with`, which takes its variable lookup and home directory as parameters and so never touches the real environment: each syntax, several variables in one path, the unknown-name and empty-value cases, a tilde that is not the first segment, stray `%` and `$` that are not variables, a name starting with a digit, and no home directory at all. One test covers the `HOME`/`USERPROFILE` aliasing against the real environment, asserting both spellings agree on whatever platform CI runs.

## Note

The UI hint mentions `~` as the portable form. I did not touch anything about how paths are validated or stored.


drbelt27:
Fixed in feb82c2.

Correct and my mistake: `Option::is_none_or` was stabilised in 1.82 and the crate declares `rust-version = "1.77.2"`. It compiled here only because my toolchain is newer. Swapped for `map_or(true, ...)`, same behaviour, with a comment saying why so nobody "modernises" it back.

drbelt27:
Rebased onto current `main` (0.25.0) — applied clean, no conflicts.

This one is completely independent of the team-share work. It came out of the same afternoon, which is why it is in the same stack, but it touches nothing the vault touches: it expands `%VAR%`, `$VAR` and `~` in the SSH key path, on all three platforms, with its own test module and no other behaviour change. It would be reviewable and mergeable if the rest of #771 never happened.

The MSRV point from the earlier review is fixed — `is_none_or` is 1.82, the crate declares 1.77.2, so it is back to `map_or`.

Same caveat on the checks as the rest of the stack: the red cross is Kilo saying the account is out of credits, not a finding. Real CI has never run — the workflow runs are stuck in `action_required` pending maintainer approval.

PR #778, opened 2026-09-16
feat(team-share): vault file format and three-way merge

drbelt27:
Second of the split proposed in #771, and the piece worth arguing about: the file format and the merge, with no knowledge yet of `connections.json`, the keychain or the UI. Those come in the next one.

> **Stacked on #776.** It builds on the `seal`/`open` primitives there, so the diff below includes that PR's two files. Review the second commit, or merge #776 first and this collapses to the vault alone.

## The file

One JSON file on a folder the team already shares. The header is plain text — format marker, revision counter, KDF parameters, a verifier — so a member can be told *their master password is wrong* rather than *the file is corrupt*. Everything else is AES-256-GCM under a key derived with Argon2id, with the salt fixed for the life of the file so every member derives the same key from the same password.

No handle is ever held. A read slurps and closes. A write goes to a sibling temp file renamed over the target while a lock file is held, released immediately after. POSIX `rename` replaces atomically, so macOS and Linux go straight to it; only Windows needs the target gone first, and that window is covered by the lock.

## The merge

Three-way, against the payload of the member's **last sync**, not against the file alone. Without that base, "my colleague added this" and "I deleted this" are the same observation — an entry present remotely and absent locally — and one of the two interpretations loses data. The base is why the merge can tell them apart.

- Removals become tombstones, so they propagate to members who have not synced yet; they expire after 90 days so the file does not grow forever.
- A record changed on both sides since the base resolves to the **local** version and reports it, so the caller can surface the fact rather than let it pass silently.
- Two identical records are short-circuited before the change detection, so a first sync with no base does not report a conflict over records that agree.

It is a pure function: no clock, no filesystem, no app handle. Everything it needs — the timestamp, the actor label, the tombstone cutoff — arrives in a `MergeContext`.

## What travels

Connections with their credentials, the groups and tags they need, the SSH profiles they tunnel through (with the tunnel password and key passphrase) and the Kubernetes tunnels they route through (which carry no secrets). A tombstone drops the credentials of the record it replaces.

You raised versioning in #771: the header carries a format `version`, and `read` refuses a file written by a newer format rather than guessing at it. If #753 reshapes the SSH model, that is the lever.

## Tests

35, all on this layer alone:

- the merge: pulling a teammate's addition, pushing a local one, a remote edit landing when nothing changed locally, a conflict resolving to local, a password change reaching the share, a deletion becoming a tombstone without credentials, an entry never synced here not being read as a deletion, identical sides not being a conflict, a remote tombstone keeping its stamp, expiry, two members adding different records, and the same rules for groups, tags, SSH profiles and K8s tunnels;
- the file: the right password opening the payload, a wrong one rejected *before* the payload is touched, credentials not being readable in the file, resealing bumping the revision, a write/read round-trip leaving no temp file behind, overwriting an existing vault, a missing file reading as none, a foreign JSON being rejected;
- the lock: a second holder refused, the lock freeing on drop, a stale lock broken;
- the appearance rules, which decide what is portable between machines.

## Not in this PR

Nothing reads or writes this file yet — `mod.rs` is only the two module declarations. The wiring, and the answers to your points about the keychain, the caches and the locked-vault UX, are in the follow-up.

I have also not yet run the OneDrive check you asked for. Two of us are using the feature daily against a SharePoint-synced folder and it is holding up, but the deliberate test — two members writing within seconds of each other — has not been done, and I am not going to describe it as verified until it has. If the lock turns out not to serialize writers there, this is the PR that changes.


drbelt27:
All three addressed; the branch is rebased on the updated #776.

**The unlock DoS (warning).** Right, and the sharpest of the three: the KDF parameters come from the plaintext header of a file that, by this feature's own threat model, several people can write. A crafted `m_cost` would hang or exhaust memory on every unlock, including wrong-password attempts, which makes it trivially reachable. Fixed at the root in #776 — the `ARGON2_MAX_*` bounds now live inside `derive_key` itself, so this path gets them without the vault having to remember to ask.

**Delete over a concurrent edit (suggestion).** Also right, and it is the same failure this feature's conflict reporting exists to prevent. The removal still wins, but a teammate's edit is being discarded, so reporting it as routine housekeeping hides exactly what the caller needs to surface. It now reports a conflict when the remote side also changed since the base, and stays a plain removal when nobody touched it. Both cases have a test — you were right that neither was covered.

**Leaked temp file on Windows (suggestion).** Fixed: the `?` returned before the cleanup, so every failed replace left an orphan in the team's folder. The temp file is removed on that path now.

37 tests on this layer, all passing.

drbelt27:
Rebased onto current `main` (0.25.0) — applied clean.

Since this is the PR the OneDrive question actually lands on, the field report, which I have also put on #771:

Two of us have been running the vault daily against a SharePoint-synced folder — since before I opened these PRs and every working day since. Connections created, edited, moved between groups, credentials rotated, SSH profiles changed, both of us working at the same time without coordinating. No concurrency problems. Nothing lost, nothing repaired by hand, no vault that failed to open.

I want to be exact about the limits of that, because @debba asked for something specific and I said I would not overclaim it. That is weeks of ordinary parallel use. It is **not** two members deliberately writing inside the same second to prove whether the lock file serialises anything through a sync client. I have not staged that test, so I am not calling the design verified.

If you want it before this merges, say so and I will run it this week — I would rather do it against a live merge target than leave it hanging. And if it turns out the lock does not hold under a sync client, this is the PR that changes: the engine is separate from the integration precisely so that finding would not force a rewrite of #779.

The review points from the last round are in: delete-over-edit now reports `ConflictLocalWins` when the remote side also changed, so a teammate's concurrent edit is surfaced rather than silently swallowed, and the Windows replace path cleans up its temp file when the rename fails.

Suite is 46/46 on the rebased branch, `cargo check --all-targets` clean.

On the checks: the red cross is Kilo reporting *"Insufficient credits to run review"* — a billing notice on the account, not a finding. The repo's own CI has never executed on this PR; the runs are sitting in `action_required` waiting for a maintainer to approve workflows for a first-time contributor.

PR #779, opened 2026-09-16
feat(team-share): wire the vault into connections, the keychain and the UI

drbelt27:
Third of the split proposed in #771: the engine from #778 put to work.

> **Stacked on #778, which is stacked on #776.** The diff below therefore includes both. Review the last commit, or merge the two beneath it first and this collapses to the integration alone.

## What a user gets

Settings → Storage → Team share points Tabularis at a vault on a folder the team already shares, and the connections they pick travel in it with their credentials.

Sharing itself is decided on the **connections screen**, on the multi-selection that is already there, so a whole selection costs one read-merge-write against the share rather than one per connection. A filter separates shared from local, and a badge marks the shared ones. A picker in Settings was the first design and I threw it away: with forty servers it is an unusable list.

## How it sits in the existing model

A shared connection is materialized into `connections.json` **without its secrets** and flagged `shared`, so the list, groups, tags and every other screen keep working untouched. `save_connections_file` strips the secrets of a shared connection centrally, so no code path can leak one into the file.

`find_connection_by_id` asks the vault for those secrets instead of the keychain, and **fails while the vault is locked** rather than falling back — that is what makes the master password actually gate anything. The IAM-auth exception is preserved: a 15-minute RDS token is stale wherever it is stored, so a shared IAM connection gets no password from the vault either, exactly as it gets none from the keychain today.

The master password is never persisted. It is asked once per launch, the key is derived once and wiped on drop (`zeroize`). Per your point, reaching a shared connection while the vault is locked now **raises the prompt** rather than only reporting an error: the backend emits an event the gate opens on. The operation still fails that once and has to be repeated — an automatic retry seemed more invasive than it was worth, but say the word.

Also per your point, a conflict resolved in favour of this machine raises a **persistent toast** naming the count, not just a log line. The motivating case is exactly the one you described: a colleague rotates a password, I had edited the same connection, and my stale copy wins.

## What else travels

- **SSH profiles** the shared connections tunnel through, with the tunnel password and key passphrase. They join and leave the share automatically with the connections that use them, moving their secrets between the keychain and the vault as they go. Without this a teammate receives a connection pointing at a profile they do not have.
- **Kubernetes tunnels**, which you asked about. They carry no credentials, so joining or leaving is only a flag. Their `kubectlPath` and `kubeconfigPath` are machine-local; #777 makes an environment variable in them portable.
- **Groups and tags**, exactly while a shared connection needs them — a group that loses its last shared connection is tombstoned rather than left as an empty folder for the whole team. Ancestors come along so the hierarchy is never orphaned.
- **Accent colour and icon**, in their portable forms. An uploaded image is a file under the member's own data directory, so it does not travel; a member's own image is kept when the team brings no icon, since syncing should not silently undo a choice they cannot get back.

Editing a shared connection — its parameters, appearance, tags, or the group it sits in — reaches the share on save rather than at the next sync.

## Caches

`ConnectionCache` is invalidated through the existing `save_connections_and_invalidate` on every materialize. `CredentialCache` entries are invalidated when a connection joins the share and repopulated when it leaves.

## Threat model

Whoever has the master password and read access to the folder has **every shared credential**. There is no per-member revocation: when someone leaves the team, the master password and the whole payload have to be rotated. That belongs in the docs before this ships, and I would rather write it against whatever shape you settle on than guess now — tell me where it should live and I will add it in this PR.

## Still open

- Whether credential fields should resolve by most recent change rather than local-wins. You suggested it and I think you are right, but it makes the merge field-aware, which is a real change to #778 and I did not want to make it unilaterally.
- The OneDrive check. Two of us use this daily against a SharePoint-synced folder and it holds up, but the deliberate concurrent-write test is not done, and I will not call it verified until it is. If the lock does not serialize writers there, #778 changes and this PR follows.

## Tests

44 Rust tests on the layer below, plus 28 frontend tests here covering the status helpers, the shared/local filter and the master-password validation. The existing suite passes; the frontend typechecks and lints clean.


drbelt27:
Rebased on the updated #778, which carries the three review fixes from that PR plus the KDF bounds from #776. Nothing in this PR's own layer changed.

The Kilo check on this PR reports a failure with no findings posted — it ran for 15 minutes and produced no comment, so it looks like it timed out on the size of the diff rather than found anything. Happy to act on it if it comes back with something.

Worth repeating from the description, since it is the one thing I cannot yet answer: the concurrent-write test on a sync client (OneDrive/SharePoint) has not been run. Two of us use this daily against such a folder and it holds up, but the deliberate two-writers test is still outstanding, and if the lock file turns out not to serialize writers there it is #778 that changes, not this PR.

drbelt27:
Rebased onto current `main` (0.25.0). This one had conflicts, now resolved — worth describing what changed, since both are cases where `main` moved in a good direction and my code had to follow:

- **`ConnectionChips`.** The badge row in the card and the list row has been extracted into a shared component. My "shared" badge was duplicated inline in both, which is exactly the duplication that refactor removed, so it now lives in `ConnectionChips` once, placed with environment and tags (provenance) rather than next to SSH/K8s (transport). Both call sites are untouched upstream code again.
- **Design tokens.** The new `check-theme-tokens` script rightly rejected every raw Tailwind colour in the team-share UI. All of it is on theme tokens now: the accents on `accent-primary`, errors on `accent-error`, warnings on `accent-warning`, the share badge and its filter on `accent-secondary`, focus borders on the `focus` token, and the primary buttons on `text-inverse` instead of `text-white`. `pnpm lint` is clean, design check included. It also means the feature now themes properly, which it did not before.

Verified on the rebased branch: `cargo check --all-targets` clean, team-share suite 46/46, and `tsc --noEmit`, `eslint` and the design-token check all clean. The vitest tests covering the files this PR touches pass, including the Connections page suite. For transparency: I do get some unrelated reds locally, all on Windows-specific ground — every theme asset checksum mismatches because git checks those files out with CRLF, and a few K8s and theme modal tests hit the 5s timeout and vary between runs. None are in files this PR touches.

Two notes for whoever picks this up.

**This is the PR that rots.** It touches `connections.json`, the keychain, the caches and the connection UI — the busiest surfaces in the repo. It has been open since 16 September and this is the second time it has needed adapting to changes in `main`. I am happy to keep it current while it is being reviewed; I cannot realistically keep it current indefinitely against a moving core with no review in progress. If there is a blocker, I would much rather hear what it is than keep rebasing.

**The two PRs below it do not have this problem.** #776 and #777 both still merge clean and neither depends on the vault, on this integration, or on how the OneDrive question in #771 resolves. Landing those two would cut the stack in half and take most of the drift out of it.

On the checks: the red cross is Kilo saying *"Insufficient credits to run review"* — your billing, not my code. Real CI has never run on this PR; the runs are stuck in `action_required` pending maintainer approval for a first-time contributor.

Issue #654, opened 2026-08-17
[Feat]: Postgresql different behaviour between double clicking on a table and running a query


mccuskk:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

When double clicking on a table you get all the nice viewers for json etc, when running a query on the same table you just get a tooltip with the contents. It feels very unnatural to me.

### Motivation

Improve workflow

aesslinger:
Investigated this — it's a frontend bug in the main app, not driver-specific despite the "Postgresql" in the title (would reproduce identically on any database driver), and not something a plugin repo could fix.

**Root cause**: both "browse a table" and "run a query" render through the same grid component — `src/components/ui/DataGrid.tsx`. The difference comes from one guard clause in `handleCellDoubleClick`:

```js
if (!tableName || readonlyProp) return;
```

Query-result and notebook-cell-result views always render with `tableName={null}` and `readonly={true}` (`ResultEntryContent.tsx:116-124,164-172`, `SqlCellResult.tsx:91`), so this guard returns immediately — before the code ever reaches the JSON/blob viewer branch that calls `openJsonViewerWindow`. Table-browse mode passes a real `tableName` and no `readonly`, so it doesn't hit the guard and the rich viewer opens normally.

The "tooltip" is just the native HTML `title` attribute set on every cell (`DataGridRow.tsx:436`) — it's present in both modes, but it's the *only* thing available for query results since the double-click handler bails out before reaching the viewer logic.

**Why this should be an easy fix**: `openJsonViewerWindow` already accepts and handles a `readonly` flag — it's called with `readonlyProp ?? false` at the one call site that's currently reachable (table-browse mode). The guard is just too broad: it's meant to gate the *edit* logic that follows (building an UPDATE, which legitimately needs `tableName`/primary key columns), but it also blocks the read-only-safe viewer-opening branch that comes before any of that edit logic.

**Confirmed not driver-specific**: `isJsonCellTarget` and the rest of the `DataGrid`/`DataGridRow` component tree have no driver-gating at all — cell-type detection is generic. This would reproduce the same way on MySQL, SQLite, etc.

**Fix location**: `src/components/ui/DataGrid.tsx`, around the `handleCellDoubleClick` guard (~line 892) — move the `!tableName || readonlyProp` check so it only gates the edit-flow code below, not the JSON/blob-viewer-opening branch above it.

PR #781, opened 2026-09-16
fix: expose JSON viewer in read-only context menus

igorzelaya-io:
## Summary

- allow the existing cell context menu to open for tableless read-only result grids
- expose **Open in JSON Editor** for JSON columns outside the editable-only action group
- keep mutation actions such as Set NULL, paste, row editing, duplication, and deletion hidden in read-only grids
- add regression coverage that opens a query-result JSON cell from the context menu and verifies the viewer is read-only

## Context

This is a focused follow-up to item 2 in @debba's review on #763, where the JSON viewer context-menu action was noted as unavailable for read-only query results. It continues the work for #654 while remaining independent of #763 and based directly on `main`.

Review note: https://github.com/TabularisDB/tabularis/pull/763#pullrequestreview-5162285928

## Validation

- DataGrid suite: 41 passed
- Full frontend suite: 271 files / 4,417 tests passed
- TypeScript typecheck: passed
- ESLint: passed
- GitNexus staged impact check: medium risk, limited to the expected DataGrid/result-view flows


debba:
Can you resolve merge conflicts? @igorzelaya-io

debba (changes requested):
Thanks for the follow-up. Checked it out locally: DataGrid suite 45 passing, typecheck and eslint clean. The change works for materialized-view browse and read-only drivers, but it does not reach the query-result case from the #763 note:

1. A query tab whose statement has no extractable table (JOIN, expression, `SELECT '{"a":1}'::jsonb AS payload`) renders with `tableName={null}` and `readonly={false}` on a writable driver, so `if (!tableName && !readonlyProp) return` still suppresses the menu entirely.
2. Stacked results (`ResultEntryContent`) and notebook cells (`SqlCellResult`) pass `readonly` but no `columnMetadata`, so `isJsonColumn(colDataType)` is always false there and the item never appears. Double-click works because `isJsonCellTarget` also detects JSON by value.

The new test passes because it combines `tableName={null}`, `readonly` and `jsonb` metadata, a shape no caller currently produces.

Suggestion: drop the early return, derive `isReadonlyGrid = !!readonlyProp || !tableName` (the edit guards already treat tableless grids as non-editable) and use it for the three mutation groups, and build the JSON item with `isJsonCellTarget(colDataType, contextMenu.row[contextMenu.colIndex])`. I prototyped this locally: the existing suite plus two tests rendering the real caller shapes (readonly without metadata, and tableless without readonly) pass, and mutation items stay hidden. The test could then render without `columnMetadata` to match the real callers.

igorzelaya-io:
Implemented the requested follow-up and merged the latest upstream main into the branch.

Changes:
- removed the early return so tableless query-result grids can open their context menu
- derived read-only behavior from either the readonly prop or a missing table name, keeping mutation actions hidden in both cases
- switched the JSON action to isJsonCellTarget so structured values work even when result components do not provide column metadata
- replaced the synthetic test setup with two real shapes: a metadata-less read-only result and a writable-driver query result with no table

Validation after merging current main:
- DataGrid suite: 46/46 passed
- full frontend suite: 4,827/4,827 passed across 314 files
- pnpm typecheck: passed
- pnpm lint: passed

Latest head: 0fbc9e39

PR #782, opened 2026-09-17
feat(autocomplete): harden nearest-table ranking edge cases and quoted identifiers

### LINKED ISSUE #760: [Feat]: Harden nearest-table autocomplete ranking edge cases
### Prerequisites

### 🚀 Feature Proposal

Follow up on #679, which implemented #507, by tightening the remaining edge cases in nearest-table column ranking:

1. **Prioritize shared column names from the nearest table.** Column suggestions are currently deduplicated by label in `scopedAliases` insertion order. If both tables expose the same column, the suggestion can be attributed to the table parsed first and lose the nearest-table boost. Iterate matching tables in nearest-first order before deduplication.
2. **Define and test comma-separated `FROM` behavior.** In `SELECT * FROM orders o, customers c WHERE |`, the analyzer currently resolves `c`, while the `SqlContextInfo` documentation describes the primary `FROM` table. Choose the intended heuristic, align the implementation and documentation, and add regression coverage.
3. **Document the `SELECT`-list limitation.** For `SELECT | FROM ...`, the analyzer only receives SQL before the cursor and therefore cannot infer tables appearing later. Update the API documentation and tests so they do not imply that the `FROM` table is available in this case.
4. **Keep quoted identifiers out of keyword control flow.** Quoted identifiers such as `"in"` or `"as"` must not be interpreted as SQL keywords when followed by `(`. Add regression tests and preserve token-kind information when evaluating parser transitions.

### Motivation

The nearest-table ranking now improves autocomplete in common `JOIN` and `WHERE` cases, but shared column names are especially common in join predicates (`id`, foreign keys, timestamps). Losing the boost for those names weakens the feature in its primary use case. The other cases should be made explicit and covered so future parser changes do not introduce ambiguous or keyword-related regressions.

### Example

```sql
SELECT *
FROM orders o
JOIN customers c ON |
```

If both tables contain `id`, the deduplicated `id` suggestion should be associated with `customers` and receive the same nearest-table rank as other `customers` columns.

Regression coverage should also include:

```sql
SELECT * FROM orders o, customers c WHERE |
SELECT | FROM orders o JOIN customers c ON c.id = o.customer_id
SELECT "in" (
SELECT "as" (
```

wwww-deeeee:
Resolves #760

### Summary of Changes

1. **Prioritize shared column names from the nearest table:**
   - In `src/utils/autocomplete.ts`, sorted `matchingTables` in nearest-table preference before applying `MAX_PARALLEL_FETCHES` truncation and extracting column completions.
   - Ensures that when multiple tables share column names (such as `id`), the nearest table's column is processed first, added to `seenColumns`, and given the highest priority boost (`0_0_`).

2. **Keep quoted identifiers out of keyword control flow:**
   - In `src/utils/sqlContext.ts` within `openFrameClause`, guarded `IN (` and `AS (` keyword transitions behind `prev.isKeyword`.
   - Prevents quoted identifiers (such as `"in" (` or `"as" (`) from being misinterpreted as SQL clauses; they now correctly fall through to `function-args`.

3. **Align and document comma-separated FROM and SELECT-list semantics:**
   - Updated JSDoc documentation on `SqlContextInfo.lastTableRef` to clarify that in comma-separated `FROM` lists (`FROM orders o, customers c WHERE`), `lastTableRef` resolves to the table alias nearest the cursor (`c`).
   - Documented that when the cursor is in the `SELECT` list before `FROM` is reached, `lastTableRef` is `null`.

4. **Regression Tests:**
   - Added unit test in `tests/utils/autocomplete.test.ts` verifying that shared column names like `id` from the nearest table receive `0_0_id`.
   - Added unit tests in `tests/utils/sqlContext.test.ts` for comma-separated `FROM` resolution, null `lastTableRef` before `FROM`, and quoted identifiers before `(`.
   - Verified 121/121 tests pass with `pnpm vitest run tests/utils/autocomplete.test.ts tests/utils/sqlContext.test.ts`.


src/utils/autocomplete.ts
```diff
@@ -347,8 +347,35 @@ export const registerSqlAutocomplete = (
...
+              (b.name.toLowerCase() === nearestRef ||
+                Array.from(scopedAliases.entries()).some(
+                  ([alias, ref]) =>
+                    alias.toLowerCase() === nearestRef &&
+                    ref.name.toLowerCase() === b.name.toLowerCase() &&
+                    (!ref.schema || !b.schema || ref.schema.toLowerCase() === b.schema.toLowerCase()),
+                )),
+          );
+          return aIsNearest === bIsNearest ? 0 : aIsNearest ? -1 : 1;
+        });
```
debba, line 378: Same predicate as the one in the `forEach` below, written once per operand. Extract it and precompute the nearest set so the comparator stays a one-liner (stable sort keeps the other tables in insertion order).

```suggestion
        // The nearest table (the one being joined in ON, the primary FROM table
        // in WHERE, ...) goes first: it wins the label dedupe below for shared
        // column names such as `id`, and it is never dropped by the fetch cap.
        const nearestRef = sqlContext.lastTableRef?.toLowerCase();
        const isNearestTable = (table: TableInfo): boolean =>
          Boolean(
            nearestRef &&
              (table.name.toLowerCase() === nearestRef ||
                Array.from(scopedAliases.entries()).some(
                  ([alias, ref]) =>
                    alias.toLowerCase() === nearestRef &&
                    ref.name.toLowerCase() === table.name.toLowerCase() &&
                    (!ref.schema || !table.schema || ref.schema.toLowerCase() === table.schema.toLowerCase()),
                )),
          );
        const nearestTables = new Set(matchingTables.filter(isNearestTable));
        matchingTables.sort((a, b) => Number(nearestTables.has(b)) - Number(nearestTables.has(a)));

        // Limit parallel fetches to prevent memory spikes
        const MAX_PARALLEL_FETCHES = 5;
```

src/utils/autocomplete.ts
```diff
@@ -359,8 +386,6 @@ export const registerSqlAutocomplete = (
 
         const seenColumns = new Set<string>();
 
-        const nearestRef = sqlContext.lastTableRef?.toLowerCase();
-
         matchingTables.forEach((table, idx) => {
           const columns = results[idx];
           const isNearest = Boolean(
```
debba, line 391: With the set from the previous suggestion, this whole `Boolean(...)` block (through the closing `);` on line 400) collapses to:

```ts
          const isNearest = nearestTables.has(table);
```

GitHub only lets me attach a one-click suggestion to lines inside the diff hunk, hence the plain snippet here.

tests/utils/autocomplete.test.ts
```diff
@@ -827,5 +827,61 @@ describe('autocomplete', () => {
...
+      const customersId = resultWhere.suggestions.find(
+        (s: { label: string; sortText?: string }) =>
+          s.label === 'id' && s.sortText !== undefined && !s.sortText.startsWith('0_0_'),
+      );
+      // If it appears, it must have a higher index prefix.
+      if (customersId) {
+        expect(customersId.sortText).toMatch(/^0_\d+_id$/);
+        const nonNearestIdx = parseInt(customersId.sortText.split('_')[1], 10);
+        expect(nonNearestIdx).toBeGreaterThan(0);
+      }
```
debba, line 884: This assertion passes on main too: in WHERE `orders` is nearest *and* first in insertion order, so the old code already claims `id` for it. The ON case is the one that breaks without your fix (I checked: `expected '0_1_id' to be '0_0_id'` against main). The trailing `if (customersId)` block is dead because the second `id` is always deduped away.

```suggestion
      // In ON, the joined table `customers c` is nearest, but `orders o` comes
      // first in alias insertion order: without nearest-first ordering the
      // shared `id` label would be claimed by orders.
      const onQuery = 'SELECT * FROM orders o JOIN customers c ON ';
      const modelOn = createMockModel(onQuery);
      const resultOn = await provider.provideCompletionItems(
        modelOn,
        { lineNumber: 1, column: onQuery.length + 1 },
      );

      const idSuggestions = resultOn.suggestions.filter((s: { label: string }) => s.label === 'id');
      // Deduplicated to a single `id`, attributed to the nearest table.
      expect(idSuggestions).toHaveLength(1);
      expect(idSuggestions[0].sortText).toBe('0_0_id');
      expect(idSuggestions[0].detail).toContain('customers');
```

debba (commented):
Thanks a lot for this @wwww-deeeee, and for the thorough write-up. I reviewed it locally: the `isKeyword` guard is a genuine fix (the new sqlContext test fails on main and passes here), sorting before the fetch cap also stops the nearest table from being truncated when more than five tables are in scope, and the full suite is green (4429/4429, tsc clean).

Two small things before merging, both attached as suggestions you can apply directly from the review:

1. The shared-column regression test currently passes on main as well. In the WHERE case `orders` is both the nearest table and the first alias in insertion order, so the dedupe already picks it without the fix. Switching to the ON case from the issue (`... JOIN customers c ON `) makes it fail on main with `0_1_id` and pass with your change. The `if (customersId)` block can go too: dedupe guarantees a second `id` never appears, so it is dead code.
2. The nearest-table predicate is now written three times (twice in the comparator, once in the forEach). Extracting it once and precomputing a Set keeps the comparator a one-liner and lets the forEach reuse it.

I could not push to your branch since maintainer edits are disabled, hence the suggestions. Once they are in I will merge.

If you would like to keep contributing you are very welcome, and feel free to join our Discord: https://discord.com/invite/K2hmhfHRSt

debba (changes requested):
Marking this as changes requested so the status reflects my review above: the two points there (the shared-column regression test should cover the ON case and the nearest-table predicate should be extracted once) are the only things left before merging. Both are attached as suggestions you can apply directly.

PR #784, opened 2026-09-17
fix: self-heal MCP driver registry on unsupported-driver misses

### LINKED ISSUE #783: MCP subprocess never reloads plugin/driver registry after startup — "Unsupported driver" for connections migrated/plugin-activated after MCP client launch
## Summary

Originally reported by @ngocmanh1712 in [tabularis-postgresql-plugin#96](https://github.com/TabularisDB/tabularis-postgresql-plugin/issues/96) ("Agent can't query with MCP after upgrade v0.23.0"). Filing here since the root cause is in this repo's MCP server, not the plugin.

> After upgraded to v0.23.0 and migrated postgres connections to new plugin, I got this error when AI Agent try to query.
> Manual query on tabularis application doesn't have this problem.
> ```
> MCP error -32000: Unsupported driver: postgresql
> ```
> Tried:
> - uninstall tabularis and reinstall but didn't help
> - degrade to build in driver and retry but got this "password authentication failed for user "postgres""

## Root cause

`register_drivers_for_mcp()` (`src-tauri/src/mcp/mod.rs`) populates the driver registry for the standalone MCP subprocess exactly once, at process startup:

```rust
async fn register_drivers_for_mcp() {
    driver_registry::register_driver(mysql::MysqlDriver::new()).await;
    driver_registry::register_driver(postgres::PostgresDriver::new()).await;
    driver_registry::register_driver(sqlite::SqliteDriver::new()).await;

    let app_config = config::load_config_from_disk();
    let plugin_configs = app_config.plugins.unwrap_or_default();
    let enabled_ids = app_config.active_external_drivers;
    plugins::manager::load_plugins_with_configs(plugin_configs, enabled_ids.as_deref()).await;
    // ...
}
```

The MCP subprocess is spawned and kept alive by the external AI client (Claude Desktop, Cursor, etc.), independently of the Tabularis GUI process — it is not restarted when the GUI installs a plugin, activates a driver, or migrates a connection. If the MCP subprocess was already running when the user migrated a connection to the `postgresql` plugin (or installed/activated the plugin), its in-memory registry never learns about it, and `resolve_db_driver`'s lookup (`drivers::registry::get_connection_driver`) fails with `"Unsupported driver: postgresql"` for every subsequent query — while the GUI, which hot-registers newly installed/activated plugins immediately, has no such problem. This matches the reporter's exact symptom: manual queries in the app work, only the agent/MCP path fails.

Confirmed present, unchanged, in both `v0.23.0` (the version the reporter ran) and current `v0.24.0`/`main` — `register_drivers_for_mcp` is byte-for-byte identical across both. This is a separate, real bug from `ec3428c7` ("fix: stop persist_activation from reverting concurrent config changes", shipped in v0.24.0), which fixed a related config-clobbering race in plugin activation but does not touch the once-at-startup driver-loading behavior described here.

## Reproduction

1. Launch an MCP client (e.g. Claude Desktop) that spawns `tabularis --mcp` and keeps it running.
2. While that subprocess is still alive, in the Tabularis GUI: install/activate a plugin driver (or migrate an existing connection to one), for a connection the MCP client already knows about.
3. Ask the AI agent to query that connection via MCP.
4. Observe `MCP error -32000: Unsupported driver: <plugin-id>`, even though the same connection queries fine from the GUI.

Workaround: fully quit and relaunch the MCP client (not just Tabularis) after any plugin/driver change, so a fresh subprocess re-reads the current config.

## Suggested fix

Either:
- Make `register_drivers_for_mcp` (or an underlying config/plugin-manager watcher) re-check `active_external_drivers` / installed plugins on each request or on a file-change signal, instead of only at process startup; or
- Surface a clear, actionable error (rather than the generic "Unsupported driver") when the driver id is known-but-not-loaded, prompting the user to restart their MCP client.

aesslinger:
## Summary
- Fixes #783: the standalone MCP subprocess populated its driver registry once at startup and never learned about plugins installed/enabled, or connections migrated to a plugin, after that — surfacing `Unsupported driver: <id>` until the user fully relaunched their MCP client.
- `resolve_db_driver` now rescans installed plugins against the latest on-disk config once and retries the registry lookup before giving up, rate-limited to once every 2 seconds so a connection with a genuinely wrong driver id can't force a filesystem scan on every call.
- The rescan (`load_plugins_with_configs`) is made idempotent — plugins already registered (driver or UI-only manifest) are skipped via a new `driver_registry::is_registered` check, with built-in ids (`mysql`/`postgres`/`sqlite`) explicitly excluded so a plugin claiming one still hits the existing collision refusal instead of being silently skipped.

## Test plan
- [x] `cargo test --lib` — new unit tests (`is_registered`, `should_skip_rescan`, `cooldown_elapsed`) plus the full existing `registry`/`mcp`/`plugins` suites pass (1262/1262, excluding 5 pre-existing environment-only failures unrelated to this change)
- [x] `cargo clippy` / `cargo fmt --check` — no new warnings or diffs in changed files; also caught and fixed a real MSRV violation (`Option::is_none_or` needs 1.82, crate declares 1.77.2)
- [x] GitNexus `detect_changes()` — blast radius contained to the touched functions, 0 affected processes, low risk
- [x] **Manual end-to-end repro**, live against real infrastructure (see PR comments for full detail): toggled the actually-installed `postgresql` plugin off/on on disk around one long-running `--mcp` subprocess (no restart), querying a real Postgres instance.
  - Currently-installed pre-fix binary (v0.24.0): query fails with `Unsupported driver: postgresql` both before *and after* the plugin re-appears on disk — bug reproduced.
  - This branch's binary: same failure while the plugin is genuinely absent, then **succeeds** after the plugin re-appears, returning real database names from the live Postgres instance — fix verified, same subprocess, no restart.


aesslinger:
## Re-verification pass

Went back through this with an adversarial eye per request. Found and fixed two real issues in the follow-up commit:

1. **Regression in the collision-refusal path.** The "skip if already registered" guard I added to `load_plugins_with_configs`'s rescan loop matched on `driver_registry::is_registered(id)`, but built-in ids (`mysql`/`postgres`/`sqlite`) are always registered before any plugin directory is ever scanned. That meant a plugin manifest claiming a built-in id would be silently skipped by the fast path instead of ever reaching `load_plugin_from_dir`'s explicit refusal — turning a logged, actionable `PluginLoadError` (the check that stops a plugin from shadowing a built-in driver and harvesting its connections' credentials) into nothing happening at all. Fixed by excluding built-in ids from the skip, so they're always routed through the explicit check. Added `should_skip_rescan` as a pure, directly unit-tested function covering this exact case.
2. **Unbounded rescan frequency.** `resolve_db_driver`'s reload-on-miss had no rate limit, so a connection whose driver id is genuinely wrong (typo, an uninstalled plugin) would force a plugin-directory filesystem scan on every single call in a tight retry loop. Added a 2-second cooldown (`cooldown_elapsed`, unit-tested on synthetic `Instant`s, no real sleeps) — bounded, doesn't affect the self-heal within a normal usage pattern.

Also caught and fixed a real MSRV violation: I'd used `Option::is_none_or` (stable since Rust 1.82), but this crate declares `rust-version = "1.77.2"` in `Cargo.toml`. `cargo clippy` flagged it via `incompatible_msrv`. Replaced with a plain `match`.

Checked for concurrency/TOCTOU risk between the registry miss check and the rescan: the MCP subprocess's JSON-RPC dispatch loop (`run_mcp_server`) reads one line from stdin and fully awaits `handle_request` before reading the next — strictly sequential, single in-flight request per subprocess — so there's no path for two `resolve_db_driver` calls to race against each other within one subprocess.

Full test suite: 1262 passed, 0 failed (5 unrelated failures — askpass unix-socket-path-length and one storage-location test — reproduce identically on the unmodified branch; confirmed by stashing my changes and re-running). `cargo clippy` and `rustfmt --check` clean on every touched file.

aesslinger:
## Live end-to-end repro (real Postgres, real installed plugin, real pre-fix binary)

Wanted more than unit tests for this one, so I reproduced the actual bug and the actual fix against real infrastructure rather than mocks — same connection, same on-disk plugin, same toggle, run twice: once against the currently-installed pre-fix binary as a negative control, once against this branch.

**Setup:**
- The already-installed `postgresql` plugin under the real Tabularis app-data plugin directory (untouched otherwise).
- A local Postgres container (`pg-tabularis-test`, port 54320) with a real `testdb` database.
- One temporary test connection (`driver: "postgresql"`) pointed at that container, added to `connections.json`/`connections.dev.json` and removed again afterward.
- A small Python driver script that spawns a `--mcp` subprocess, sends `initialize` then a real `tools/call list_databases` over stdin/stdout, and toggles the plugin directory's presence on disk *while the subprocess stays alive* — this is the exact scenario in #783 (plugin installed after the MCP subprocess already started).

**Negative control — currently-installed binary (v0.24.0, pre-fix):**
1. Plugin directory hidden at subprocess startup → `list_databases` → `Unsupported driver: postgresql` (expected).
2. Plugin directory restored on disk, **same subprocess, no restart** → `list_databases` again → still `Unsupported driver: postgresql`.

Step 2 is the bug: the plugin is present on disk and would load fine in a fresh process, but this already-running subprocess's in-memory registry never learns about it.

**Fix — this branch, freshly built (`cargo build`):**
1. Same setup, plugin hidden at startup → `Unsupported driver: postgresql` (correct — it genuinely isn't there yet).
2. Plugin directory restored, waited past the 2s reload cooldown, same query on the **same subprocess** → succeeded:
   ```json
   ["postgres", "tabularis_test_secondary", "testdb", "testdb_ssl", "testdb_uri"]
   ```
   Real database names from the live container, no restart.

Environment was fully restored afterward — verified the plugin directory, `connections.json`, and `connections.dev.json` all match their original state, and no stray processes were left running.

aesslinger:
@debba this is ready for review.

debba (dismissed):
Hi, thanks for this one, and for the thoroughness of the write-up. I didn't take the description on faith and went through it end to end, and everything checks out.

**What I did**

I built two debug binaries, main at the merge-base and your branch, and drove both over JSON-RPC on stdin against an isolated config dir (`TABULARIS_DATA_DIR`), using the postgresql plugin I actually have installed and my local Postgres container. Same subprocess throughout, no restart:

- plugin disabled in `activeExternalDrivers` at startup, `list_databases` → `Unsupported driver: postgresql` on both binaries, as expected
- plugin re-enabled on disk, wait 2.5s, same query → main still fails, your branch returns the real database list
- three more queries → exactly one plugin process, no duplicates
- bogus driver id, 5 calls in ~10ms → one rescan visible in stderr, so the cooldown does its job

I also re-read the MCP dispatch loop and agree it's strictly sequential, so the miss → rescan → retry sequence can't race with itself. The 9 new unit tests pass on the branch, `detect_changes` against the merge-base shows 8 symbols and 0 affected processes.

**Two small things, neither blocking**

1. The reload also fires when `get_connection_driver` fails for a reason that is *not* a registry miss. `for_connection` can fail for plugins declaring `connection_metadata` when the DB is unreachable, and in that case we do a pointless rescan and repeat the `get_connection_metadata` RPC, which can double the wait up to `PLUGIN_CALL_TIMEOUT`. Gating the reload on `driver_registry::get_driver(&db_params.driver).await.is_none()` would make the intent explicit. No installed plugin declares `connection_metadata` today, so this is purely defensive.
2. `STARTUP_ERRORS` grows on every failed rescan in the MCP process: a plugin that can't load (missing executable, `min_runtime_version` too high) gets retried on every miss and pushes a new `PluginLoadError` each time, and nobody reads that vec in the subprocess. Slow and harmless in practice, but unbounded.

And one observation, out of scope for this PR: disabling a plugin still doesn't unregister it from a running MCP subprocess. The GUI calls `unregister_driver`, the subprocess never does. Worth a follow-up issue if you feel like it.

Approving. Nice fix, and the live repro in your comments made my job a lot easier.

aesslinger:
Thanks for the thorough review and the live repro, @debba — really appreciate it.

Both small things are addressed in 0ac26252:

1. The reload/retry is now gated on `driver_registry::get_driver(&db_params.driver).await.is_none()` (split into `resolve_driver_for_params` so it's directly testable) — a registered driver's own `for_connection` failure now surfaces immediately instead of triggering a pointless rescan and repeating the same failing RPC call.
2. `STARTUP_ERRORS` is now capped at 50 entries (oldest evicted first), so the standalone MCP subprocess's periodic rescans can't grow it unboundedly.

Filed the disable/unregister observation as #787 for follow-up.

PR #786, opened 2026-09-18
feat(updates): add core and plugin update badges and startup toast

debba:
## Summary

Surface available core and plugin updates without opening Settings. The sidebar now shows the total count, while Plugins and Info show their own counts with contextual tooltips. The Plugins badge previously counted configuration entries, including built-in drivers.

- Share plugin registry state across navigation, Settings, and the startup notification. Count installed plugins with an update compatible with the current app version and platform, and refresh the shared state after plugin changes.
- Show one dismissible startup toast when plugin updates are found. Clicking its message or **Open Plugins** opens Settings with the **Updates** filter, including when Settings is already open.
- Keep the core update badge and Info details available after **Remind me later**, while continuing to suppress the dismissed modal on background checks.
- Use subtle theme-derived badge colors and balanced toast icon, text, and action colors. Add localized tooltip and notification strings for all supported languages.

## Validation

- 126 targeted tests passed across 14 files, covering shared registry updates, stale request handling, offline recovery, counts and tooltips, toast navigation, dismissed core updates, existing toast behavior, and themes.
- `pnpm build` passed, including TypeScript compilation.
- ESLint passed for all changed source files; `git diff --cached --check` passed.
- Chromium rendered the actual badge, navigation, and plugin-update toast components on all 12 built-in themes. Verified the toast action dismisses the notification on each theme.
- Measured minimum contrast: badge numbers **5.89:1**; toast title **5.19:1**, message **4.69:1**, action **6.87:1**, and icon **5.89:1**.

## Theme previews

Actual components rendered in Chromium with simulated update data, across all 12 built-in themes.

### Update badges

Update badges across all 12 built-in themes

### Plugin update toast

Plugin update toast across all 12 built-in themes

## Manual verification

1. Start the app with an installed plugin that has a compatible newer release. Verify the sidebar and Plugins badges show the available update count and hover each badge to inspect its tooltip.
2. Click the startup toast or **Open Plugins**. Verify Settings opens directly on Plugins with the Updates filter, without repeating the toast during navigation.
3. Update or remove the plugin. Verify the counts refresh and disappear when no updates remain.
4. Detect a core update, then choose **Remind me later**. Verify the sidebar and Info retain the core badge and version details.
5. Repeat with no updates and an offline registry. Verify no startup update toast is shown for an empty or failed initial check.


aesslinger:
## UX/presentation review

Ran through this end-to-end (traced the code plus rendered the two screenshots you attached) to understand exactly what a user sees, and I think the layering is right — glanceable sidebar badge → contextual tab badges → one-time startup toast is a standard, proven pattern. A few execution details are worth tightening before this ships, roughly in priority order:

**1. The plugin-update toast has no persisted dismissal, unlike core updates.**
Core updates persist `lastDismissedVersion` to config, so "Remind me later" actually stays dismissed across restarts. `PluginUpdateToast`'s "already notified" flag is a plain `useRef(false)` — it resets every process launch. If a plugin update sits unresolved, the exact same toast fires again on every single app open, indefinitely. That's the setup for real notification fatigue over weeks of use (users learn to reflex-dismiss without reading), and it's inconsistent with the pattern this same PR's sibling feature already gets right one tab over. Worth giving plugin-update notifications the same "seen this version, don't repeat" persistence core updates have.

**2. Core and plugin update badges are visually identical.**
Sidebar, Plugins tab, and Info tab badges all render the same muted green with no differentiator beyond which row they're attached to (confirmed across all 12 themes in your screenshot — always the same green). A user can't tell "this needs a restart" from "this needs a click into Plugins" without hovering for tooltip text. Worth a distinct visual cue (icon, shape, or a second accent) for the two categories, since they have different remediation paths.

**3. Two competing tooltip mechanisms on the sidebar icon.**
`NavItem` now sets both a native `title` attribute and keeps the existing custom hover-flyout showing the same text — two different UI systems, two different timings/styles, on one element. Worth picking one (the existing themed flyout is more consistent with the rest of the app).

**4. `pluginsTooltip` concatenates every affected plugin into one run-on string.**
Fine for 1–2 plugins (as tested), but `"Updates available for: A 1.0→2.0, B 1.0→2.0, C ..."` squeezed into a native title tooltip won't scale gracefully for a user with several outdated plugins — no line-wrap guarantee, gets unreadable fast. A short list/popover would degrade better than a concatenated string.

## Idea to consider: per-connection update flag

One thing I'd flag *before* it gets built, not after: it's tempting to also surface "this connection's driver has an update" directly on each connection card in `Connections.tsx` — contextual, close to where the user actually works. I think that's worth doing, but only in one specific shape.

The risk: a plugin driver is shared across every connection that uses it. If 4 connections all use the `postgresql` plugin and each card shows an independent "Update" action, clicking one silently resolves all 4 — which breaks the mental model a user forms from seeing 4 separate flags in the first place (looks like 4 things, is actually 1), and mis-scopes the action itself (there's no such thing as "update this connection's driver," only "update the plugin," full stop).

The good news: this codebase already has the right template for exactly this shape of problem. `ConnectionCard.tsx` already renders a `deprecated` chip per card — driven off `driverManifest?.deprecated` (a driver-level fact, shared across every connection using that driver), rendered as a small **passive, non-interactive** text chip with a `title` tooltip, no `onClick`. That's the pattern to reuse:
- A read-only chip, styled like the existing deprecated chip, on every card whose driver has a pending update.
- No count on the chip — a count would imply "N updates for me" when there's exactly one shared update, reintroducing the exact confusion above.
- If it's clickable at all, it should only navigate to Plugins → Updates filter (same as the toast/sidebar badge already do) — never trigger the update inline from the card.

Happy to expand on any of these if useful — the underlying feature and test coverage are solid, this is purely about how the signal reads to someone using it day to day.

debba:
Thanks for the detailed review! I broadly agree with your points. I’d actually already considered persisting the plugin notification dismissal, so it makes sense to address that.

I particularly like the idea of adding an update indicator to the connection cards, with the behavior you described: keeping it informational and linking to Plugins → Updates, since the driver is shared across connections.

I’ll also look at making the core and plugin badges more distinguishable and cleaning up the tooltips. I’ll go ahead and work through these changes.

debba:
Thanks again @aesslinger — I’ve worked through all four points and added the connection-card indicator.

**Status:** implemented and validated locally; these follow-up code changes have **not yet been pushed to this PR**. The regenerated screenshots below show that local implementation.

- **Persistent plugin notifications:** the app now remembers the last release shown for each plugin in config. This is “seen this version, don’t repeat” persistence, recorded when the toast is shown, rather than only on explicit dismissal. Restarting, reordering the registry, or updating/removing another plugin won’t repeat an already-seen release. A new release can notify again; badges remain visible until the updates are resolved.
- **Distinct core/plugin badges:** different icons and theme-derived accents, with separate core/plugin counts in the sidebar instead of one visually ambiguous total.
- **One sidebar tooltip:** removed the native `title`; the existing themed flyout is retained.
- **Readable plugin details:** a themed, wrapping list showing up to five plugins plus the remaining count. It works on keyboard focus, closes with Escape, and renders outside the Settings scroll container to avoid clipping.
- **Connection-card indicator:** an informational chip with **no count**, shown on each connection using the affected plugin. Its tooltip explains that the plugin is shared. Clicking only navigates to **Plugins → Updates**; it never installs an update or connects to the database inline.

**Validation:** 4,454 frontend tests and 21 Rust config tests passed; build, TypeScript, and ESLint checks passed. Regenerated the previews in Chromium across all 12 built-in themes using actual components with simulated update data. Minimum measured badge/connection-chip text contrast: **5.58:1**.

<details>
<summary>Updated badges and connection indicators — all 12 themes</summary>

Core/plugin badges and connection indicators across all 12 themes

</details>

<details>
<summary>Updated plugin tooltips — all 12 themes</summary>

Plugin update tooltips across all 12 themes

</details>

debba:
## Follow-up: unified visual language for update cues (4567a9a9)

Connections, the Plugin Center and the sidebar rail now share one tone system (`src/utils/tones.ts`) and one `Chip` / `CountBadge` primitive, driven by theme tokens instead of hardcoded palette classes: `primary` for the current item and for updates, `success` for live or healthy state, `warning` for deprecated or disabled, `danger` for production and errors, `neutral` for attributes. The update cue is a static pill with an arrow and the target version; the rail carries one aggregated counter, Settings splits it into Info and Plugins.

Plugin Center changes: metric tiles double as filter shortcuts, the toolbar lives once in the header, cards keep one status row (version, installed or built-in, up to date, pending update) with taxonomy as low-emphasis tokens, and the footer has a soft split button (action plus version picker as a real listbox: keyboard navigation, flips upwards, scrolls instead of clipping) on the left and icon-only actions on the right. The Info tab opens its updates section with a status card (version, state, Release notes, Check for updates, Download and install). The connections list shows the same chips, in the same order, as the grid.

Tests: 4543 pass; `tests/themes/toneCoverage.test.ts` checks chip text contrast (WCAG AA), border visibility and update-versus-warning separation on every bundled preset.

Actual components rendered in Chromium with example data, all 12 built-in themes.

### Connections

Connections across all 12 built-in themes

### Plugin Center

Plugin Center across all 12 built-in themes

### Sidebar rail

Sidebar rail across all 12 built-in themes

Issue #788, opened 2026-09-18
Plugin driver shutdown kills in-flight requests instead of draining them


aesslinger:
## Summary

`unregister_driver` (`src-tauri/src/drivers/registry.rs`) removes a plugin driver from the registry and immediately calls `shutdown()`, which kills the underlying plugin subprocess without waiting for any request currently in flight against it. A request that already grabbed its `Arc<dyn DatabaseDriver>` before the unregister gets an abrupt, confusing failure instead of its real result — even though nothing about *its own* execution was wrong.

Found while investigating #787 (self-healing the standalone MCP subprocess's driver registry); this race is independent of #787 and already exists today via the GUI's `disable_plugin`/`uninstall_plugin`/`restart_plugin_process` commands.

## Details

`PluginProcess`'s management task (`src-tauri/src/plugins/driver.rs`) runs a `select!` loop racing three things: the shutdown signal, new outgoing calls, and incoming responses from the child process. The moment the shutdown `oneshot` resolves, the loop unconditionally calls `child.kill()` and breaks — regardless of what's sitting in `pending_requests`.

Concretely:
1. Request A calls `driver_registry::get_connection_driver(params)`, gets a clone of the `Arc<dyn DatabaseDriver>` (backed by `Arc<PluginProcess>`), and is awaiting its `oneshot::Receiver` inside `call_detailed`.
2. Concurrently, the plugin gets disabled/uninstalled/restarted (or, after #787 lands, self-heals out via the new reconcile check). `unregister_driver` removes the entry from `REGISTRY` and calls `shutdown()`.
3. The management task's loop observes the shutdown signal, kills the child, and breaks — dropping `pending_requests` (and Request A's response sender) without ever answering it.
4. Request A's `rx.await` resolves to a `RecvError`, which `call_detailed` turns into a generic `"Plugin process did not respond"` error.

This is not a memory-safety issue — the `Arc` keeps the Rust struct alive, there's no use-after-free. It's a liveness/UX issue: an in-flight call fails with a misleading message at the exact moment someone else (unrelatedly) disables/uninstalls/restarts that plugin. Whether a write query actually landed before the kill is left ambiguous.

New callers are already handled correctly — once removed from `REGISTRY`, any *new* `get_connection_driver` lookup fails cleanly with "Unsupported driver". The race is specifically about work that was already in flight before the unregister.

## Suggested fix direction

[...]

PR #789, opened 2026-09-18
fix: reconcile MCP driver registry against disabled/uninstalled plugins

### LINKED ISSUE #787: MCP subprocess keeps a disabled plugin's driver registered until restart
## Summary

Disabling a plugin's driver in the GUI does not propagate to a running standalone MCP subprocess: the driver stays registered there until the subprocess is restarted.

## Details

The GUI process calls `driver_registry::unregister_driver` when a plugin driver is disabled, which shuts down its process and removes it from the in-memory registry (`src-tauri/src/drivers/registry.rs`). The standalone MCP subprocess (`--mcp`) has no equivalent path: it only *adds* newly-enabled/installed drivers via a lazy rescan-on-miss (`resolve_driver_for_params` in `src-tauri/src/mcp/mod.rs`, from #783/#784), but nothing ever calls `unregister_driver` there in response to a disable.

Net effect: a connection using a driver that was disabled after the MCP subprocess started keeps working against that subprocess (using the stale, still-registered driver instance) until the subprocess is restarted, even though the GUI and any newly-spawned MCP subprocess would both correctly report it as unsupported.

## Suggested fix direction

Some way for the standalone MCP subprocess to observe a disable and unregister the driver — e.g. checking `active_external_drivers` against the current registry on the same rescan path introduced in #783, or another light-weight config-diffing mechanism.

## Context

Raised by @debba as an out-of-scope observation while reviewing #784:

> disabling a plugin still doesn't unregister it from a running MCP subprocess. The GUI calls `unregister_driver`, the subprocess never does. Worth a follow-up issue if you feel like it.

aesslinger:
## Summary

Fixes #787.

Disabling or uninstalling a plugin's driver in the GUI doesn't propagate to a running standalone MCP subprocess: the driver stays registered there until the subprocess is restarted. The subprocess previously only ever *added* drivers, via the registry-miss self-heal from #783/#784 — a driver that's still registered keeps resolving successfully, so it never produces the miss that self-heal relies on to notice anything.

- `drivers::registry::reconcile_active_drivers` diffs the registry against `active_external_drivers` and unregisters anything absent from it (builtins excluded via `manifest().is_builtin`). Covers disable *and* uninstall, since both write the same on-disk config field.
- `resolve_driver_for_params` now runs this reconcile check on every driver resolution (not just on a miss), gated by its own cooldown, independent of the existing miss-driven rescan so one path can't starve the other's responsiveness.
- `active_ids` is threaded in as a parameter (from each caller's already-loaded `AppConfig`) rather than read from disk inside `resolve_driver_for_params` — keeps the function testable without depending on the host machine's real `config.json`, and avoids a redundant disk read per call.

Full writeup of the investigation and design tradeoffs: https://github.com/TabularisDB/tabularis/issues/787#issuecomment-5733225913

A related, pre-existing race — a plugin driver's subprocess can be killed mid-in-flight-request by *any* unregister (this new path, or the existing `disable_plugin`/`uninstall_plugin` commands) — was found during this work and filed separately as #788, since it's independent of this fix and deserves its own review.

## Commits

1. Mechanical: split `drivers::registry`'s inline tests into a sibling `tests.rs`, no behavior change.
2. The actual fix.

## Test plan

- [x] `cargo test --lib` — all tests pass except 5 pre-existing failures unrelated to this change (askpass socket-path-length on this machine, one `storage_location_tests` assertion) — confirmed present on `main` too via `git stash`.
- [x] `cargo build`, `cargo fmt --check`, `cargo clippy` clean on touched files.
- [x] New unit tests for `drivers_to_unregister` (pure decision logic) and `reconcile_active_drivers` (registry mutation, including the builtin-exemption and `None` no-op cases).
- [x] Manual, against a real build and a real installed plugin, isolated via `TABULARIS_DATA_DIR` (no real user data touched): started `tabularis --mcp` with the `postgresql` plugin active, ran `list_databases` against a connection using it (real connection error — proves the driver resolved). Flipped `activeExternalDrivers` to exclude it on disk *without restarting the subprocess*, waited past the cooldown, ran the same call — got `"Unsupported driver: postgresql"`. Flipped it back on — the existing add-path self-healed it (real connection error again). Confirmed the exact `Reconciled driver registry against active_external_drivers, removed: ["postgresql"]` log line fired at the right moment.


aesslinger:
## Local review

Followed the repo's local-PR-review process: read against `.rules/rust.md` and `.rules/general.md`, verified every claim in the description against the actual code, ran gitnexus impact analysis, and staged a real end-to-end test rather than stopping at unit coverage.

### Claims verified against code (not taken on faith)

- **Builtins can't be spoofed into exemption.** `reconcile_active_drivers` excludes anything with `manifest().is_builtin`. Grepped every `is_builtin` assignment in the crate: hardcoded `true` only inside `mysql`/`postgres`/`sqlite`'s own source, `false` on every plugin-load path, and the `ConfigManifest` a plugin's own `.tabularium` deserializes into has no such field at all — no way for a plugin to mark itself immune.
- **Disable *and* uninstall are both covered.** Confirmed `PluginsTab.tsx`'s uninstall handler (`~L804`) and disable toggle (`~L842`) both write `activeExternalDrivers` — the same field `reconcile_active_drivers` reads. One mechanism, not two.
- **`active_ids` threading is complete.** All 5 real call sites (`handle_read_resource`, `tool_list_tables`, `tool_list_databases`, `tool_describe_table`, `tool_run_query`) pass it through; grepped for any remaining call to the old `resolve_driver_for_params(&db_params)` signature — none.

### Impact analysis (gitnexus)

`reconcile_active_drivers`: MEDIUM risk, 5 direct callers — all its own tests plus `resolve_driver_for_params`, exactly as expected. `tool_list_tables`/`tool_run_query`: LOW. `affected_processes: []` throughout — no surprise blast radius.

### Tests — full module paths, not partial names

- `cargo test --lib drivers::registry::tests::` → 10/10 pass.
- `cargo test --lib mcp::tests::` → 11/11 pass.
- Full `cargo test --lib` → same 5 pre-existing failures with or without this diff (`askpass` socket-path-length, one `storage_location_tests` assertion) — confirmed identical on the diff-free branch via `git stash`. Nothing new.
- `cargo fmt --check` / `cargo clippy` clean on all 4 touched files.

### Real data, not "test it manually"

Built the actual `tabularis` binary and used the already-installed real `postgresql` plugin (no fake stub). Isolated via `TABULARIS_DATA_DIR=/tmp/pr789-sandbox` — the plugins directory itself always follows the real default per `storage_location.rs`'s own docs, so nothing there needed touching, and the real `~/Library/Application Support/tabularis/config.json`/`plugins/` were never modified (verified timestamps unchanged before/after).

One long-running subprocess, no restart, three calls:

| Call | `activeExternalDrivers` on disk | Result |
|---|---|---|
| 1 | `["postgresql"]` | real connection error → driver resolved correctly |
| 2 | `[]` (simulated disable, same subprocess still running) | `"Unsupported driver: postgresql"` |
| 3 | `["postgresql"]` (re-enabled) | real connection error again → self-healed via the existing #783 add-path |

The exact log line `Reconciled driver registry against active_external_drivers, removed: ["postgresql"]` fired between calls 1 and 2, at precisely the moment it should have.

### Non-blocker notes

1. `disable_check_cooldown_elapsed`/`DISABLE_CHECK_COOLDOWN`/`LAST_DISABLE_CHECK_ATTEMPT` stay inline in `mcp/mod.rs` rather than a sibling module (rust.md 1). This mirrors the *pre-existing* `cooldown_elapsed`/`RELOAD_COOLDOWN` placement from #783/#784 — consistent with local precedent, not a new violation, but worth revisiting if that file's cooldown logic ever gets extracted on its own.
2. #788 (the pre-existing driver-shutdown race) is correctly left untouched here — filed separately, cross-linked in the description.

### Verdict

**Approve.** Root-cause fix, not a shortcut; tightly scoped to #787; follows the repo's own Rust conventions (pure helper extracted + tested, structural/behavioral changes split into separate commits); no security gap found; now backed by a real end-to-end run against the actual binary and a real plugin, not just unit tests.

aesslinger:
@debba ready for review

debba (approved):
Reviewed locally on the PR branch. Fix is correct and the claims in the description check out: tool calls load config per request so the threaded `active_ids` is fresh, builtins are exempt, and both GUI disable and uninstall write `activeExternalDrivers`.

Tests: `drivers::registry::tests` + `mcp::tests` 21 passed. Full `cargo test --lib` 1271 passed with only the known environment-only failures.

Also verified end to end with a debug build of this branch, `--mcp` against an isolated `TABULARIS_DATA_DIR`, the real postgresql plugin and a sqlite control connection, flipping `activeExternalDrivers` on disk without restarting the subprocess: `["postgresql"]` resolves (connection refused), `[]` gives `Unsupported driver: postgresql` after the cooldown, sqlite keeps working, re-enabling self-heals via the existing add path, and removing the key (None) is a no-op as documented. The `Reconciled driver registry ... removed: ["postgresql"]` log line fires at the right moments.

Non-blocking notes:

1. `disable_check_cooldown_elapsed` duplicates `reload_cooldown_elapsed` with its own static. A small `Cooldown` struct holding the duration and the last-attempt mutex would remove the copy. Cosmetic.
2. A call landing within 2s of a disable still uses the old driver. By design and documented, just worth keeping in mind when testing manually.
3. `PluginsTab` removes plugins that failed to start in the GUI from `activeExternalDrivers`. With this change a running MCP subprocess that loaded the same plugin fine will drop it too. Consistent with the field's meaning, no change needed, but it is a new observable effect.

debba:
The notes above are purely optional. As far as I'm concerned this is ready to merge as is.

